PR #10, opened 2020-09-02
Configure Purgecss for Tailwind

themsaid:


driesvints (approved):

themsaid:
I don't think the style StyleCI is trying to force is good. Leaving it to Taylor.

driesvints:
Ping @GrahamCampbell: can we force StyleCI to allow the current syntax? Not sure why it's not the default?

PR #13, opened 2020-09-02
Include autocomplete on all relevant elements.

Bradez:
This pull request adds an autocomplete attribute on all relevant elements to better comply with the HTML standard and improve user experience through improved browser autocomplete detection.

### DIFF
--- a/stubs/livewire/resources/views/profile/update-profile-information-form.blade.php
+++ b/stubs/livewire/resources/views/profile/update-profile-information-form.blade.php
@@ -49,7 +49,7 @@
         <!-- Name -->
         <div class="col-span-6 sm:col-span-4">
             <x-jet-label for="name" value="Name" />
-            <x-jet-input id="name" type="text" class="mt-1 block w-full" wire:model="state.name" />
+            <x-jet-input id="name" type="text" class="mt-1 block w-full" wire:model="state.name" autocomplete="nickname" />
             <x-jet-input-error for="name" class="mt-2" />
         </div>
 
--- a/stubs/livewire/resources/views/profile/update-profile-information-form.blade.php
+++ b/stubs/livewire/resources/views/profile/update-profile-information-form.blade.php
@@ -49,7 +49,7 @@
         <!-- Name -->
         <div class="col-span-6 sm:col-span-4">
             <x-jet-label for="name" value="Name" />
-            <x-jet-input id="name" type="text" class="mt-1 block w-full" wire:model="state.name" autocomplete="nickname" />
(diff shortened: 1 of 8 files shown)


jbrooksuk (approved):

stubs/livewire/resources/views/profile/update-profile-information-form.blade.php
```diff
@@ -49,7 +49,7 @@
         <!-- Name -->
         <div class="col-span-6 sm:col-span-4">
             <x-jet-label for="name" value="Name" />
-            <x-jet-input id="name" type="text" class="mt-1 block w-full" wire:model="state.name" />
+            <x-jet-input id="name" type="text" class="mt-1 block w-full" wire:model="state.name" autocomplete="nickname" />
```
innocenzi, line 52:
```suggestion
            <x-jet-input id="name" type="text" class="mt-1 block w-full" wire:model="state.name" autocomplete="name" />
```

stubs/resources/views/auth/register.blade.php
```diff
@@ -11,7 +11,7 @@
 
             <div>
                 <x-jet-label value="Name" />
-                <x-jet-input class="block mt-1 w-full" type="text" name="name" :value="old('name')" required autofocus />
+                <x-jet-input class="block mt-1 w-full" type="text" name="name" :value="old('name')" required autofocus autocomplete="nickname" />
```
innocenzi, line 14:
```suggestion
                <x-jet-input class="block mt-1 w-full" type="text" name="name" :value="old('name')" required autofocus autocomplete="name" />
```

innocenzi (commented):
I think it's better to keep `name` instead of specifying `nickname`. Either way, good change 👍

Bradez:
@innocenzi, the reason for not using `name` is because `name` denotes full name (ie. `Sir Timothy John Berners-Lee`) and it's an assumption that it would be used for a full name that includes honorific prefix / suffix and middle name.

`nickname`, on the other hand, is typically a shorter name (could be a nickname, or a screen name). That seems more likely what's going to be inserted here.

Finally, there's `username` although that's use-case specific. In Laravel's default configuration, it's not unique and therefore not a username.

In the end, `nickname` seems more appropriate than `name`.

innocenzi:
@Bradez While I agree that it's more likely that a shorter name is going to be used, the MDN still recommends using `name` instead of breaking down the field with more specific attributes. 

> Using "name" rather than breaking the name down into its components is generally preferred because it avoids dealing with the wide diversity of human names and how they are structured

In the eventuality that the user wants to register a full name, the autocompletion will be more complete. So I still think `name` is more appropriate than `nickname` for the vast majority of cases, and the developers using Jetstream will be able to change that easily if they want more specificity. 👍

Issue #17, opened 2020-09-02
Ask for password before enable/disable 2FA


zaherg:
Hey,

Since the user might login from multiple devices, it would be better to protect the ability to enable/disable (and for sure generating new recovery code) something like 

1. Just the enable/disable https://d.pr/v/itaIyQ
2. All buttons require confirmation https://d.pr/v/hrQW5R

I can submit a Draft PR with the code I added/changed to be reviewed and edited as required.

Thanks

taylorotwell:
Closing as this is a feature request.

zaherg:
hey @taylorotwell can I have your thoughts about this https://github.com/laravel/jetstream/pull/20

PR #20, opened 2020-09-02
WIP: Ask for password confirmation when enable/disable 2FA  [livewire stack]

zaherg:
This is a draft PR for #17 , it is still in early stages but if I got an approval I'll working to add inertia support too.

For now I have removed the ability to submit the password when pressing enter, as enabling two factor should be considered sensitive and should be dealt with carefully.

Thanks.

### DIFF
--- a/src/Http/Livewire/TwoFactorAuthenticationForm.php
+++ b/src/Http/Livewire/TwoFactorAuthenticationForm.php
@@ -3,6 +3,8 @@
 namespace Laravel\Jetstream\Http\Livewire;
 
 use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\Hash;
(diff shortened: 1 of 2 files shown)


taylorotwell:
Is this required? Is it standard practice to ask for passwords before enabling 2FA? 

Also, this would need to be updated for the recent Livewire 2.0 PR that was merged.

zaherg:
I am semi sure it is a standard thing, as it will make sure only those who knows the password can enable this feature.

I am in the process of updating my code to use livewire 2

RichardStyles:
Also don't most 2FA require a generated code to validate that the authenticator bit has been done before it's actually enabled.

zaherg:
Not always, *Some* will use the QR and do one step more where they test the
generated code to make sure that your code will works before enabling it.

PR #23, opened 2020-09-02
Support Livewire V2

calebporzio:
There are no breaking changes in V2 that affect Jetstream, however, there are some tweaks I'd make.

Summary:
* Upgrade AlpineJS
* Refactor Blade modal components to use `wire:model` && `@entangle` instead of `show=`
* Defer updating text inputs with `wire:model.defer`
* Remove unnecessary `wire:loading` statements

Take whatever you want, and leave the rest!

I did test all of this in a fresh install. Here are two things I had to do to make it work:
* `artisan view:clear`
* Update livewire in `composer.json` to version `^2.0` and run `composer update livewire/livewire`

> Note: The `install` command will continue to install Livewire V1 in its current setup until V2 is tagged. You could add a version constraint inside the `InstallCommand` (I added a branch alias so you can depend on `^2.0`), or wait for V2 to be officially tagged.


resources/views/components/button.blade.php
```diff
@@ -1,3 +1,3 @@
-<button {{ $attributes->merge(['type' => 'submit', 'class' => 'inline-flex items-center px-4 py-2 bg-gray-800 border border-transparent rounded-md font-semibold text-xs text-white uppercase tracking-widest hover:bg-gray-700 active:bg-gray-900 focus:outline-none focus:border-gray-900 focus:shadow-outline-gray transition ease-in-out duration-150']) }}>
+<button {{ $attributes->merge(['type' => 'submit', 'class' => 'inline-flex items-center px-4 py-2 bg-gray-800 border border-transparent rounded-md font-semibold text-xs text-white uppercase tracking-widest hover:bg-gray-700 active:bg-gray-900 focus:outline-none focus:border-gray-900 focus:shadow-outline-gray disabled:opacity-25 transition ease-in-out duration-150']) }}>
```
calebporzio, line 1: I added `disabled:opacity-25` so that opacity is automatically handled by CSS rather than `wire:loading.class="opacity-25"`.

> Note: I didn't re-build the CSS, so I'm guessing this class is missing from the build because of PurgeCSS.

resources/views/components/confirmation-modal.blade.php
```diff
@@ -1,6 +1,6 @@
-@props(['id' => null, 'show', 'maxWidth'])
+@props(['id' => null, 'maxWidth' => null])
 
-<x-jet-modal :id="$id" :show="$show" :max-width="$maxWidth ?? '2xl'">
+<x-jet-modal :id="$id" :maxWidth="$maxWidth" {{ $attributes }}>
```
calebporzio, line 3: The default of `$maxWidth ?? '2xl'` is already declared in the underlying modal component.

I figure it's best to forward all props as well as `$attributes`.

stubs/livewire/resources/views/api/api-token-manager.blade.php
```diff
@@ -13,7 +13,7 @@
             <!-- Token Name -->
             <div class="col-span-6 sm:col-span-4">
                 <x-jet-label for="name" value="Token Name" />
-                <x-jet-input id="name" type="text" class="mt-1 block w-full" wire:model="createApiTokenForm.name" autofocus />
+                <x-jet-input id="name" type="text" class="mt-1 block w-full" wire:model.defer="createApiTokenForm.name" autofocus />
```
calebporzio, line 16: I've deferred all text inputs. I suppose you could defer ALL inputs, but this will save on the majority of requests. Up to you.

stubs/livewire/resources/views/api/api-token-manager.blade.php
```diff
@@ -39,7 +39,7 @@
                 Created.
             </x-jet-action-message>
 
-            <x-jet-button wire:loading.class="opacity-25" wire:loading.attr="disabled">
+            <x-jet-button>
```
calebporzio, line 42: Livewire automatically disables submit buttons on form submission. Also, because we added `disabled:opacity-25`, we don't need to worry about that either.

stubs/livewire/resources/views/api/api-token-manager.blade.php
```diff
@@ -94,7 +94,7 @@
     @endif
 
     <!-- Token Value Modal -->
-    <x-jet-dialog-modal show="displayingToken">
+    <x-jet-dialog-modal wire:model="displayingToken">
```
calebporzio, line 97: To me `wire:model`, even if it's not actually used as `wire:model` in the underlying component is the best way to express data binding to inputs.

> Note: this also supports `wire:model.defer`

stubs/livewire/resources/views/layouts/app.blade.php
```diff
@@ -16,7 +16,7 @@
         @livewireStyles
 
         <!-- Scripts -->
-        <script src="https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.2.1/dist/alpine.js" defer></script>
+        <script src="https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.6.0/dist/alpine.js" defer></script>
```
calebporzio, line 19: There is a fix somewhere along the way in Alpine that prevents an obscure bug in the password forms caused by the One Password chrome extension. Moving to the most recent version fixes it.

3rgo:
Sorry to dig this up, but I was wondering it the difference in AlpineJS version between the layouts was intended :

- `stubs/resources/views/layouts/guest.blade.php` includes `alpine@v2.2.1`
- `stubs/livewire/resources/views/layouts/app.blade.php{ now includes `alpine@2.6.0`

@calebporzio you might have forgotten to update the guest layout ;)

PR #24, opened 2020-09-02
make jetstream extendable for other stacks by the community

Naoray:
This PR introduces the possibility for the community to create own presets based on Jetstream's existing functionality. For example if someone would rather use Vue without the Inertia stuff in the backend he could create a Vue preset by using

```
InstallCommand::macro('vue', function () {
  // do steps to setup vue preset with views etc.
});
```

Those little additions would help the community which is not yet so deep into Livewire and Inertia to come up with there own presets for Jetstream.


taylorotwell:
There are lots of formatting changes in this PR that make it impossible for me to see the actual changes you want.

Naoray:
Sorry, will raise another PR without formatting

PR #26, opened 2020-09-02
make jetstream extendable for other stacks by the community

Naoray:
This time without formating :)

This PR introduces the possibility for the community to create own presets based on Jetstream's existing functionality. For example if someone would rather use Vue without the Inertia stuff in the backend he could create a Vue preset by using

```
InstallCommand::macro('vue', function () {
  // do steps to setup vue preset with views etc.
});
```

Those little additions would help the community which is not yet so deep into Livewire and Inertia to come up with there own presets for Jetstream.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -6,6 +6,7 @@ use Exception;
 use Illuminate\Console\Command;
 use Illuminate\Filesystem\Filesystem;
 use Illuminate\Support\Str;
+use InvalidArgumentException;
 use Symfony\Component\Process\Process;
 
 class InstallCommand extends Command
@@ -60,6 +61,15 @@ class InstallCommand extends Command
             '\Laravel\Jetstream\Http\Middleware\AuthenticateSession::class',
             app_path('Http/Kernel.php')
         );
+
+        // Install custom Stack
+        if (static::hasMacro($this->argument('stack'))) {
+            return static::{$this->argument('stack')}();
+        }
+
+        if (!in_array($this->argument('stack'), ['livewire', 'inertia'])) {
+            throw new InvalidArgumentException('Invalid stack.');
+        }
 
         // Install Stack...
         if ($this->argument('stack') === 'livewire') {
--- a/src/JetstreamServiceProvider.php
+++ b/src/JetstreamServiceProvider.php
@@ -162,6 +162,10 @@ class JetstreamServiceProvider extends ServiceProvider
      */
     protected function configureRoutes()
     {
+        if ($this->stackIsNotSupported()) {
+            return;
+        }
+
         Route::group([
             'namespace' => 'Laravel\Jetstream\Http\Controllers',
             'domain' => config('jetstream.domain', null),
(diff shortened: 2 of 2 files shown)


ConsoleTVs:
This is great!

taylorotwell:
I'm gonna be honest, I'm not sure of the community's ability to maintain 1:1 compatible stacks with Jetstream in the long term.

ConsoleTVs:
I understand the concerns here @taylorotwell , however I think jetstream should not force users to use specific adapters, this would somehow be gatekeeping other people's ability to extend this to other frameworks that are quite much popular than the stack used here. Of course this would put maintainers of those new adapters in the position of keeping a 1:1 integration. I highly doubt this should be of our concern in this library since at the end of the day, everybody can use the official provided adapters if they want.

Perhaps it would be cool to explore other ideas on how this should be handled if this solution was not of your liking 😉

innocenzi:
@taylorotwell Even though I think the same, I still think it'd be nice to let the ability for users to extend Jetstream

taylorotwell:
Well, we aren't really gate keeping that. Anyone can write a Jetstream frontend already and publish it as a package and tell people to `composer require laravel/jetstream` and then run `php artisan some-package:install` instead of `jetstream:install` and that package can do whatever it wants.

ConsoleTVs:
> Well, we aren't really gate keeping that. Anyone can already write a Jetstream frontend and publish it as a package and tell people to composer require laravel/jetstream and then run php artisan some-package:install instead of jetstream:install and that package can do whatever it wants.

Oh well, this is a solution I was not aware of, great to hear that!

taylorotwell:
As I noted above, you can already make your own Jetstream stack and publish them. No changes are needed in Jetstream to support this.

AnandPilania:
@taylorotwell https://github.com/laravel/jetstream/pull/26#issuecomment-686064646 solution is good, but can u separate the non-stack installation, like, `Actions` & `model` + `sanctum`, so that `custom-package` can directly call it and the core functionality of package remains same https://github.com/laravel/jetstream/pull/26#issuecomment-685881220l.

Actually I'm using it in my local package, no stack, only tailwind + jQuery,

after `$this->installFortifyServiceProvider()`:
```
$this->installJetDefaults(); //will install & publish Sanctum + add Provider, User model & Actions
$this->installJetstreamServiceProvider();
```

replaced Install Stack conditions with:
```
if ($this->argument('stack') === 'livewire') {
  $this->installLivewireStack();
} elseif ($this->argument('stack') === 'inertia') {
  $this->installInertiaStack();
} else {
  $this->installNoStack();
}
```

AnandPilania:
@taylorotwell https://github.com/laravel/jetstream/pull/26#issuecomment-686064646 solution is good, but can you separate the non-stack installation + configuration from `console + service provider`, so that `custom-package` will only handle `routes => controllers and views` and the **core functionality of package remains same** https://github.com/laravel/jetstream/pull/26#issuecomment-685881220l.

You can check the basic setup (only skeleton setup yet) @[https://github.com/AnandPilania/laravel-jetstream](url) where I'm using only tailwind, like:

after `$this->installFortifyServiceProvider()`:
```
$this->installJetDefaults(); //will install & publish Sanctum + add Provider, User model & Actions + update configuration with `null` instead of any stack
$this->installJetstreamServiceProvider();
```

replaced Install Stack conditions with:
```
if ($this->argument('stack') === 'livewire') {
  $this->installLivewireStack();
} elseif ($this->argument('stack') === 'inertia') {
  $this->installInertiaStack();
} else {
  $this->installNoStack(); // handling all tasks related to non-stack
}
```

PR #31, opened 2020-09-03
Add password confirmation to 2FA - livewire

zaherg:
This PR adds a password confirmation dialog when enable/disable or generate a new codes for 2FA.
More info about how this works can be seen here https://d.pr/v/6C9rjI

Most companies require the user to confirm their identity before enable/disable the 2FA, and even require them to test the code after scanning the QR and before enabling 2FA.

We need to remember the user can login from multiple devices and can have the remember me checked, and if he forget to logout from a public computer any one can enable the 2FA and lock the main user out of his own account. Requiring the password makes such thing **a bit hard**.

Thanks.

### DIFF
--- a/src/Http/Livewire/TwoFactorAuthenticationForm.php
+++ b/src/Http/Livewire/TwoFactorAuthenticationForm.php
@@ -3,6 +3,8 @@
 namespace Laravel\Jetstream\Http\Livewire;
 
 use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\Hash;
+use Illuminate\Validation\ValidationException;
 use Laravel\Fortify\Actions\DisableTwoFactorAuthentication;
 use Laravel\Fortify\Actions\EnableTwoFactorAuthentication;
 use Laravel\Fortify\Actions\GenerateNewRecoveryCodes;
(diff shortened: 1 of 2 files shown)


taylorotwell:
I guess I'm not sure what we're solving here. They can already lock the user out of their account by just changing the email address associated with the account to something the user does not know. In addition, they already have full access to the account and could already perform any destructive action they want?

taylorotwell:
But, I will say that one thing that

taylorotwell:
Also, I should note that the malicious user could just execute a backend request to disable 2FA directly and thus bypassing the modal based password check. The check doesn't put anything in the session to confirm the password was actually entered.

zaherg:
> They can already lock the user out of their account by just changing the email address associated with the account to something the user does not know. 

Then I guess changing the email should some something also protected by password.

> In addition, they already have full access to the account and could already perform any destructive action they want?

True, and I can't argue with that, I have seen multiple companies protect the following action by passowrd:

1. Changing the password, which required the current password by default.
1. Changing the email.
1. Changing the 2FA status, and 2FA will have its own page and they wont show you the recovery key at all, like in google account for example, if you need a new recovery keys you will have to disable/enable your 2FA again.

Even though as you have mentioned, if anyone have full access to the account they could perform any destructive action. but this does not spare us from the responsibility of protecting some of these actions, at least to raise awareness about security and protecting their data.

Again, this is just a suggestion for something **most companies** are doing, I am far away from being a security expert to say that this action will surely protect the user data.

zaherg:
> The check doesn't put anything in the session to confirm the password was actually entered.
This is something I missed and didn't thought about it, you are correct.

You know the coder better than me, and I am sure you can (when needed) provide a more accurate/safer code to protect those functionality.

Thanks

codytooker:
I'd like to add that it is slightly confusing that 2FA is active as soon as someone clicks the enable button.  This means someone can lock themselves out of their account because they simply clicked the button not knowing what 2FA even is.

Seems like the process should be.

**To Setup**
1. Click enable button
2. show QR code and recovery codes
3. Prompt user to insert a valid 2FA token - until this happens we don't know if the user has saved anything.
4. On success save 2FA tokens in DB

**To Deactivate**
1. click disable
2. prompt user to insert a valid 2FA token
3. on success remove 2FA tokens from DB

Issue #32, opened 2020-09-03
Bug: API Token permissions aren't bound properly to livewire


bellalMohamed:
- Jetstream Version: ^0.6.0
- Laravel Version: ^8.0
- PHP Version: 7.3.19
- Database Driver & Version: Mysql 8.0.19
- Livewire/Inertia: Livewire

### Description:
Created Token permissions are always wrong

### Steps To Reproduce:
1- Create fresh laravel project using the new laravel installer ```laravel new foo --jet --dev``` 
2- Configure the DB & run migrations
3- Register new user
4- Go to API Tokens Page
5- Select all Permissions, Enter token name and Create A new token
6- Check ```personal_access_tokens``` table or click on token permissions you will find that not all permissions are set as you selected

PR #33, opened 2020-09-03
fix: fixes livewire permissions binding

bellalMohamed:
Fixes: #32 

I did check the documentation of Livewire and actually didn't find anything mentioning ```wire:model.defer``` removing .defer or adding both .lazy work fine


taylorotwell:
Removing `defer` is likely not the answer. I will ask Caleb.

calebporzio:
Ok, this was a bug in Livewire V2.

Adding `wire:model.defer` didn't work with checkboxes, because Livewire would always use the value of the latest checkbox checked.

Checkboxes are unique to other inputs in that they have multiple elements that combine to form a unified value. Livewire now accounts for this in the following commit: https://github.com/livewire/livewire/commit/68fbb550f97d950d7e16c8fab131a5e6e69d4a84

If you update Livewire in your application using `composer update livewire/livewire`, you should be all set 👌

taylorotwell:
Thanks!

PR #34, opened 2020-09-03
Using assertSame to make assert equals strict

peter279k:
# Changed log

- Using the `assertSame` to replace `assertEquals` to make assertion equals checking strict.


driesvints:
Thanks

PR #36, opened 2020-09-03
make user friendly installer 

joisarjignesh:
make user friendly installer as you @taylorotwell  share on twitter like this
Screenshot from 2020-09-03 22-47-07


taylorotwell:
I think I will leave this as-is for now.

PR #40, opened 2020-09-03
Always show scroll bars

bonzai:
Currently, the vertical scroll bar is only shown when needed, so sometimes the page is "jumping":

scroll

After this PR, the scroll bars will always be visible.

### DIFF
--- a/stubs/inertia/resources/views/app.blade.php
+++ b/stubs/inertia/resources/views/app.blade.php
@@ -17,7 +17,7 @@
         <script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.26.0/moment.min.js"></script>
         <script src="{{ mix('/js/app.js') }}" defer></script>
     </head>
-    <body class="font-sans antialiased">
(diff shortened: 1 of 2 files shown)


zaknesler:
`overflow: scroll` never looks good on Windows

image

bonzai:
@zaknesler Nothing wrong with your screenshot.

zaknesler:
@bonzai An empty scrollbar isn't a common choice of any website, it should only appear when necessary. The "jumping" is expected when switching to a page with longer content.

image

octoper:
@bonzai The page is has more content that's why the Profile settings page has scrollbars, the Dashboard page on the other hand it doesn't so is has no scrollbars.
You don't need to show scrollbars on every page its not attractive to have disable scrollbars or even have scrollbars on a page that never scrolls, the users might be confused sometimes

bonzai:
@zaknesler Well, I would expect that the menu or other elements of the page are always in the same place, no matter if the scroll bar is visible or not.

taylorotwell:
This is a stub so you can change it in your own app.

PR #55, opened 2020-09-04
Fix InstallCommand.php

dleicam:
On windows, the Services Providers was not added to the app.php file.

image

After some tests I realized that it would be enough to remove the PHP_EOL tag.


dleicam:
This update is important for Windows installations. I think that by logic it will not affect the other operating systems.

taylorotwell:
Isn't it going to have two line breaks now for Mac users?

dleicam:
> Isn't it going to have two line breaks now for Mac users?


Hi Taylor,

It is not supposed to. I don't have a Mac to test.

But as it is now, on windows it will be necessary to add it manually.

PR #56, opened 2020-09-04
Make language dynamic on layout files

m1guelpf:
This PR sets the language attribute on the HTML tag of both stacks to follow the application's configured locale, like the welcome view does.

PR #58, opened 2020-09-05
Use application language on layouts

m1guelpf:
This PR sets the language attribute on the HTML tag of both stacks to follow the application's configured locale, like the welcome view does.

See #56


jbrooksuk:
Thanks!

PR #63, opened 2020-09-05
set fully qualified url for dashboard,api tokens,profile show,team se…

joisarjignesh:
set fully qualified URL for dashboard,api tokens,profile show,team settings and create new team

### DIFF
--- a/stubs/livewire/resources/views/layouts/app.blade.php
+++ b/stubs/livewire/resources/views/layouts/app.blade.php
@@ -27,14 +27,14 @@
                         <div class="flex">
                             <!-- Logo -->
                             <div class="flex-shrink-0 flex items-center">
-                                <a href="/dashboard">
+                                <a href="{{ url('dashboard') }}">
                                     <x-jet-application-mark class="block h-9 w-auto" />
                                 </a>
                             </div>
 
                             <!-- Navigation Links -->
                             <div class="hidden space-x-8 sm:-my-px sm:ml-10 sm:flex">
-                                <x-jet-nav-link href="/dashboard" :active="request()->routeIs('dashboard')">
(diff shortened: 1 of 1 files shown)


nathanheffley:
Does this solve an issue you were having? I'm interested in what benefit you think absolute URLs provide here.

kdunn2856:
Nathan, 

Thank you so very much for your gracious help!

I hope to some day return the favor.

Kim Dunn

kdunn2856:
Thank you very much for the information and direction.  It is very helpful.

Kim Dunn

pedroufv:
Why use url('/dashboard') instead of route('dashboard')?

I saw on the welcome.blade using url('/dashboard') too, but i don't know if has benefits

joisarjignesh:
> Why use url('/dashboard') instead of route('dashboard')?
> 
> I saw on the welcome.blade using url('/dashboard') too, but i don't know if has benefits

i did't know route('dashboard') exists or not so i used url instead of route.

joisarjignesh:
> Does this solve an issue you were having? I'm interested in what benefit you think absolute URLs provide here.
@nathanheffley @taylorotwell 


When using absolute URLs, the link always points to the correct URL and when you have multiple projects structure  like
```
html
      /laravel7
      /laavel8
      /jestream
     /project/xyz
     /project/foo
 ```
and you are using like apache(or nginx,xampp) and  your single(only one)  host file pointing(documentRoot /var/www/html) the main **apache folder html.**.  then you are using relative url it will fall down (because it will always point to main root folder)
 
In general, it is considered best-practice to use relative URLs, so that your website will not be bound to the base URL of where it is currently deployed. For example, it will be able to work on localhost, as well as on your public domain, without modifications.
see question on stackoverflow https://stackoverflow.com/questions/2005079/absolute-vs-relative-urls

so sent you pull request for absolute url

PR #54, opened 2020-09-04
Provide a way of customizing/disabling the default routes

m1guelpf:
This PR introduces a way of disabling registration of the default routes like other Laravel packages already allow (Cashier, for example). It also adds an option to publish the routes for the adapter you're currently using to your routes folder.

See #44

PR #64, opened 2020-09-05
Allow custom routes

MarkIvanowich:
Matching a proposed PR in Fortity, This PR adds the ability to override Jetstream's routes file with an option in the configuration file.

Related issue/feature request: https://github.com/laravel/fortify/issues/9 and Fortify PR: https://github.com/laravel/fortify/pull/10

For my example use case, I put a route prefix before all authentication endpoints. This PR would allow Jetstream and Fortify the same flexibility.

When using the laravel/ui package, defining routes was an opt-in by calling the Auth::routes() on the app's routes file. In Jetstream and Fortify, they are booted with autoloading.

(I :heart: Laravel)

### DIFF
--- a/src/JetstreamServiceProvider.php
+++ b/src/JetstreamServiceProvider.php
@@ -160,6 +160,12 @@ class JetstreamServiceProvider extends ServiceProvider
      */
     protected function configureRoutes()
     {
+        if ($path = config('jetstream.routes', null)) {
+            $this->loadRoutesFrom($path);
+
+            return;
+        }
+
         Route::group([
             'namespace' => 'Laravel\Jetstream\Http\Controllers',
             'domain' => config('jetstream.domain', null),


rodrigopedra:
Hi @MarkIvanowich 

I think in this PR you should also try changing the hard-coded routes to use the `route()` helper and get them by name.

This way when someone customize the routes they can still use the shipped view files if they keep the same route names.

On my PR to `laravel/fortify` I didn't find any hard-coded route path (actually on logout action it redirects to `route('/')` but that is harmless)

MarkIvanowich:
> changing the hard-coded routes to use the route() helper and get them by name

I'm 50/50 about going in and changing the hard-coded routes by name.

Originally, I was thinking quite narrowly with my specific use case. I thought your commit was more flexible [overriding a routes file path], which I then copied here.

If the user edits `config/jetstream.php` to add the line
```php
'routes' => base_path('routes/jetstream.php')
```
They would need their own copy of the package's routes file, which is the same as a `vendor:publish`. If they publish the routes, then they are likely to have already published the views.

I'll create another branch and PR for it, its not detrimental.

m1guelpf:
See #54

rodrigopedra:
Hi @m1guelpf 

I was looking at commit https://github.com/laravel/jetstream/commit/70fe396419f2ed31b6055625b0c559950c99939c

To see the changes as PR #54 UI lists no changed files (it seems you have pull bot enabled on your account https://github.com/apps/pull )

One doubt: just publishing the routes will make the service provider aware it would need to load the user-land publishes routes and not the package one?

Asking because in line 167 below the path to the routes file is hard-coded:

https://github.com/laravel/jetstream/blob/49e57d42ba019453e78083e255e08af427cc26da/src/JetstreamServiceProvider.php#L161-L169

I think we still need to tweak the `configureRoutes` method to load the published file.

Let me know if I am missing something

MarkIvanowich:
After the strangeness from the pull bot is gone, I much prefer https://github.com/laravel/jetstream/pull/67 over this, so I'm closing.

Now to get similar operation into Fortify...

PR #66, opened 2020-09-05
Typo: DeleteUserWithTeams was sometimes DeleteUser

chimalsky:
At first I thought this was some strategy to allow Developers to flexibly use `DeleteUser` Action but after digging thought maybe it was just a typo.


taylorotwell:
Current code is correct.

PR #67, opened 2020-09-05
Provide a way of customizing/disabling the default routes

m1guelpf:
This PR introduces a way of disabling registration of the default routes like other Laravel packages already allow (Cashier, for example). It also adds an option to publish the routes for the adapter you're currently using to your routes folder.

Resubmission of #54

### DIFF
--- a/src/Jetstream.php
+++ b/src/Jetstream.php
@@ -11,6 +11,13 @@ use Laravel\Jetstream\Contracts\UpdatesTeamNames;
 class Jetstream
 {
     /**
+     * Indicates if Jetstream routes will be registered.
+     *
+     * @var bool
+     */
+    public static $registersRoutes = true;
+
+    /**
      * The roles that are available to assign to users.
      *
      * @var array
@@ -298,5 +305,17 @@ class Jetstream
     public static function deleteUsersUsing(string $class)
     {
         return app()->singleton(DeletesUsers::class, $class);
+    }
+
+    /**
+     * Configure Jetstream to not register its routes.
+     *
+     * @return static
+     */
+    public static function ignoreRoutes()
+    {
+        static::$registersRoutes = false;
+
+        return new static;
     }
 }
--- a/src/JetstreamServiceProvider.php
+++ b/src/JetstreamServiceProvider.php
@@ -151,6 +151,10 @@ class JetstreamServiceProvider extends ServiceProvider
             __DIR__.'/../database/migrations/2020_05_21_100000_create_teams_table.php' => database_path('migrations/2020_05_21_100000_create_teams_table.php'),
             __DIR__.'/../database/migrations/2020_05_21_200000_create_team_user_table.php' => database_path('migrations/2020_05_21_200000_create_team_user_table.php'),
         ], 'jetstream-team-migrations');
+
+        $this->publishes([
+            __DIR__.'/../routes/'.config('jetstream.stack').'.php' => base_path('routes/jetstream'),
+        ], 'jetstream-routes');
     }
 
     /**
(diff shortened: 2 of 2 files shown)


MarkIvanowich:
Now that we can see the changes, I much prefer this solution to my own.

src/Jetstream.php
```diff
@@ -10,6 +10,13 @@
 
 class Jetstream
 {
+    /**
+     * Indicates if Cashier routes will be registered.
```
MarkIvanowich, line 14: Change doc comment to "Jetstream"

m1guelpf: Good catch!

taylorotwell:
Do you have a really good reason to do this? I'm trying to understand if this would ever actually be useful.

m1guelpf:
My use-case for this is to have more control over the path each route uses (for my application, I wanted to leave the `/profile` route for actual profiles and use `/account` for Jetstream settings, which I don't think will be a super common use-case, but I think giving the option to the user is still a good idea.

rodrigopedra:
Hi @taylorotwell , 

Now that this PR was accepted, can you reconsider the similar PR for Fortify?

PR https://github.com/laravel/fortify/pull/13 by @MarkIvanowich 

It basically does the same as this one.

For some projects I work on, customizing the authentication routes is a requirement. Specially when migrating old ones that already uses custom routes and have some integration in place with other systems.

Thank you very much.

yob-yob:
Hi, can I ask how can i use this feature?

How should I override the value of `$registersRoutes`? is there any good recommendation? like should I do it on `AppServiceProvider`? or on `JetStreamServiceProvider`? and how should I do it?

should I just say `Laravel\Jetstream\Jetstream::registersRoutes = false;` on the boot methods of any of this files?

m1guelpf:
You just need to run `Jetstream::ignoreRoutes()` on the boot method of your `AppServiceProvider`.

yob-yob:
okay, so I guess after that I should be able to add this code to my `JetstreamServiceProvider`

```

**Comment by rhurling** [comment]
I had to move `Jetstream::ignoreRoutes()` to the register method of my `AppServiceProvider`
(since the `Laravel\Jetstream\JetstreamServiceProvider::boot` method runs before the `App\Providers\AppServiceProvider::boot` method in the default cached configuration)

File: src/JetstreamServiceProvider.php [hunk 2]
```diff
@@ -160,12 +164,14 @@ protected function configurePublishing()
      */
     protected function configureRoutes()
     {
-        Route::group([
-            'namespace' => 'Laravel\Jetstream\Http\Controllers',
-            'domain' => config('jetstream.domain', null),
-        ], function () {
-            $this->loadRoutesFrom(__DIR__.'/../routes/'.config('jetstream.stack').'.php');
-        });
+        if (Jetstream::$registersRoutes) {
```
joelbutcher, line 167: @m1guelpf Could you add a matching method to the Jetstream class and call that here, instead of directly accessing the class property? It would keep the code consistent with the patterns used across this package

joevallender:
This is a great option to have. I recently wanted to change the Team(s) vocabulary to Project(s) and this worked perfectly. 

I've written up the steps I followed to make the change in a blog post in case it's useful for others 
https://aboutweb.dev/blog/change-laravel-jetstream-route-paths/

mokhosh:
@m1guelpf this doesn't let you customize the routes or disable only some of them, right? All or nothing?

joevallender:
@mokhosh You can change whatever you want. Look at the routes file which is published https://github.com/laravel/jetstream/blob/2.x/routes/inertia.php

Once you've completed the necessary steps, you'll have a copy of this (or the Livewire equivalent) file in your project and can edit it however you see fit

PR #72, opened 2020-09-06
[livewire][teams]Update state property to include all team fields

ClementN:
```src/Http/Livewire/UpdateTeamNameForm.php```

Mount ```$state``` property to include *all* existing fields of the ```Team``` Model instead of the ```name``` only 
in order to allow the implementation of additional, custom, fields for the ```Team``` Model. 

That way, **custom fields can be added** to the ```teams``` table, and be able to update them through ```app/Actions/Jetstream/UpdateTeamName.php``` and ```resources/views/teams/update-team-name-form.blade.php```.

### DIFF
--- a/src/Http/Livewire/UpdateTeamNameForm.php
+++ b/src/Http/Livewire/UpdateTeamNameForm.php
@@ -32,7 +32,7 @@ class UpdateTeamNameForm extends Component
     {
         $this->team = $team;
 
-        $this->state = ['name' => $team->name];
+        $this->state = $team->toArray();
     }
 
     /**

### CODE REVIEW

**Comment by calebporzio** [comment]
I could see the need for something like this, given the current setup is SO close to allowing you to add additional fields like you want.

I personally wouldn't assign a public property in Livewire to a `->toArray()` call.

That would mean that the `$state` property would now include fields from the model like `id`, and `created_at` and would be exposed in plain text to the front-end. That means any field that gets added to the `teams` table would be exposed to the browser without the developer knowing.

Here's what I would personally do if I encountered this need: extend the `UpdateTeamNameForm` Livewire component with your own component and modify the `mount()` method to make it do what you want.

Here's exactly how to do it:

* Create a new file and fill it with these contents:  `app/Http/Livewire/UpdateTeamForm.php`
```php
<?php

namespace App\Http\Livewire;

use Laravel\Jetstream\Http\Livewire\UpdateTeamNameForm;

class UpdateTeamForm extends UpdateTeamNameForm
{
    public function mount($team)
    {
        $this->team = $team;

        $this->state = [
            'name' => $team->name,
            'description' => $team->description,
        ];
    }

    public function render()
    {
        return view('teams.update-team-name-form');
    }
}
```
* Now that we've extended the original component, we need to update the reference from Jetstream's component to our new one. Open `resources/views/teams/show.blade.php` and change the following line:
```php
// Before
@livewire('teams.update-team-name-form', ['team' => $team])

// After
@livewire('update-team-form', ['team' => $team])
```
* Now you can add your new input fields in `resources/views/teams/update-team-name-form.blade.php` and everything should work nicely.
* Oh, of course, you'll have to update `app/Actions/Jetstream/UpdateTeamName.php` to validate and persist your new form.

image

If this is all too confusing or seems like overkill, I would recommend creating your own component completely and just pasting in the existing view stub as a starting point.

I think this a good start to the ongoing discussion of extending Jetstream behavior. There's probably other good patterns yet to emerge.

ClementN:
Thank you for the extended explanations and solutions, Caleb.

While I understand why having ALL model fields in a public property is not recommended, 
**this idea raises the same issue for another file**, ```vendor/laravel/jetstream/src/Http/Livewire/UpdateProfileInformationForm.php```,
as this file assign all the ```User``` fields to a ```->toArray()``` call. ( it was the source of inspiration for my solution ). 

```
    public function mount()
    {
        $this->state = Auth::user()->withoutRelations()->toArray();
    }
```

taylorotwell:
@ClementN so one thing I'm interested in here is *what* exactly you wanted to add to this panel. It's just for updating the team name so I would expect any other team related things you want to do should just be new panels on the team/show page instead of trying to modify this panel?

ClementN:
@taylorotwell having a separate panel for any other fields related to teams (except the name) seems ok. 

To answer your question: Team ```status``` was one field I added to the ```teams``` table 
Having it in the same table with the ```name``` field I was tempted to implement it in the same way and same panel.

PR #76, opened 2020-09-06
Fixes Error not found on Delete API Token 

gpanos:
- Jetstream Version: 0.6.4
- Laravel Version: 8.x-dev
- PHP Version 7.4 

## Description:

There is a weird behavior when trying to delete a token after first opening the manage token permissions modal. On deletion I get an error 404 not found. 

## Steps to reproduce: 

1) Go to API Tokens management. 
2) Create a token 
3) Click on permissions to open the modal and then click on Nevermind button 
4) Try to delete the token 

Here is a gif showcasing the issue: 
capture

Changing the visibility of the `managingPermissionsFor` property to protected fixes the issue.

### DIFF
--- a/src/Http/Livewire/ApiTokenManager.php
+++ b/src/Http/Livewire/ApiTokenManager.php
@@ -45,7 +45,7 @@ class ApiTokenManager extends Component
      *
      * @var \Laravel\Sanctum\PersonalAccessToken|null
      */
-    public $managingPermissionsFor;
+    protected $managingPermissionsFor;
 
     /**
      * The update API token form state.


calebporzio:
Thanks for submitting this @gpanos.

**In short:** I was able to reproduce the issue (it only happens sometimes) and have fixed it. Run `composer update livewire/livewire` and you shouldn't see that issue anymore.

Here's the fix commit for reference: https://github.com/livewire/livewire/commit/802935ac448aefcba1240a7c304fa8aaae6197e1

**In long:** Because that Jetstream modal uses the new `@entangle` syntax for entangling the "shown" value with Livewire, here's what was happening:

* (Assume the delete confirmation modal is open)
* You press the "delete" button
* A Livewire AJAX request is triggered to call the `deleteApiToken` method in `ApiTokenManager`
* Inside the `deleteApiToken` method `$this->confirmingApiTokenDeletion = false;` is called.
* When Livewire's JS receives the request, it notices that `confirmingApiTokenDeletion` has been changed to false
* Because `confirmingApiTokenDeletion` is `@entangle`d with the Alpine modal, Livewire updates the state of the Alpine modal component
* Because `@entangle` is "watching" for Alpine state changes, it notices the state change and fires a Livewire request (which is now redundant)
* That send request went undetected because it's normally benign. However, because we've already "deleted" the model, when Livewire tries to serialize the model for that second round-trip, you get a 404.

If you didn't follow that. Basically there was a circular dependency with `@entangle`, and a prevention mechanism has now been put in place.

gpanos:
@calebporzio yep you nailed it thanks a lot for the deep explanation! From reporting to a fix in 15 minutes... You are a machine :)

Issue #79, opened 2020-09-06
Impossible to copy User model


Destinais:
- Jetstream Version: 0.7.0
- Laravel Version: ^8.0
- PHP Version: 7.3
- Database Driver & Version: X

### Description:
When you install an old project where the User model isn't in Models folder :
```php artisan jetstream:install livewire``` failed with this trace 
```
  at vendor/laravel/jetstream/src/Console/InstallCommand.php:165
    161▕
    162▕         $this->installJetstreamServiceProvider();
    163▕
    164▕         // Models...
  ➜ 165▕         copy(__DIR__.'/../../stubs/app/Models/User.php', app_path('Models/User.php'));
    166▕
    167▕         // Actions...
    168▕         copy(__DIR__.'/../../stubs/app/Actions/Fortify/CreateNewUser.php', app_path('Actions/Fortify/CreateNewUser.php'));
    169▕         copy(__DIR__.'/../../stubs/app/Actions/Fortify/UpdateUserProfileInformation.php', app_path('Actions/Fortify/UpdateUserProfileInformation.php'));
```
### Steps To Reproduce:
Use a project without User model in Models folder

m1guelpf:
See [Taylor's response](https://github.com/laravel/jetstream/pull/80#issuecomment-687990916)

PR #80, opened 2020-09-06
#79 - Create Models folder if not exist

Destinais:
#79


taylorotwell:
Should exist because this is only for new 8.x applications.

Destinais:
Do we have to pass the models into Models folder during a migration from laravel 7 to 8 ?

PR #91, opened 2020-09-08
Delete User password confirmation

SeanBourke:
Introduces password confirmation for user deletion on Livewire and Inertia

### DIFF
--- a/src/Http/Livewire/DeleteUserForm.php
+++ b/src/Http/Livewire/DeleteUserForm.php
@@ -26,7 +26,7 @@ class DeleteUserForm extends Component
     public $password = '';
 
     /**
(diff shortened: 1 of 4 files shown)


src/Http/Livewire/DeleteUserForm.php
```diff
@@ -16,6 +18,27 @@ class DeleteUserForm extends Component
...
 
+    /**
+     * The user's current password.
+     *
+     * @var string
+     */
+    public $password = '';
+
+    /**
+     * Confirm that the user would like to logout from other browser sessions.
```
Howelley, line 29: Confirm that the user would like to **delete their account**.

SeanBourke: Good pickup, resolved.

PR #97, opened 2020-09-08
Add missing extension when publishing routes

arnordhaenens:
Add missing file extension when publishing jetstream routes


arnordhaenens:
@taylorotwell sorry to bother, but why was this reverted? Some further adjustments needed?

PR #109, opened 2020-09-09
Fix for vendor:publish routes

abdullahnaseer:
When publishing routes for jetstream, it copy routes file to "routes/jetstream" without .php extension.


abdullahnaseer:
After publishing routes, it dont use that new file. 
I think, more changes are required in JetstreamServiceProvider@configureRoutes

PR #111, opened 2020-09-09
Login by username

fhprietor:
I have an administration system that does not want its users to register, they must be created by an administrator, therefore it requires a modal dialog that allows adding a user.
Likewise, he would like access to the system not through email, but through a new field in users called username


lostdesign:
Uh this has no changes and wants to merge `jetstream/master` into `jetstream/1.x`.

The username can be changed in the config, registering can be disabled in the config. If you need to add people to the system by modal or other, implement it for your very app?

PR #114, opened 2020-09-09
Make team menu entry "truncate" (text-overflow)

leganz:
If you make use of an to long team name the text "overflows" the box.

eg:

Bildschirmfoto 2020-09-09 um 17 15 53

This PR would fix that by adding "truncate" class to the menu entry:

Bildschirmfoto 2020-09-09 um 17 26 08


taylorotwell:
For future reference please update the Inertia component as well.

PR #115, opened 2020-09-09
add missing --teams argument on livewire install

hebbet:


taylorotwell:
It's not missing. I'm just showing both examples.

Issue #118, opened 2020-09-09
Do not hardcore model names


supertassu:
Currently Jetstream.php has multiple hardcoded class names:

https://github.com/laravel/jetstream/blob/89d7f8b174cd2d4eac4f063492f9db64fc935554/src/Jetstream.php#L46
https://github.com/laravel/jetstream/blob/89d7f8b174cd2d4eac4f063492f9db64fc935554/src/Jetstream.php#L230
https://github.com/laravel/jetstream/blob/89d7f8b174cd2d4eac4f063492f9db64fc935554/src/Jetstream.php#L252

Could those be made configurable so the models can be located somewhere else (in a subdirectory, for example)? Thanks!

mallardduck:
@supertassu - I think you're overlooking the easiest way to address this issue. Extend and overwrite. There's nothing forcing you to use the `Laravel\Jetstream\Jetstream` class without customizing it and adjusting the behavior.

```
<?php


namespace App;

use App\SomeOther\Folder\User;

class MyJetstream extends \Laravel\Jetstream\Jetstream
{
    public static $userModel = User::class;

    /**
     * Get the name of the team model used by the application.
     *
     * @return string
     */
    public static function teamModel()
    {
        return 'App\\SomeOther\\Folder\\Team';
    }

    /**
     * Get the name of the team model used by the application.
     *
     * @return string
     */
    public static function membershipModel()
    {
        return 'App\\SomeOther\\Folder\\Membership';
    }
}
```

This will work just fine and all you need to change from here is to adjust your Apps `JetstreamServiceProvider` to use this implementation instead of the OOTB version shipped with the package.

mallardduck:
@supertassu - now that #121 is merged in this should all be even easier to pull off.

I overlooked that a `useUserModel` method existed earlier, so between #119 and #121 this issue should be moot and can be set solved/closed.

Now you can simply do:

```
        Jetstream::useUserModel(App\MyStuf\Models\User::class);
        Jetstream::useTeamModel(App\MyStuf\Models\Team::class);
        Jetstream::useMembershipModel(App\MyStuf\Models\Membership::class);
```

... within the `boot` method of your apps `JetstreamServiceProvider`.

supertassu:
Thank you @mallardduck!

PR #119, opened 2020-09-09
Use static properties for Team and Membership too

mallardduck:
Keeping some parody with the way the User model is managed by the Jetstream service, the Team and Membership models are now treated the same way.

This was inspired by #118 since that made me realize editing models location wouldn't be done in a consistent manner. So if all you wanted to do was point models to a different location you'd be overriding properties and methods. So by exposing all as static properties, now you only would add properties to your extension of Jetstream.

---

# What extending Jetstream to adjust models looks like:

## Before:
```
<?php


namespace App;

use App\SomeOther\Folder\User;

class MyJetstream extends \Laravel\Jetstream\Jetstream
{
    public static $userModel = User::class;

    /**
     * Get the name of the team model used by the application.
     *
     * @return string
     */
    public static function teamModel()
    {
        return 'App\\SomeOther\\Folder\\Team';
    }

    /**
     * Get the name of the team model used by the application.
     *
     * @return string
     */
    public static function membershipModel()
    {
        return 'App\\SomeOther\\Folder\\Membership';
    }
}
```

## After
```
<?php


namespace App;

use App\SomeOther\Folder\User;
use App\SomeOther\Folder\Team;
use App\SomeOther\Folder\Membership;

class MyJetstream extends \Laravel\Jetstream\Jetstream
{
    public static $userModel = User::class;
    public static $teamModel = Team::class;
    public static $membershipModel = Membership::class;
}
```

### DIFF
--- a/src/Jetstream.php
+++ b/src/Jetstream.php
@@ -46,6 +46,20 @@ class Jetstream
     public static $userModel = 'App\\Models\\User';
 
     /**
+     * The team model that should be used by Jetstream.
+     *
+     * @var string
+     */
+    public static $teamModel = 'App\\Models\\Team';
+
+    /**
+     * The membership model that should be used by Jetstream.
+     *
+     * @var string
+     */
+    public static $membershipModel = 'App\\Models\\Membership';
+
+    /**
      * Determine if Jetstream has registered roles.
      *
      * @return bool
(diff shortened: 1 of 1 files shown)


taylorotwell:
Thanks but we should also add methods to set these without having to extend the class.

mallardduck:
@taylorotwell - good call, I overlooked that existing already for User.
Will send over an additional PR soon.

mallardduck:
@taylorotwell if I were to document the `use*Model` methods somewhere in the docs repo, where do you think makes more sense?

1. Getting Started > Configuration
2. Customization > Models

Issue #99, opened 2020-09-08
Changes to Profile Name or Team Name not immediately reflected in menu...


telkins:
- Jetstream Version: 1.0.0
- Laravel Version: 8.0.0
- PHP Version: 7.4.9
- Database Driver & Version: n/a

### Description:
When changing the team name or profile name, neither the team name in the drop-down menu nor the avatar (initials) reflect the changes until a refresh is made.

**This probably isn't a big deal**, but I thought I'd point it out. 🤓 

### Steps To Reproduce:
Install new app: `laravel new l8app --jet --teams --stack=livewire`
Build front-end: `npm install && npm run dev`
Visit site
Register
Go to Team Settings or Profile
Change team name or name
Notice that the user avatar (initials) or drop-down menu don't update to reflect the changes

mallardduck:
Seems like the easiest fix would be to force those actions to trigger a complete page reload. I wonder if Livewire provides functionality for that OOTB, or if they have another solution to this issue to force other elements on page update.

stevebauman:
>  I wonder if Livewire provides functionality for that OOTB

I think we could use the `$this->emit()` [feature](https://laravel-livewire.com/docs/2.x/events#firing-events) in Livewire to emit an update event to other components that an account update has occurred and to refresh themselves?

mallardduck:
Ahh @stevebauman great point!

I only just started working with livewire a few weeks back, so overlooked that.
Great idea that should be a fancier UX than doing a full reload.

calebporzio:
I proposed a solution in this PR: #132

telkins:
> I proposed a solution in this PR: #132

Thx...!  🤓 

I thought the PR made good sense, but I also suggested changing the event name to instead say what's happening vs what should happen. 😜

PR #125, opened 2020-09-10
[1.x] Livewire : update header avatar on profile updated

tonyyb:
According to : https://github.com/laravel/jetstream/issues/99

Now an event is dispatched to browser from Livewire to get a window event `profile-updated`. 

Catch it from Alpine JS and update avatar infos of header (with a new Alpine component).


calebporzio:
Good PR, although I personally would prefer a backend-centric solution using Livewire's event system.

I opened a PR here: #99

driesvints:
Closing in favor of Caleb's pr.

PR #129, opened 2020-09-10
[1.x] Inertia: Use check instead of authorize for canCreateTeams

### LINKED ISSUE #113: Returning false from TeamPolicy->create causes 403 error on all pages when user is logged in
- Jetstream Version: 1.0.2
- Laravel Version: 8.0.1
- PHP Version: 7.4.7
- Database Driver & Version: mysql  Ver 14.14 Distrib 5.7.29, for osx10.15
- Stack: Inertia

### Description:
I wanted to limit the creation of teams, for example only when the user has an active subscription. But, when returning false from `App\Policies\TeamPolicy->create` all pages respond with a 403 when the user is logged in.

### Steps To Reproduce:
- Create a new Laravel + Jetstream project using `laravel new saas --jet --stack=inertia --teams`
- Update mysql credentials in `.env`
- Create a user by navigating to http://saas.test/register
- Change the return statement in `App\Policies\TeamPolicy->create` to false.
- Refresh the dashboard or navigate to another page, now a 403 error will be shown

RobertBoes:
The authorize call in ShareInertiaData causes an AuthorizationException, this means the data is not shared with Inertia. Changing authorize to check returns the boolean value and will populate canCreateTeams correctly when TeamPolicy->create returns false.

This closes #113

### DIFF
--- a/src/Http/Middleware/ShareInertiaData.php
+++ b/src/Http/Middleware/ShareInertiaData.php
@@ -25,7 +25,7 @@ class ShareInertiaData
                 return [
                     'canCreateTeams' => $request->user() &&
                                         Jetstream::hasTeamFeatures() &&
-                                        Gate::forUser($request->user())->authorize('create', Jetstream::newTeamModel()),
+                                        Gate::forUser($request->user())->check('create', Jetstream::newTeamModel()),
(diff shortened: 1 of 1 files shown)


lostdesign:
Could reproduce the behavior and confirm it works :D Not that my comment matters but still wanted to comment. 
image

PR #132, opened 2020-09-10
[Livewire] Refresh navigation dropdown when teams/profile forms save

calebporzio:
This is in response to issue #99 

## The Problem:
* Updating the "name" in the profile form doesn't update the placeholder avatar in the navigation dropdown until a page refresh
* Updating the "team name" in the team form doesn't update the "team name" in the navigation dropdown until a page refresh

## Potential Solutions:

A) Close this PR and do nothing. This is a minor UI flaw, I wouldn't even call it a bug.
B) Close this PR and instead, force a full page refresh EVERY time one of these fields are updated. This is currently being done when a new profile photo is uploaded: [Source code reference](https://github.com/laravel/jetstream/blob/29a7d03dd88cda946592ea6f96ab1efb31d00a0a/src/Http/Livewire/UpdateProfileInformationForm.php#L56)
C) Merge this PR. Now saving either of those forms will emit a Livewire event that the dropdown is listening to so it will refresh anytime they are updated.

## PR Overview:
* Extract the navigation dropdown from `layouts.app` to its own dedicated Livewire component (`NavigationDropdown`)
* By doing so we can now listen for a `refresh-navigation-dropdown` event to be emitted by other Livewire components and do a simple re-render when one occurs
* Emit a `refresh-navigation-dropdown` event from both the profile and team name form components

## Breaking Changes:
This PR is safe to merge and its changes are completely compatible with the existing stubs.


telkins:
@calebporzio I like it...and had considered suggesting that one might experience the same "issue" when changing the user's profile photo.

I agree that it's not really a bug...but it would definitely be nice to have, I think.

Regarding your solution (I've only read the PR Overview bullets), I have only one complaint...the name of the event implies that it knows what needs to be done.  I think it would be nicer to name it something like `teams-or-profile-changed`...or even have more than one event: `team-changes` and `profile-changes`.

Whatever the case, I think the event(s) should say what has happened vs what needs to happen.

Thoughts...?

GabrielFemi:
> @calebporzio I like it...and had considered suggesting that one might experience the same "issue" when changing the user's profile photo.
> 
> I agree that it's not really a bug...but it would definitely be nice to have, I think.
> 
> Regarding your solution (I've only read the PR Overview bullets), I have only one complaint...the name of the event implies that it knows what needs to be done. I think it would be nicer to name it something like `teams-or-profile-changed`...or even have more than one event: `team-changes` and `profile-changes`.
> 
> Whatever the case, I think the event(s) should say what has happened vs what needs to happen.
> 
> Thoughts...?

Yeah... I totally agree.

telkins:
Happy for the PR's functionality.... 👍 Not so crazy about the event name....  👎

PR #133, opened 2020-09-10
Get correct Http kernel filename & User model filename

ahmadmayahi:
The Http Kernel could be in a different location than the default one, so the following line is incorrect:

```php
app_path('Http/Kernel.php')
```

In one of our projects, the Http kernel is located in the `src/App/HttpKernel.php`:
```php
$app->singleton(
    Illuminate\Contracts\Http\Kernel::class,
    App\HttpKernel::class
);
```

So, installing jetstream in such projects won't be possible due to the location path.

The same applies to the `User` model, which might be in different folder than the `App/Models`.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -499,7 +499,7 @@ EOF;
     }
 
     /**
-     * Get Http kernel file name
+     * Get Http kernel file name.
      *
      * @return  false|string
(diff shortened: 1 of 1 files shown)


taylorotwell:
The install command is meant to only be used on fresh Laravel 8 applications, as noted in the documentation.

PR #142, opened 2020-09-10
Add includes file for session messages

GabrielFemi:
This removes the repetitive session message used in views.
I just added a file that is included in any view.


taylorotwell:
If I were going to extract this I would maybe make it a component instead of an include, although I'm not even sure if that's necessary.

GabrielFemi:
> If I were going to extract this I would maybe make it a component instead of an include, although I'm not even sure if that's necessary.

Okay... thanks I learnt something new

PR #153, opened 2020-09-11
best practice for use key v-bind every v-for

albasyir:
at AppLayout.vue, for `v-for` props for teams list, we need add key binding like my change..


taylorotwell:
Please describe what is actually broken.

PR #164, opened 2020-09-12
Fix CreateNewUserWithTeams class name

chrysanthos:
The name of the class should match the filename but this one was wrong possibly from copy-pasting the actual CreateNewUser class.


chrysanthos:
My bad, the name is changed during installation depending what you chosed.

Issue #162, opened 2020-09-12
Jetstream / Teams: Column not found: 1054 Unknown column 'current_team_id' in 'field list'


McPan:
- Jetstream Version: 1.1.2
- Jetstream Stack: Livewire
- Laravel Version: 8.1.0
- PHP Version: 7.4.10
- Database Driver & Version: mysql / mariadb 10.4.14

### Description: Exeception Error after registering a new user and redirect to dashboard

ErrorException
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'current_team_id' in 'field list' (SQL: update `users` set `current_team_id` = 1, `users`.`updated_at` = 2020-09-12 09:07:10 where `id` = 1) (View: <PATH_TO_PROJECT>\resources\views\layouts\app.blade.php) (View: <PATH_TO_PROJECT>\resources\views\layouts\app.blade.php) 

Described column 'current_team_id' is missing in the 'users' table.

### Steps To Reproduce:

Create a new Laravel 8 project

```
composer create-project --prefer-dist laravel/laravel .\
```

Include 'jetstream' package

```
composer require laravel/jetstream
```

Create and run migrations

```
php artisan jetstream:install livewire --teams
npm install && npm run dev
php artisan migrate
```

Open the website of the new project and register a new user. Redirecting to the 'dashboard' will throw the exception specified above.

### (Temporary) solution

Create and run another migration

```php
<?php

use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

class AddCurrentTeamIdToUsersTable extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        Schema::table('users', function (Blueprint $table) {
            $table->foreignId('current_team_id')
                    ->after('remember_token')
                    ->nullable();
        });
    }

    /**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
        Schema::table('users', function (Blueprint $table) {
            $table->dropColumn('current_team_id');
        });
    }
} 
```

Try to register a new user and everything works fine.

dotancohen:
I've got an open [pull request](https://github.com/laravel/jetstream/pull/166) to resolve this issue, but a change to the build process must be undertaken first.

taylorotwell:
Re-migrate your database after installing Jetstream.

DulbyMendez:
[...]

PR #166, opened 2020-09-12
Convert migration overwrite to a new migration. Resolve Issue #162

dotancohen:

### DIFF
--- a/database/migrations/2020_09_12_000000_update_users_table.php
+++ b/database/migrations/2020_09_12_000000_update_users_table.php
@@ -4,7 +4,7 @@ use Illuminate\Database\Migrations\Migration;
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Facades\Schema;
 
-class CreateUsersTable extends Migration
+class UpdateUsersTable extends Migration
 {
     /**
      * Run the migrations.
@@ -13,16 +13,9 @@ class CreateUsersTable extends Migration
      */
     public function up()
     {
-        Schema::create('users', function (Blueprint $table) {
-            $table->id();
-            $table->string('name');
-            $table->string('email')->unique();
-            $table->timestamp('email_verified_at')->nullable();
-            $table->string('password');
-            $table->rememberToken();
+        Schema::table('users', function (Blueprint $table) {
             $table->string('current_team_id')->nullable();
             $table->text('profile_photo_path')->nullable();
-            $table->timestamps();
         });
     }
 
@@ -33,6 +26,8 @@ class CreateUsersTable extends Migration
      */
     public function down()
     {
-        Schema::dropIfExists('users');
+        Schema::table('users', function (Blueprint $table) {
+            $table->dropColumn(['current_team_id', 'profile_photo_path']);
+        });
     }
 }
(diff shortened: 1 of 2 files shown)


dotancohen:
The test seems to be failing because the build process has changed. Previously, it was **forbidden** to run the initial migration before installing Jetstream. With this change, it is **required** to run the initial migration before installing Jetstream.

@taylorotwell Can this build process be changed? I don't see where I can contribute to that.

taylorotwell:
This change isn't really necessary. You can just `php artisan migrate:fresh` if you run your migrations before installing Jetstream.

dotancohen:
@taylorotwell Thank you. Though your are correct about the workaround, that add burden to the developer. I'll be honest, In three years of Laravel I've never had to use `migrate:fresh` and it seemed, still seems, like a workaround for when problems arise. This issue is an example of a problem that doesn't have to exist.

In fact, the idea that prior migration files could be overwritten seems completely backwards. Why have migrations instead of a canonical CREATE TABLE definition, if they are just going to be overwritten anyway?

From a [prior bug](https://github.com/laravel/jetstream/issues/162) and a [discussion online](https://www.reddit.com/r/laravel/comments/irethy/column_missing_in_database/) it seems to me that a significant part of the community of Laravel coders agree that migrations should not be overwritten.

Of course I defer to your experience but do bear in mind that a majority of Laravel's target audience, us developers, would like to see this change implemented. Thank you!

nabeelio:
> In fact, the idea that prior migration files could be overwritten seems completely backwards. Why have migrations instead of a canonical CREATE TABLE definition, if they are just going to be overwritten anyway?

Yeah this seems brittle. If you're distributing something and you modify the original migration files, do your originals get overwritten? Or if it's an in-prod app that's being migrated? Is that what could happen here?

driesvints:
You'll need to create a new migration. **Jetstream is meant to be used with new applications.** We cannot know what kind of migrations or table modifications you've already done on your users table so you'll need to figure out which changes you need yourself if you want to install jetstream on an existing app.

dotancohen:
> We cannot know what kind of migrations or table modifications you've already done on your users table

If that is the reason, then overwriting the `2014_10_12_000000_create_users_table.php` file is doubly backwards. You are establishing that Laravel packages can overwrite extant migration files if they are "meant to be used with new applications". What happens when the dev installs two such packages? One of them will get it's overwriting migration file overwritten!

The purpose of having migrations is to prevent these issues. Whoever authors it does not need to know what other modifications have been made to the table. Declare the fields required and move on.

driesvints:
Hey @dotancohen. I think you're right about overwriting the migration. Perhaps we can do better and throw a warning to the user? Can you maybe send in a PR for that?

dotancohen:
@driesvints 
Sure, no problem. How would you like it to warn? I'm thinking if the users table already exists, just throw an exception.

driesvints:
@dotancohen I think we'd also have to take in mind the behavior of `--force` to skip the warning. Instead of an exception I'd recommend using `$this->warn(`. Maybe a question to confirm overwriting?

dotancohen:
@driesvints Sure, no problem. Thank you.

Just to clarify, if `--force` is used, then drop the current table and install the new table.

driesvints:
Yeah, force should overwrite any "decisions" that need to be taken.

PR #176, opened 2020-09-13
[1.x] Fix 'laravel new project-name --jet'

antonioribeiro:
If you create an app using

```
laravel new project-name --jet
cd project-name
valet link
```

And try to open the page right away, this is what you will see:

```
Illuminate\Database\QueryException
SQLSTATE[HY000] [2002] Connection refused (SQL: select * from `sessions` where `id` = fZXzv4NDHB0ZxfUwdQV5Z4FmAuIdjEMyOy7zKwop limit 1)
```

Because you probably don't have a database configured. This PR fix this by only configuring the your `.env` if you actually have a database connection configured. So new users will still see the new Laravel app page.


taylorotwell:
I think it's fine as is for now.

PR #182, opened 2020-09-14
[1.x] Add MustVerifyEmail Interface to User Model

amayer5125:
The stub for the User model is missing the `Illuminate\Contracts\Auth\MustVerifyEmail` interface which is in the default [Laravel project](https://github.com/laravel/laravel/blob/master/app/Models/User.php#L5) and mentioned in the [Jetstream docs](https://jetstream.laravel.com/1.x/features/authentication.html#email-verification) as already being present.

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,6 +1,6 @@
 php:
   preset: laravel
   disabled:
-    - unused_use
+    - no_unused_imports
 js: true
 css: true
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,4 +1,6 @@
 php:
   preset: laravel
+  disabled:
+    - unused_use
 js: true
 css: true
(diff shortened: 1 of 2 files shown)


alxmtr:
As email verification is disabled by default, adding the `Illuminate\Contracts\Auth\MustVerifyEmail` interface to the User model will throw a `RouteNotFoundException` (Route [verification.notice] not defined) when accessing the dashboard.

This will force us to enable the feature in `config/fortify.php` on a fresh install or remove the interface from the User model if we don't want to use email verification.

.styleci.yml
```diff
@@ -1,4 +1,6 @@
 php:
   preset: laravel
+  disabled:
+    - unused_use
```
GrahamCampbell, line 4: Should be `no_unused_imports`.

driesvints:
@amayer5125 can you rebase and fix the conflicts?

GrahamCampbell (changes requested):

PR #190, opened 2020-09-14
[2.x] Remove the registersRoutes variable.

KaniRobinson:
After comments from https://github.com/laravel/sanctum/pull/194. 
I have created this PR to move the way the routes get pulled in to follow the current sanctum flow.


driesvints:
Same reason as I said in the fortify pr: not worth the breaking change to me. The current way is in line with other 1st party packages like Cashier, Passport etc.

taylorotwell:
Yeah I didn't mean remove any existing functionality... I meant you could just add support for the config if it is present.

PR #198, opened 2020-09-14
Preserve the current URL when changing teams in InertiaJS

securit:
This change preserves the current URL of the browser when switching teams so that you don't always return to the Dashboard.vue, but rather the page on which you were on when the team was switched.


taylorotwell:
That is not going to be practical. What if they are viewing a page that has the ID of a particular team's model in the URL?

PR #210, opened 2020-09-16
[1.x] Ensure models directory exists

rojtjo:
I noticed while attempting to install Jetstream in an existing project that it crashed halfway through because I didn't have a models directory.
This change ensures the directory exists.


taylorotwell:
All new 8.x projects already have a models directory.

rojtjo:
True, the problem only arises when upgrading from an older version, or when you decide to install Jetstream later after already having moved the `User` model to a different directory (and obviously removing the `Models/` directory).

driesvints:
@rojtjo Jetstream is meant for new Laravel installs. If you're migrating your existing app to Jetstream you'll need to do some manual work.

rojtjo:
Yeah, I figured already! Just assumed (when installing the first time) that it was checking if a models directory exists like it does for the `make:model` command. 😅

Anyways, thanks for the explanation!

PR #222, opened 2020-09-18
[1.x] Add option to publish assets without composer+npn

chrysanthos:
Since most of JetStream files are published into our own apps, it is a good practice for us to republish these files every now and then so we are up to date.

This will allow us to run `jetstream:install` on existing projects with the `--no-install` option and skip the composer and npm updates since the packages should already be installed. 

This is a new backwards compatible feature.


taylorotwell:
Once you install Jetstream it is my opinion that you are now responsible for maintaining the frontend. It will not be possible to merge any future updates with your potential changes.

PR #225, opened 2020-09-18
[1.x] Allow full access to ui-avatars api

chrysanthos:
This PR exposes the manipulation of the default profile photo through the config.

After merging, we will be able to change the default text color and background color for the default profile photo but most importantly we will be able to specify any [ui-avatar](https://ui-avatars.com/) parameter we want.


taylorotwell:
Just override the method on your model.

PR #227, opened 2020-09-18
[1.x] Add a Current Team area next to the settings dropdown

### LINKED ISSUE #223: Idea: Displaying currentTeam if having multiple teams?
I am working on two apps where I have the same issue regarding reminding myself about what is my currently active team.

Wouldn't it be a nice to have a display of the currently active team on the left of the user avatar, if the current user has more than one team?

I have built this on both of my Jetstream projects so far and have a feeling that I am going to build it many more times in the future :) Would it be something that you think is a good idea to make a PR for?

jimmitjoo:
This PR adds an indicator of which team that is currently active for a logged in user that has more than one team. If the user only has one team, there are nothing displayed.

Resolves #223

### DIFF
--- a/stubs/livewire/resources/views/navigation-dropdown.blade.php
+++ b/stubs/livewire/resources/views/navigation-dropdown.blade.php
@@ -18,8 +18,20 @@
                 </div>
             </div>
 
-            <!-- Settings Dropdown -->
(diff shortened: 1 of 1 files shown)


taylorotwell:
You can add this to your own app if desired 🙂

PR #237, opened 2020-09-20
Show password validation errors on reset password page

ali-alharthi:
Jetstream uses Fortify's `ResetUserPassword` class to reset the user password that returns an error-bag called `updatePassword`. 

Therefore, no validation errors on the reset-password page.

There are other ways that could fix this using the existing components; However, it seemed more appropriate to make a component to return all the errors within an error-bag.


taylorotwell:
I think this is a bag and ResetPassword should not use a custom bag at all.

PR #251, opened 2020-09-22
[1.x] Added hasTeamRole check on the HasTeams trait.

wotta:
I have added this check because we do have a `hasTeamPermission` function but I missed a function to check if the team has a specific role.
When I checked the trait I saw the following function `teamRole` which returns the role object.
So when I saw that I wanted to add a function that returns true or false based on the fact if the role actually exists.

I would love to read your opinion on this pull request.

### DIFF
--- a/src/HasTeams.php
+++ b/src/HasTeams.php
@@ -116,6 +116,34 @@ trait HasTeams
             'id', $this->id
         )->first()->membership->role);
     }
+   
+    /**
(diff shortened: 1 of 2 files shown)


wotta:
Thanks for merging🙂

driesvints:
@wotta can you also send in a PR to the docs for this method?

wotta:
Sure thing. Mind if I do that tonight?

Issue #252, opened 2020-09-22
Make users.current_team_id an integer...?


telkins:
I just noticed that `users.current_team_id` is a string.  Is there a reason for this?  Wouldn't it be better to make it an integer...?

I might be missing something.  Admittedly, I didn't look into the code to see if it uses `teams.name` or something for that field.  My assumption is that it would use `teams.id`, which is a big integer, afaik.

So...I'm happy to put together a PR, but just wanted to get some background information in case I'm off-base on this. 🤓 

Thx...!  👍

driesvints:
Yeah I think that's a leftover from when we were still using uuid's. Feel free to send in a PR 👍 

Thanks

PR #253, opened 2020-09-22
Changed current_team_id to big integer.

telkins:
Please see brief conversation between myself and @driesvints : #252 

Also...I'm guessing it's best to **not** introduce foreign key constraints here...?  Personally, I would like to see them, but I'm guessing the idea would be to leave them out for now and let individual developers add them as they see fit...?

Tests passed locally.

If I'm missing anything, please let me know. 🤓

### DIFF
--- a/database/migrations/2014_10_12_000000_create_users_table.php
+++ b/database/migrations/2014_10_12_000000_create_users_table.php
@@ -20,7 +20,7 @@ class CreateUsersTable extends Migration
             $table->timestamp('email_verified_at')->nullable();
             $table->string('password');
             $table->rememberToken();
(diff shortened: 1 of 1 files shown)


driesvints:
I think this can be sent to 1.x

taylorotwell:
Yeah this should be sent to 1.x I believe. Thanks.

telkins:
@driesvints @taylorotwell Hey...not sure what's going on exactly.  It looks like I created the PR to merge against master, but it should have been against the 1.x branch.  So, I just resubmitted (#254).

I hope this is OK.  I'm not trying to make extra noise.  😛  If I need to (not) do something, please let me know.  Thx.  🤓

PR #254, opened 2020-09-22
[1.x] Changed current_team_id to big integer.

telkins:
(Retrying #253, which was previously attempting to merge against master.)

Please see brief conversation between myself and @driesvints : #252 

Also...I'm guessing it's best to **not** introduce foreign key constraints here...?  Personally, I would like to see them, but I'm guessing the idea would be to leave them out for now and let individual developers add them as they see fit...?

Tests passed locally.

If I'm missing anything, please let me know. 🤓

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,7 +242,7 @@ Jetstream::role('editor', 'Editor', [
 <a name="team-authorization"></a>
 ### Authorization
 
-A user's team permissions may be determine using the `hasTeamPermission` method available via the `Laravel\Jetstream\HasTeams` trait:
+A user's team permissions may be determined using the `hasTeamPermission` method available via the `Laravel\Jetstream\HasTeams` trait:
 
 ```php
 if ($request->user()->hasTeamPermission($team, 'read')) {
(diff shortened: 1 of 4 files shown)


driesvints:
Seems like some things went wrong here. I think it's best that you try again by branching off fresh from 1.x and recommitting.

taylorotwell:
I can fix it.

taylorotwell:
Hmm yeah Dries is right probably need to branch off 1.x and recommit.

telkins:
> Seems like some things went wrong here. I think it's best that you try again by branching off fresh from 1.x and recommitting.

Yep...didn't think about that.  But, looks like @taylorotwell is on it...so I'll wait to see what shakes loose. 🤓 

Oh....looks like I should do it, anyway.  No worries....third time's the charm, right...?  😆

telkins:
@taylorotwell or @driesvints  Hey...can you please confirm for me that you've just made this change?  I pulled the latest 1.x locally and it was already changed.  I just noticed that it's also change when viewing the source via github.  Thx.

driesvints:
@telkins seems like Taylor has indeed fixed this already: https://github.com/laravel/jetstream/commit/1883071d771193e59ef7ed4509576e9b853dac43

Thanks for notifying us 👍

Issue #258, opened 2020-09-23
Ability to override fetching roles and fetching permissions functionality


wprk:
Not sure if it's possible and I've just not been able to work it out but in my application I have a `permissions` and a `team_role_permissions` table. I'd like to be able to overload the `hasRoles`, `findRole`, `hasPermissions` and `permissions` methods used by JetStream to custom ones enabling me to populate team roles and permissions from the database.

I can obviously call something like: `Jetstream::permissions(TeamPermission::all()->toArray());` to get all my permissions into the application but I don't have access to the currentTeam in order to loop over the current teams roles and call something like the below in my JetstreamServiceProvider.

```
    $roles = TeamRole::whereTeamId(Auth::user()->currentTeam->id)->get();

    foreach($roles in $role) {
        Jetstream::role(
            $role->key
            $role->label,
            $role->permissions->toArray()
        )->description($role->description);
    }
```

Am I missing something or is this functionality not available yet? Would you be open to a PR?

wprk:
Closing as PR closed.

PR #259, opened 2020-09-23
Allow for permissions and roles to be fetched from database models

wprk:
As proposed in Issue #258 I think it would be useful to have the ability to optionally pull permissions and roles from database tables instead.

I've had a quick go at how that might look. This would obviously need more work before it could be approved. Not everyone would want this functionality and it clutters up the Jetsteam.php class somewhat so would be interested in your thoughts on how we could optionally include this functionality. Possibly add it behind a `--database-roles-and-permissions' flag similar to the way you can optionally include teams.

Happy to continue working on this but didn't want to get too far into it before you've had a chance to look over issue 258 and decide if this is something you're interested in.

### DIFF
--- a/database/migrations/2020_09_23_220000_create_permissions_table.php
+++ b/database/migrations/2020_09_23_220000_create_permissions_table.php
@@ -1,0 +1,33 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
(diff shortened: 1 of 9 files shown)


taylorotwell:
No plans to add this right now.

PR #261, opened 2020-09-24
[1.x] Allow teams uri customization

stvnrlnd:
This pull request contains a single commit that allows customization of the `/teams/` URI.

This was a feature in Laravel Spark that was useful for route localization or simply to present the Teams feature differently in the URL. In my case, it makes more sense for each user to have a personal workspace and shared workspaces, rather than a personal team and shared teams. In many view files this can already be done by just changing the word Team to Workspace, this extends that effort to URIs as well.

The default terminology remains `teams`, users that want to change that should do so exactly like they would in Laravel Spark. In the JeststreamServiceProvider, you can now add:

```
Jetstream::prefixTeamsAs('bands');
```

I hope this is helpful, thanks!

### DIFF
--- a/routes/inertia.php
+++ b/routes/inertia.php
@@ -36,15 +36,15 @@ Route::group(['middleware' => config('jetstream.middleware', ['web'])], function
 
         // Teams...
         if (Jetstream::hasTeamFeatures()) {
-            Route::get('/teams/create', [TeamController::class, 'create'])->name('teams.create');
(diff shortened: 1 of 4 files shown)


driesvints:
You can already publish the routes so you can overwrite them so I don't think this is necessary.

taylorotwell:
All the route names would still refer to teams, etc.

PR #266, opened 2020-09-25
[1.x] Disable save button while profile pic is uploading

octoper:
This pull request adds disable attribute to save button while profile picture is uploading.


driesvints:
@octoper you might want to explain why this is needed.

octoper:
This is needed because in slow connections like 3G, when the user clicks the save button before the profile pic is uploaded the page refreshes and the picture will not be uploaded.

Issue #274, opened 2020-09-27
User Navigation Menu Dropdown Styling


amayer5125:
When `Features::profilePhotos()` is disabled the styling of the user's name in the navigation bar doesn't look very good. I believe this is because the image alt text is what is being rendered. It has a gray circle around the text name that usually surrounds the profile image.
![image](https://user-images.githubusercontent.com/3212673/94353069-c77ea100-003a-11eb-9b32-e900b870b3b0.png)

When profile photos are disabled the name should be shown on one line possibly with a chevron beside it to denote that the name has a submenu.

driesvints:
Seems like you already attempted a pr 👍

PR #275, opened 2020-09-27
[1.x] Add Submenu Style For When Profile Photos Are Unmanaged

amayer5125:
Display user name nicer when `Features::profilePhotos()` is not enabled.

Styling matches inactive navigation items.
image

Upon hover or clicking, styling matches active state of other navigation items.
image


driesvints:
See https://github.com/laravel/jetstream/issues/274

PR #292, opened 2020-10-01
[1.x] Add show toggle to password field

Chrissi2812:
This will add a simple toggle to show the password using Alpine.js.
password_toggle


Larsklopstra:
Is the button focus-able? It looks like the on click handler is placed on a `div` rather than a `button`

taylorotwell:
Along with the concern noted above, I wonder if this would make more sense as a component? That way it could be re-used... some sort of `<x-jet-revealable-password>` component, etc.

PR #293, opened 2020-10-01
[1.x] Add reveal-password component

Chrissi2812:
This is a follow up on #292 it will ad a new blade component `<x-jet-revealable-password>`, and a new vue component as well.

I have replaced most usage of the old password inputs except for password and password confirmation dialogs. 
Otherwise you could simply copy&paste the password from one field to the other.

### DIFF
--- a/resources/views/components/revealable-password.blade.php
+++ b/resources/views/components/revealable-password.blade.php
@@ -1,0 +1,17 @@
+@props([
+    'class',
+    'disabled' => false,
+])
+
+<div class="{{ "relative flex items-center $class" }}" x-data="{ showPassword: false }">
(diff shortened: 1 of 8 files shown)


taylorotwell:
Honestly think I may hold off on this. Not sure everyone would want it.

emmadesilva:
> Honestly think I may hold off on this. Not sure everyone would want it.

@taylorotwell Would you be open to reconsidering this? It's a very good "best practise" for accessibility, and improves the UX of the form a lot. https://web.dev/articles/sign-in-form-best-practices#show-password

I think the most common scenario where one would not want this is if they want a "cleaner" UI, and for those edge cases they can just modify the component. I don't think disabling this is a common enough desire for it to warrant the complexity of a config option. But I do think it would improve both Jetstream and Breeze to have this.

PR #298, opened 2020-10-02
[2.x] Inertia.js: Add support for all Jetstream-provided views.

claudiodekker:
This PR adds Inertia variants for all Jetstream included views. This allows a more native Inertia experience for users of this stack, and solves several issues and difficulties that makes for an all-in-all smoother experience.

For example, some Jetstream users have been asking questions on our Discord regarding difficulties they're experiencing with getting Jetstream to play well with Inertia. Specifically, the most common issue is when a session expires, as the user then gets redirected to the login page, and because that login page is a Blade view, Inertia renders it in a so-called 'error modal'.

To solve or prevent this, users should instead use Inertia for their login pages as well, and while some have gotten to the point where they're attempting to (re)configure Fortify to use Inertia views instead, this too isn't the most straight-forward thing to do.


reinink:
@taylorotwell For what it's worth, I definitely think this is the right approach to go. The best Inertia.js experience is to do your whole app with Vue based page components. Session expiration "just works" this way, since users are automatically redirected back to the login page, which itself is an Inertia endpoint. I don't think I'd ever, personally, build an app that had a mix of Blade views and Vue page components. That said, this does create an extra maintenance burden on this project, since these auth views can no longer be shared between the two stacks. I feel like that's a worthwhile tradeoff though. 🤷‍♂️ 

That all said, the session expiration issue could be solved using an `Inertia::location($url)` (409) response that's called in the authentication middleware. This would force a "full page redirect" to the login page. To me this is more of a band-aid, and I'd again prefer the solution in this PR. 👍

@claudiodekker Excellent work brother, this looks great. 🙌

claudiodekker:
Thanks! 

On a side-note (just so that I don't forget), Once https://github.com/laravel/fortify/pull/107 gets tagged, I can remove the `->toResponse($request)` calls from this PR as well. Since it hasn't yet, I've decided to leave them in for now.

reinink:
Yup, and while we're creating todo lists, it would be great to get `@inertiajs/inertia` on `v0.3.0`, and then update all the Jetstream code to use the new event callbacks. When we make that change, we should add `@inertiajs/progress` to the project, since right now there isn't any loading indicators, since the NProgress CSS isn't included.

taylorotwell:
@claudiodekker I've just tagged a new laravel/fortify release.

claudiodekker:
@taylorotwell Thanks! I've updated the PR. I also updated the `composer.json` to reflect the minimum Fortify version to be `^1.6.1`, let me know if you'd rather leave that unchanged.

stubs/inertia/resources/js/Jetstream/AuthenticationCardLogo.vue
```diff
@@ -0,0 +1,8 @@
+<template>
+    <a href="/">
+        <svg class="w-16 h-16" viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg">
```
m1guelpf, line 3: Couldn't we just reuse the `ApplicationMark` component here instead of duplicating?

claudiodekker: Yeah, I looked at that, but it's different, and because I'm trying to replicate the other views to keep branding consistent I copied this one instead.

taylorotwell:
@claudiodekker  Does this need to be rebased or something? It seems to contain some random other commits?

claudiodekker:
@taylorotwell Yeah, I merged 1.x into it.. I think that's not yet merged back in to master.

taylorotwell:
I've merged 1.x into master.

claudiodekker:
@taylorotwell Fixed, thanks! 👍

taylorotwell:
Sick. Will probably review and get this merged in tomorrow.

claudiodekker:
Cool, let me know if there's anything else you want me to do 👍

taylorotwell:
@claudiodekker One bug I've noticed so far... on two-factor code entry screen after login... entering the code and then hitting "enter" just toggles you to recovery code form instead of submitting auth code.

claudiodekker:
Oof, yeah, that shouldn't happen. 

I'll look at it in a bit, and will also double check the keyboard behavior of other forms.

taylorotwell:
That was the only thing I hit so far. I think I've tested every form.

claudiodekker:
Great, thanks for letting me know! 🙏 

Having dinner right now, so it might take an hour or so before I have a fix. Want me to ping you when it's done?

taylorotwell:
I think I'l have it fixed. Don't worry about it.

stubs/inertia/resources/js/Pages/Auth/TwoFactorChallenge.vue
```diff
@@ -0,0 +1,89 @@
...
+                recovery: false,
+                form: this.$inertia.form({
+                    code: '',
+                    recovery_code: '',
+                })
+            }
+        },
+        methods: {
+            toggleRecovery() {
+                this.recovery ^= true
```
m1guelpf, line 72:
```suggestion
                this.recovery = true
```

stubs/inertia/resources/js/Pages/Auth/TwoFactorChallenge.vue
```diff
@@ -0,0 +1,89 @@
...
+                <jet-label for="code" value="Code" />
+                <jet-input ref="code" id="code" type="text" class="mt-1 block w-full" v-model="form.code" autofocus autocomplete="one-time-code" />
+            </div>
+            <div v-else>
+                <jet-label for="recovery_code" value="Recovery Code" />
+                <jet-input ref="recovery_code" id="recovery_code" type="text" class="mt-1 block w-full" v-model="form.recovery_code" autocomplete="one-time-code" />
+            </div>
+
+            <div class="flex items-center justify-end mt-4">
+                <button @click.prevent="toggleRecovery" class="text-sm text-gray-600 hover:text-gray-900 underline cursor-pointer">
```
m1guelpf, line 29:
```suggestion
                <button type="button" @click.prevent="toggleRecovery" class="text-sm text-gray-600 hover:text-gray-900 underline cursor-pointer">
```

stubs/inertia/resources/js/Pages/Auth/TwoFactorChallenge.vue
```diff
@@ -0,0 +1,89 @@
...
+                    <template v-if="!recovery">
+                        Use a recovery code
+                    </template>
+                    <template v-else>
+                        Use an authentication code
+                    </template>
+                </button>
+                <jet-button class="ml-4" :class="{ 'opacity-25': form.processing }" :disabled="form.processing">
```
m1guelpf, line 37:
```suggestion
                <jet-button type="submit" class="ml-4" :class="{ 'opacity-25': form.processing }" :disabled="form.processing">
```

stubs/inertia/resources/js/Pages/Auth/TwoFactorChallenge.vue
taylorotwell: They are using this to toggle between 0 and 1.

stubs/inertia/resources/js/Pages/Auth/TwoFactorChallenge.vue
taylorotwell: Already the default type for jet buttons.

stubs/inertia/resources/js/Pages/Auth/TwoFactorChallenge.vue
m1guelpf: Wouldn't it be less confusing to use a boolean?

claudiodekker: This is actually correct. It's a bitwise XOR that changes true to false, and false to true in this situation (hence toggle) 👍

m1guelpf: That's actually a neat trick!

claudiodekker: For sure :) Also works in PHP ;)

PR #314, opened 2020-10-06
Use named routes on Inertia with Ziggy

### LINKED ISSUE #157: Use routes on Inertia views
Similar to how Livewire uses named routes instead of path strings, it would be awesome to have Inertia do the same so that changes to the paths on the published routes or adding a prefix to Fortify routes doesn't require a change to the views.

To accomplish this, we could use [Tighten's Ziggy package](https://github.com/tightenco/ziggy), which I believe is already the standard practice for using routes in Vue code.

Would you welcome a PR that introduces this change? :octocat:

m1guelpf:
## Description

This PR replaces all URLs on the Inertia skeleton with named routes, making use of [the Ziggy package from Tighten](https://github.com/tightenco/ziggy).

## Related Issues

Closes #157

## Future Improvements

I wasn't able to replace the URLs on the two-factor authentication component since their routes aren't named. If requested, I could submit a PR to fortify that adds names for these routes for completion's sake.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -247,7 +247,7 @@ EOF;
     protected function installInertiaStack()
     {
         // Install Inertia...
-        (new Process(['composer', 'require', 'inertiajs/inertia-laravel', 'laravel/sanctum:^2.6'], base_path()))
+        (new Process(['composer', 'require', 'inertiajs/inertia-laravel', 'laravel/sanctum:^2.6', 'tightenco/ziggy'], base_path()))
                 ->setTimeout(null)
                 ->run(function ($type, $output) {
                     $this->output->write($output);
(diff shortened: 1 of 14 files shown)


driesvints:
Think we'll need to target 2.x here since ziggy is now basically required and existing installs won't have the dependency.

taylorotwell:
@m1guelpf I don't think it matters since this only really affects the installation process and stubs. Existing apps won't have it but new apps will.

driesvints:
@taylorotwell correct, didn't saw that only the stubs were updated 👍

taylorotwell:
@m1guelpf after merging this and testing locally I get 419 CSRF problems on every POST.

m1guelpf:
I don't get how this should affect anything, the result should be exactly the same

taylorotwell:
@m1guelpf you don't have this problem?

m1guelpf:
I didn't, but I also didn't fully test all endpoints (just made sure no pages throwed errors and tried a few), so it's possible I didn't trigger any POST ones. The interesting thing is, if you `dd` on the `tokensMatch` function of the `VerifyCsrfToken` middleware before the check, both tokens match, but if you let the check happen, they fail.

taylorotwell:
So you are able to reproduce what I'm talking about now?

PR #315, opened 2020-10-06
[1.x] Bump Tailwind & Inertia

rennokki:


gradedcatfood:
Thank you! This looks like it will fix my issue with `this.$inertia.get is not a function`. I was confused as the inertia lists this method multiple times in their docs but this update will fix this.

Issue #326, opened 2020-10-08
Inertia Sharing and Versioning


devingray:
Hi @taylorotwell and the Laravel Team. 

I am wondering if a PR to allow Inertia versioning and sharing to be published would be something that the package would be interested in. 

Some context:

Versioning: 
Currently because Jetstream does not version the Inertia Bundle it needs a cleared cache reload before seeing changes on the front end. 

This can be easily achieved in the register method of `AppServiceProvider.php` but I personally like to create an InertiaServiceProvider to keep everything in their neat little rows. 

An example code would be something like 

```php
            Inertia::version(function () {
                return md5_file(public_path('mix-manifest.json'));
            });
```

Sharing

I noticed that sharing is provided at a middleware level but is not published, this houses some data such as has_two_factor enabled, I guess it could be overwritten if I were to create my own but this just adds steps to the request cycle.

My idea is to create an InertiaServiceProvider with some default boilerplate to get this going. 

Feel free to close if this is not the direction of the package

claudiodekker:
Saw this requested on Twitter as well, and I think it'd be a great addition:
- Laravel uses Laravel Mix to generate the `mix-manifest.json` out of the box
- In my experience working on Inertia with @reinink, it's very rare that people don't want this automatic asset versioning

I was already planning on creating a PR for this later today, but since now there's a ticket for it, feel free to assign it to me @driesvints 👍

reinink:
It's worth noting that unless you actually use `mix.version()`, this won't help. You need the two features enable for versioning to work...otherwise the md5 hash never changes.

Would adding `.version()` to the `webpack.mix.js` be too opinionated?

driesvints:
I always add it by default myself but unsure how @taylorotwell feels about it.

reinink:
Yeah same.

And, just to be clear, I'm only suggesting that you'd add `.version()` to the Jetstream `webpack.mix.js` stub, not the Laravel default. 👍

driesvints:
I'd do both but that's maybe a bit too opinionated 😄

m1guelpf:
Keep in mind that this still doesn't cover cases like Vapor where there's no mix manifest file. If this is added, I'd rather have it in userland where it can be easily removed.

claudiodekker:
@m1guelpf Good point.

As for adding it to the `webpack.mix.js`, I considered that, but didn't do it because it's shared with Livewire. 
We could add the `.version()` to it, but then it would be automatically enabled over there as well. 
[...]

PR #327, opened 2020-10-08
[1.x] Inertia: Automatically publish & register Inertia Middleware

claudiodekker:
Solves #326 

This PR automatically publishes and registers the new `HandleInertiaRequests` Inertia Middleware with the Http Kernel, to enable automatic asset versioning (cache busting) and error sharing as of default. To take full advantage of this, it also enables versioning as of default in the `webpack.mix.js`.

Additionally, because Jetstream [automatically registers the `ShareInertiaData` middleware on the 'web' group as priority middleware](https://github.com/laravel/jetstream/pull/327/files#diff-232229d3c322d98dbcce6ac50983b7ec119b365097def676313c286fa15b6d00R206-R207), this PR also adds logic to ensure the `HandleInertiaRequests` middleware always runs directly after it when it exists. This gives the user the ability to overwrite the Inertia built-in defaults (such as the `user` key) if they want to.

Finally, because this Middleware will exist entirely in userland, it allows the user to change it in to any configuration they want, including disabling it entirely.


src/Http/Middleware/ShareInertiaData.php
```diff
@@ -20,6 +20,10 @@ class ShareInertiaData
      */
     public function handle($request, $next)
     {
+        Inertia::version(function () {
+            return md5_file(public_path('mix-manifest.json'));
```
m1guelpf, line 24: What if this file didn't exist? Wouldn't that make Laravel automatically 500?

m1guelpf (changes requested):

src/Http/Middleware/ShareInertiaData.php
reinink: Yup.

claudiodekker: Good point. I'll do some more testing on this, including seeing if I can find a good solution for Vapor.

m1guelpf: For Vapor, you can use `config('app.asset_url')`, but my point is that this shouldn't be on vendor code.

claudiodekker: Yeah, I already had a fix for it ready, I just didn't get around to testing it yet as I got interrupted by some other things. I just pushed the changes. Do you still think it's a bad idea now?

I could move it to a publishable Service Provider, but then it'd always boot. 
The nice thing about the current solution (middleware) is that it's only booted when it's actually being used.

claudiodekker:
Alright, so, tons of changes. Let me guide you through it.

First, I've solved the issue of the `Inertia::version` being part of the vendor, by moving it to a newly-published `InertiaServiceProvider`. 

Because as a result of this we ended up with a lot of duplicated logic (`installFortifyServiceProvider`, `installJetstreamServiceProvider`, `installInertiaServiceProvider`), I've refactored this methods to a single `installServiceProvider($name, $after)` method.

Next, I ran into the dilemma that while Jetstream's `Inertia::share` would be nice in userland, doing so would cause a lot of repetition. We already had two different files (`JetstreamServiceProvider.php`, `JetstreamWithTeamsServiceProvider.php`), and making this change would result in 4 files, as we'd also need one for Inertia as well as one for Inertia with Teams.

To solve this issue, I remembered a thing I learned from @taylorotwell (Laravel ☁️ ) where he used Blade views to 'generate' scripts. I basically did the same thing by moving the `JetstreamServiceProvider` into a single 'template', and rendering the file into the state in which it's needed on-the-fly.

Hope this makes sense!

taylorotwell:
To me it feels odd to call Inertia::share outside of a web request context. For example, I would pretty much expect it to be done in a middleware or controller where there is definitely an HTTP request happening. A console command will fire up all service providers and this will call Inertia::share in that case.

claudiodekker:
Yeah, I definitely understand, but the inability to define your own defaults and override the user there is kind of problematic, as it'd force you to pretty much drop Jetstream in that situation.

Personally, I don't think its too big of a problem to have them bootstrap like this. Yes, it's maybe not the most efficient, but I kind of see Inertia::share as a way to register container bindings that aren't resolved yet

Perhaps there's something else we can try? Maybe even a change that we can make upstream in Inertia's Laravel adapter?

claudiodekker:
Done. I've reverted the changes and will update the original message of this PR to reflect this as well.
(Sorry for the frequent force-pushes, having a bit of a fight here here with my machine.. )

In addition, I've briefly had a chat about this topic with @reinink, since we're using `Inertia::share` inside the `AppServiceProvider` on our [Inertia best-practices / demo repository (PingCRM)](https://github.com/inertiajs/pingcrm). I'll create a PR to that repository over the next few days to use Middleware instead, similar to what you're doing here. 👍

claudiodekker:
I've also gone ahead and enabled `.version()` in the `webpack.mix.js` to take advantage of versioning as of default, as was suggested in #326. I do realize it's now also enabled for Livewire, but I think that's fine?

taylorotwell:
Is this ready? If so, can you mark it as ready. I have my GitHub set to filter out draft PRs.

claudiodekker:
Yeah, it's ready, but @reinink asked me to re-draft it so he could talk to you about it.. Still want me to mark it as ready?

src/JetstreamServiceProvider.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Laravel\Jetstream;
 
+use App\Http\Middleware\PrepareInertia;
```
bonzai, line 5: This will not work if my namespace is not `App`.

taylorotwell:
Status? Should this be made active?

claudiodekker:
Sorry, forgot to post an update here.

Because of the conversation that happened here, we've decided to change Inertia's Laravel adapter to use Middleware as-of default for sharing data etc. as well, resulting in this PR: https://github.com/inertiajs/inertia-laravel/pull/161

The plan right is to release that PR tomorrow, at which point I'll update this one and will mark it ready for review.

claudiodekker:
Alright, so, I think this is ready, but I'll test it to make sure tomorrow.

claudiodekker:
@taylorotwell Alright, everything seems to be working 👍  

I also went ahead and split the `webpack.mix.js` file to be stack-specific, because we already had [an ongoing discussion about this over here](https://github.com/laravel/jetstream/pull/328#issuecomment-706404285), as well as that I didn't really like the idea of touching something that Livewire also gets affected by if I don't have to.

taylorotwell:
@claudiodekker  Can this be re-targeted to master so we can release all of these Inertia improvements at once?

claudiodekker:
@taylorotwell Sure, give me 5 mins.. verifying that everything will keep working

claudiodekker:
@taylorotwell Done 👍

taylorotwell:
I have a feeling all that middleware replacement using PHP_EOL isn't going to work on Windows. 👀

src/Console/InstallCommand.php
```diff
@@ -424,21 +411,52 @@ protected function ensureApplicationIsTeamCompatible()
...
+     *
+     * @param  string $after
+     * @param  string $name
+     * @param  string $group
+     * @return void
+     */
+    protected function installMiddlewareAfter($after, $name, $group = 'web')
+    {
+        $httpKernel = file_get_contents(app_path('Http/Kernel.php'));
+        $middlewareGroups = Str::before(Str::after($httpKernel, '$middlewareGroups = ['.PHP_EOL), '];'.PHP_EOL);
```
taylorotwell, line 442: Particularly here. On Windows I'm not sure this Str::after call will return what you want.

claudiodekker: Yeah, I think you're right.. I totally forgot about Windows doing things like this differently.

Let me see if I can adjust this.

taylorotwell:
Honestly my gut would be to just replace the HTTP kernel entirely with a stub that contains exactly what you want. It will have some duplication with Laravel itself but would prevent the need for this complicated replace logic.

claudiodekker:
Yeah, agreed, but it does add extra maintenance if there's a change made in Laravel itself.

claudiodekker:
@taylorotwell How's this? Or do you still prefer creating a HTTP kernel stub?

taylorotwell:
image

I get this. I assume because you're installing the Inertia dependency but then calling a command in the same process such that for that process it is not installed yet. Make sense?

claudiodekker:
Ah shoot! Yeah, I get what you're saying, it's re-using the container that's already booted.

taylorotwell:
Probably you would need to use Symfony Process to run the command.

taylorotwell:
OK seems fine to me now. Good to merge?

claudiodekker:
Yep, sure. I just got a pretty crazy idea that might actually be a better alternative to the 'patching' that I'm doing now, but I can just PR that later as it's not breaking 👍

claudiodekker:
@taylorotwell In case you were curious: I played around with that idea I mentioned for a bit, but I don't think the end result would be better.

Basically, what I wanted to do, is use [ReflectionClass::getDefaultProperties](https://www.php.net/manual/en/reflectionclass.getdefaultproperties.php) to load the Kernel's class properties, and add the middleware to that. Then, I'd basically implode the array, and inject it into a `Kernel.stub` file, followed by overwriting the result to the application's `Kernel.php`

While this would work just fine, all comments or middleware that's included but 'commented' by default would be stripped out, and without a good way to grab those I think the end result would be less useful compared to what we have now.

PR #328, opened 2020-10-08
[1.x] Added js directory path alias

slavarazum:
With `@` alias for js folder, we can use absoulte path instead of relative.

Imports will be more clear:

```diff
-import JetLabel from './../../Jetstream/Label'
+import JetLabel from '@/Jetstream/Label'
```

Its standard practice from vue cli preset. Also, you can check out article [https://laravel-news.com/laravel-mix-alias](https://laravel-news.com/laravel-mix-alias) with example of implementation.

### DIFF
--- a/stubs/webpack.mix.js
+++ b/stubs/webpack.mix.js
@@ -1,4 +1,5 @@
 const mix = require('laravel-mix');
+const path = require('path');
 
 /*
  |--------------------------------------------------------------------------
@@ -15,4 +16,11 @@ mix.js('resources/js/app.js', 'public/js')
     .postCss('resources/css/app.css', 'public/css', [
         require('postcss-import'),
         require('tailwindcss'),
-    ]);
+    ])
+    .webpackConfig({
+        resolve: {
+            alias: {
+                '@': path.resolve(__dirname, 'resources/js/')
+            }
+        }
+    });
--- a/stubs/webpack.mix.js
+++ b/stubs/webpack.mix.js
@@ -20,7 +20,7 @@ mix.js('resources/js/app.js', 'public/js')
     .webpackConfig({
         resolve: {
             alias: {
-                '@': path.resolve(__dirname, 'resources/js/')
+                '@': path.resolve('resources/js'),
             }
         }
     });
--- a/stubs/webpack.mix.js
+++ b/stubs/webpack.mix.js
@@ -1,5 +1,4 @@
 const mix = require('laravel-mix');
-const path = require('path');
 
 /*
  |--------------------------------------------------------------------------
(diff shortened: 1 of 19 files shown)


stubs/webpack.mix.js
```diff
@@ -15,4 +16,11 @@ mix.js('resources/js/app.js', 'public/js')
     .postCss('resources/css/app.css', 'public/css', [
         require('postcss-import'),
         require('tailwindcss'),
-    ]);
+    ])
+    .webpackConfig({
+        resolve: {
+            alias: {
+                '@': path.resolve(__dirname, 'resources/js/')
```
reinink, line 23: I wonder if this should be shortened to:

```js
'@': path.resolve('resources/js'),
```

(That's what I've [always done](https://github.com/inertiajs/pingcrm/blob/master/webpack.mix.js#L40).)

claudiodekker:
I personally do really like this, but I do have two concerns:
- Aliasing to `@` does cause issues with discovery in PHPStorm (and perhaps other editors?) [There are workarounds for this](https://stefanbauer.me/tips-and-tricks/autocompletion-for-webpack-path-aliases-in-phpstorm-when-using-laravel-mix), but this makes me doubt whether this is a bit _too opinionated_ to be a default?
- This file is shared with Livewire. Are we sure it doesn't cause issues there either?

Regarding the file being shared, [I already mentioned this earlier today on this issue](https://github.com/laravel/jetstream/issues/326#issuecomment-705566445), so splitting it might be the way forward anyway.

claudiodekker:
Additionally, it might make more sense to target v2 (`master`), as this is not really a bugfix, and even more importantly because there's newly added Inertia views there (see #298) that we'd otherwise need to create a separate PR for anyway.

stubs/webpack.mix.js
slavarazum: Just tested, it works 👍
Pushed the update.

slavarazum:
@claudiodekker This update haven't any breaking changes and doesn't force you to use alias.

P.S. I have successfully setup alias resolving in VSCode and PHPStrom

taylorotwell:
Maybe we should break out the webpack stubs to be different for Inertia vs. Livewire? Also, any more thoughts on the PHP Storm issue? I don't use PHP Storm so can't really comment on that.

slavarazum:
Root directory alias might be useful for both presets. PHPStorm supports webpack.config detection with alias resolving.

claudiodekker:
@slavarazum Yes, that applies to `webpack.config.js`, but not to the `webpack.mix.js` that Laravel (Mix) uses & is being changed here.

One way around it [that I linked earlier](https://stefanbauer.me/tips-and-tricks/autocompletion-for-webpack-path-aliases-in-phpstorm-when-using-laravel-mix) is to create a `webpack.config.js` (that PHPStorm can use) and then to import that config into the `webpack.mix.js`, but I don't see that being implemented here, hence this being an issue.

taylorotwell:
@claudiodekker @slavarazum where do we go from here? We probably need to solve this PHP Storm issue that @claudiodekker is raising.

slavarazum:
@taylorotwell Let me export webpack config to separate file as @claudiodekker described.

PR #330, opened 2020-10-09
DeleteTeam: Check if personal team

RJFares:
App\Actions\Jetstream:

Enforce to check if the team is personal from the action itself too before deleting.


driesvints:
Seems like something went wrong with branching here.

Issue #277, opened 2020-09-27
Users logged in with "Remember me" disappear from sessions list


olumby:
- Jetstream Version: 1.3.0
- Jetstream Stack: Livewire
- Laravel Version: 8.6.0
- PHP Version: 7.4.8
- Database Driver & Version: PostgreSQL 9.6.19

### Description:

Users logged in with "Remember Me" checked are removed from "Browser Sessions" once the session lifetime expires, however remain logged in due to remember me.

I understand that the session has expired, and is recreated once the page is requested again on that device. However this could mean users are given a false sense of security, believing they are not logged in on a device they may well be logged in on, but haven't used recently.

I also understand this is more of a limitation with the Laravel database session driver, but thought it made sense to bring up here.

### Steps To Reproduce:

1. Install Jetstream
2. Configure & migrate database
3. User in _Browser 1_ logs in ticking Remember Me
4. Same user in _Browser 2_ logs in.
5. Session lifetime expires, ~100 requests later garbage collection clears expired sessions
6. User in _Browser 2_ visits  `/user/profile`
7. Only sees current session in "Browser Sessions" list
8. User in _Browser 1_ refreshes is logged in and reappears on "Browser Sessions" list

For testing, the situation can be accelerated by setting

`SESSION_LIFETIME=1` in `.env`
`'lottery' => [100, 100],` in `config/session.php`

driesvints:
I can't reproduce this with these steps, sorry. Can you first please try one of the support channels below? If you can actually identify this as a bug, feel free to report back and I'll gladly help you out and re-open this issue.

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://webchat.freenode.net/?nick=laravelnewbie&channels=%23laravel&prompt=1)

Thanks!

olumby:
I can identify the bug, just not sure how to explain it clearly so apologies for the wall of text.

The database session driver in Laravel clears the session table of expired sessions on a lottery, by default the chance is 2% on every request. It makes no distinction if this session is via "Remember me" all expired sessions are deleted. If the user didn't tick "Remember me", the user is effectively logged out, if the user did tick Remember me, their session is recalled using the Cookie in their browser. 

[...]

PR #332, opened 2020-10-09
[1.x] Note remember me for browser sessions

driesvints:
Closes https://github.com/laravel/jetstream/issues/277


paras-malhotra:
@driesvints actually it will logout the users who've logged in using "remember me". It would just not show the session in the list above.

taylorotwell:
@paras-malhotra is correct. Will adjust the wording.

PR #343, opened 2020-10-14
Added accessor for registers routes property

joelbutcher:
Following the common conventions used throughout this package, this PR adds an accessor for the `$registersRoutes` property added in #67.


taylorotwell:
Not really necessary IMO.

PR #344, opened 2020-10-14
Upgrade Jetstream to the latest InertiaJS

claudiodekker:
This PR updates Jetstream to use the latest version of Inertia (`0.4`) & Inertia Vue (`0.3`).

**Please note:** Changes in this PR rely upon the changes introduced in https://github.com/laravel/jetstream-js/pull/8, meaning this PR is breaking right now.


stubs/inertia/resources/js/Pages/API/ApiTokenManager.vue
```diff
@@ -227,9 +227,10 @@
             createApiToken() {
                 this.createApiTokenForm.post(route('api-tokens.store'), {
                     preserveScroll: true,
-                }).then(response => {
-                    if (! this.createApiTokenForm.hasErrors()) {
-                        this.displayingToken = true
+                    onSuccess: () => {
+                        if (! this.createApiTokenForm.hasErrors()) {
+                            this.displayingToken = true
+                        }
```
taylorotwell, line 233: This is still called `onSuccess` even if there are validation errors? I guess because it's a redirect on the server side?

claudiodekker: Yeah, it does.. `onSuccess` in Inertia doesn't have much to do with the HTTP request itself succeeding, but rather that a valid (usable) Inertia response came back. https://inertiajs.com/manual-visits#promise-deprecation

reinink: Yes, exactly. Success = successful Inertia visit. It's like a `then()` before.

taylorotwell:
@claudiodekker I have released the jetstream-js PR as 1.0.0. We can update the master branch of this repository to install ^1.0.0 of `laravel-jetstream` in the `InstallCommand` so that the correct version is pulled in.

taylorotwell:
I have made that change.

claudiodekker:
Dude, you're ⚡ 🤣

stubs/inertia/resources/js/app.js
```diff
@@ -2,12 +2,12 @@ require('./bootstrap');
 
 import Vue from 'vue';
 
-import { InertiaApp } from '@inertiajs/inertia-vue';
+import { app as InertiaApp, plugin as InertiaPlugin } from '@inertiajs/inertia-vue';
```
taylorotwell, line 5: What is `app` used for here? I see later we declare a `const app` as a different value?

reinink: We renamed the named exports to be more inline with what's done in the JS community (removing the "Inertia" prefixes, and using title case for components). Technically `InertiaApp` still exists, but we've added `App` and `plugin` as well. For forwards compatibility, this change is good, and shouldn't interfere with `app` below, since it's just an alias.

That said, @claudiodekker, I'd recommend using the uppercase version of `App`.

```suggestion
import { App as InertiaApp, plugin as InertiaPlugin } from '@inertiajs/inertia-vue';
```

claudiodekker: Yeah, so, right here it's just used as the 'selected' thing to import, and it's directly being aliased to `InertiaApp`, meaning that as a variable it doesn't exist yet.

Basically, Jonathan decided a while back that he wanted to deprecate `Vue.use(InertiaApp)` in favor of importing the plugin directly, and at the same decided that it might be better to export more 'generics', as other libraries also do this.

Because this makes the `InertiaApp` itself semi-deprecated (not yet, but it could be at some point) and look inconsistent otherwise, I figured I might as well go ahead and alias this at the same time.

taylorotwell:
Just noticed. Doesn't this entire PR need to go to `master` branch?

taylorotwell:
@claudiodekker In light of changing the branch to `master`, we'll now need to update the Inertia authentication pages you worked on.

claudiodekker:
@taylorotwell Yeah, sure, no worries.. 👍

claudiodekker:
@taylorotwell Done.

taylorotwell:
Thanks!

PR #346, opened 2020-10-15
fix: do not show profile pic if it is disabled

ibrunotome:
The profile photo is displayed in responsive setting options even with `Features::profilePhotos()` removed. This PR will hide it like the desktop version.


taylorotwell:
Couple of things here. 

1) The change also needs to be made to the Livewire version.

2) Because this has a px-4 when the profile photo is present... it looks a bit odd when the photo missing. It would need more like a px-1 in that case so that the text is more lined up on the left side with the other text below it.

Could you re-submit this with those changes?

ibrunotome:
@taylorotwell done #347. I never used livewire/alpine, so I don't know if there is a beautiful way than that.

PR #347, opened 2020-10-15
fix: don’t show profile pic if it’s disabled

ibrunotome:
The profile photo is displayed in responsive layout even with Features::profilePhotos() removed. This PR will hide it from livewire and inertia, like the desktop version.


taylorotwell:
Honestly I think I will just hold off on this for now. This is in a stub that is exported to your application meaning you can change it however you want.

PR #353, opened 2020-10-16
[1.x] Add Inertia select component

aarondfrancis:
Hey yall 👋 

I needed a select form input and wanted it to match the rest of jetstream, so I made a new component. Thought it might be nice to have in the library!

Here's a screenshot:

image

It uses the same SVG chevron as the profile dropdown.

Let me know if anything needs to be changed 🤙

### DIFF
--- a/stubs/inertia/resources/js/Jetstream/Select.vue
+++ b/stubs/inertia/resources/js/Jetstream/Select.vue
@@ -1,18 +1,12 @@
 <template>
-    <div class='relative'>
-        <select class="form-input w-full rounded-md shadow-sm" v-bind='$attrs' :value='value' @input="$emit('input', $event.target.value)" ref="input">
-            <option v-if='placeholder' :value='null' disabled>{{ placeholder }}</option>
-            <slot></slot>
-        </select>
-        <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" class='absolute fill-current text-gray-300 right-0 -mt-2 mr-2 h-4 w-4 pointer-events-none' style='top: 50%'>
-            <path fill-rule="evenodd" d="M5.293 7.293a1 1 0 011.414 0L10 10.586l3.293-3.293a1 1 0 111.414 1.414l-4 4a1 1 0 01-1.414 0l-4-4a1 1 0 010-1.414z" clip-rule="evenodd" />
-        </svg>
-    </div>
+    <select class="form-select w-full rounded-md shadow-sm" :value='value' @input="$emit('input', $event.target.value)" ref="select">
+        <option v-if='placeholder' :value='null' disabled>{{ placeholder }}</option>
+        <slot></slot>
+    </select>
 </template>
 
 <script>
     export default {
-        inheritAttrs: false,
         props: {
             value: {
                 required: false,
@@ -25,7 +19,7 @@
 
         methods: {
             focus() {
-                this.$refs.input.focus()
+                this.$refs.select.focus()
             }
         }
     }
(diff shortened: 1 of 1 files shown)


bonzai:
Is there any reason for not using the `form-select` class? It would make code much simpler:

```html
// no <div> wrapper
<select class="form-select rounded-md shadow-sm">
    <slot></slot>
</select>
```

Zrzut ekranu 2020-10-16 o 23 24 55

Also, you should make the Livewire version if you want your PR to be accepted.

aarondfrancis:
Only that I totally forgot the form-select class existed! Updated.

Thanks @bonzai

stubs/inertia/resources/js/Jetstream/Select.vue
```diff
@@ -0,0 +1,26 @@
+<template>
+    <select class="form-select w-full rounded-md shadow-sm" :value='value' @input="$emit('input', $event.target.value)" ref="select">
```
bonzai, line 2:
```suggestion
    <select class="form-select rounded-md shadow-sm" :value="value" @input="$emit('input', $event.target.value)" ref="select">
```

stubs/inertia/resources/js/Jetstream/Select.vue
```diff
@@ -0,0 +1,26 @@
...
+        <option v-if='placeholder' :value='null' disabled>{{ placeholder }}</option>
```
bonzai, line 3:
```suggestion
        <option v-if="placeholder" :value="null" disabled>{{ placeholder }}</option>
```

bonzai: I think `placeholder` prop is not needed - we can use `<slot></slot>` for that.

aarondfrancis: IMO it's nice to have built in, and if you want to use the `<slot>` for that you still can, it doesn't prevent you in any way.

aarondfrancis: Gah sorry, I thought StyleCI would catch all the quotes, etc.

mous2890:
I love this idea of prebuilt components to use in Vue. Thank you.

claudiodekker:
Hi, thanks for this, it's definitely something that will come in useful in some of my Inertia apps! 

That said, I'm a bit mixed on this being added to Jetstream, because to me it doesn't feel like it's the right place. My understanding is that Jetstream isn't a UI-component library but rather a UI-stack bootstrapping library.

Perhaps there's a different open source library that this would be more suitable in? Or am I missing something and this would also replace the existing selects in the Inertia Jetstream stack?

aarondfrancis:
Yeah it's a good question! I don't know if I'm the person to answer it tbh. 

If Jetstream is a stack to bootstrap apps, then by necessity it will include some UI components, as it already does. The hard question is "how many UI components?" Probably only Taylor can answer that. 

From my point of view it seems reasonable to have a select form element to go along with inputs and buttons. 

I'd probably argue _against_ pulling Select2 in or something, as that is too much in my opinion. But having a native select with the right classes and a bit of dev UX seems right to me. 

All that to say... beats me! Just thought it was worth a shot.

aarondfrancis:
Also @claudiodekker I must say that your response was extremely thoughtful and kind. You were basically saying "not sure this belongs here" but it was so nice and considerate. 

Thanks for that.

taylorotwell:
I kind of agree that with @claudiodekker and I'm not sure I want to commit to maintaining components like this. However, I do really like this idea for an add-on package someone could build for Jetstream that has this component and other useful components like it.

aarondfrancis:
Totally understood! Will assess my personal appetite for another package 🤔  😂

hugorodriguesbr (commented):
Hi, tag select has no attribute value.

PR #362, opened 2020-10-19
[1.x] Add Jetstream route prefix config

MaFx:
This PR adds config for prefixing jetstream routes in the same pattern that is used in Laravel Fortify route binding.

This allows adding prefix like /admin to routes if the application requires changes in url structure.


taylorotwell:
Why?

bilogic:
@MaFx I thought it would be good if this option also shows up in the `config/jetstream.php`.

PR #368, opened 2020-10-20
Update AppLayout.vue

Eliott-Champo:
Hi there, I suggest we add a computed property to watch if a user put something in the header slot in which case it shows if not, it does not ^^

### DIFF
--- a/stubs/inertia/resources/js/Layouts/AppLayout.vue
+++ b/stubs/inertia/resources/js/Layouts/AppLayout.vue
@@ -193,7 +193,7 @@
         </nav>
 
         <!-- Page Heading -->
-        <header class="bg-white shadow">
+        <header v-if="hasHeader" class="bg-white shadow">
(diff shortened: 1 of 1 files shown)


taylorotwell:
I am closing this pull request because it lacks sufficient explanation, tests, or both. It is difficult for us to merge pull requests without these things because the change may introduce breaking changes to the framework.

Feel free to re-submit your change with a **thorough** explanation of the feature and tests - integration tests are preferred over unit tests. Please include it's benefit to end users; the reasons it does not break any existing features; how it makes building web applications easier, etc.

Thanks!

Eliott-Champo:
Hi, sorry about not doing this the right way but it is my first PR... 

I am suggesting this because when you don't have anything in the header slot, it still display on the page as a transparent div. 

By doing this tiny change you give the opportunity for the user to haver the header slot as "optional" in the AppLayout.vue.

Attached and image showing what I get if I don't check if the slot has content.

Screenshot 2020-10-20 at 23 00 36

PR #371, opened 2020-10-21
[1.x] Lock inertiajs/inertia-laravel to ^0.2.4

claudiodekker:
This prevents issues with `PUT`/`POST`/`PATCH` requests due to the Middleware changes introduced in `inertiajs/inertia-laravel` `^0.3`. This is not an issue on `master`, because on that version these changes are already implemented (#327)

See https://twitter.com/don_jon243/status/1318926992645959683 for more info


reinink:
Good catch @claudiodekker. 👍

PR #377, opened 2020-10-22
[1.x] Update webpack.mix.js to use const instead of require

devingray:
This commit adds a `const` for `webpack.config.js` for more readability in the code. 

I know this may seem futile and pointless, but I think not using require in the middle of a file is better


taylorotwell:
Hmm, the initial version seems simpler... no extra temp variable?

PR #383, opened 2020-10-23
[1.x] Adds custom redirects after creating or deleting a team

thisiskj:
### Problem

In the application I'm developing, I would like the user to be redirected to a specific page after creating or deleting a team. Currently when a new team is created or deleted, the user is redirected to path defined by the `fortify.home` configuration variable. 

### Solution

In Laravel (without Jetstream) this is accomplished by implementing the `redirectTo()` method or a `$redirectTo` property on the LoginController, ResetPasswordController, etc...

My proposal is to allow customization of the redirects by allowing a `redirectTo()` method or a `$redirectTo` property on the following actions.

- CreatesTeams
- DeletesTeams

The consuming controllers would look for the existence of `redirectTo()` on the action and use it, otherwise default to `fortify.home`.

### DIFF
--- a/src/Http/Controllers/Inertia/TeamController.php
+++ b/src/Http/Controllers/Inertia/TeamController.php
@@ -11,9 +11,12 @@ use Laravel\Jetstream\Contracts\CreatesTeams;
 use Laravel\Jetstream\Contracts\DeletesTeams;
 use Laravel\Jetstream\Contracts\UpdatesTeamNames;
 use Laravel\Jetstream\Jetstream;
+use Laravel\Jetstream\RedirectsActions;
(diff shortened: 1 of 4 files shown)


jaulz:
@thisiskj thanks for the PR, this sounds like what I need but I am not yet sure how I can properly use it. Could you please give me a quick hint?

thisiskj:
@jaulz Sure, in the CreateTeams or DeletesTeams actions, you can just define a property named $redirectTo and give it the route. Something like this: 

```php
class CreateTeams
{

    public $redirectTo = '/my/route';

    ...
}
```

PR #395, opened 2020-10-26
[1.x] Constrain Ziggy version to ^0.9.4

bakerkretzmar:
Related to #392, this PR ensures that Jetstream v1 will install Ziggy `^0.9.4` (current as of today), _not_ Ziggy v1.

@mattstauffer pointed out that if a user adds Jetstream to their project now but doesn't run the `install` command for a month, they'll get Ziggy v1 which isn't compatible with Jetstream v1. Adding this version constraint will prevent that.


driesvints:
Whoops

PR #399, opened 2020-10-26
[1.x] Delete the users profile picture

### LINKED ISSUE #396: Deleting a profile does not delete avatar
- Jetstream Version: "1.5.1"
- Jetstream Stack: Inertia 
- Laravel Version: 8.11.2
- PHP Version: 7.4.10
- Database Driver & Version: MySQL & don't know rn - not necessary I guess?

### Description:
When deleting your account in Jetstream, the associated image it not deleted. 

### Steps To Reproduce:
- install a fresh laravel 8 project using --jet and teams
- create an account
- delete the account in the profile section
- check disk, image is still given

### Possible Fix:
I'd assume calling unlink/delete on the file using the path from the users table `profile_photo_path` field should fix it. https://github.com/laravel/jetstream/blob/6d7030031c20249bd9a0795819918f986cebe81f/stubs/app/Actions/Jetstream/DeleteUser.php#L15

lostdesign:
When a user deletes an account, the image will now be deleted as well. fixes #396

### DIFF
--- a/stubs/app/Actions/Jetstream/DeleteUser.php
+++ b/stubs/app/Actions/Jetstream/DeleteUser.php
@@ -14,6 +14,7 @@ class DeleteUser implements DeletesUsers
      */
     public function delete($user)
     {
+        $user->deleteProfilePhoto();
         $user->delete();
     }
 }
--- a/stubs/app/Actions/Jetstream/DeleteUser.php
+++ b/stubs/app/Actions/Jetstream/DeleteUser.php
@@ -3,6 +3,7 @@
 namespace App\Actions\Jetstream;
 
 use Laravel\Jetstream\Contracts\DeletesUsers;
+use Laravel\Jetstream\Features;
 
 class DeleteUser implements DeletesUsers
 {
@@ -14,7 +15,9 @@ class DeleteUser implements DeletesUsers
      */
     public function delete($user)
     {
-        $user->deleteProfilePhoto();
+        if (Features::managesProfilePhotos()) {
+            $user->deleteProfilePhoto();
+        }
         $user->delete();
     }
 }
--- a/stubs/app/Actions/Jetstream/DeleteUser.php
+++ b/stubs/app/Actions/Jetstream/DeleteUser.php
@@ -3,7 +3,6 @@
 namespace App\Actions\Jetstream;
 
 use Laravel\Jetstream\Contracts\DeletesUsers;
-use Laravel\Jetstream\Features;
 
 class DeleteUser implements DeletesUsers
 {
@@ -15,9 +14,7 @@ class DeleteUser implements DeletesUsers
      */
     public function delete($user)
     {
-        if (Features::managesProfilePhotos()) {
-            $user->deleteProfilePhoto();
-        }
+        $user->deleteProfilePhoto();
         $user->delete();
     }
 }
--- a/stubs/app/Actions/Jetstream/DeleteUserWithTeams.php
+++ b/stubs/app/Actions/Jetstream/DeleteUserWithTeams.php
@@ -36,7 +36,7 @@ class DeleteUser implements DeletesUsers
     {
         DB::transaction(function () use ($user) {
             $this->deleteTeams($user);
-
+            $user->deleteProfilePhoto();
             $user->delete();
         });
     }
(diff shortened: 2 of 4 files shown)


lostdesign:
I just realized that the profile picture feature has to be enabled, otherwise the method to delete the profile image is not available on the User model. Guess this should then be rather added to the docs? 

So when a user is deleted and the profile feature is enabled, it should also delete the image or more like they need to call `$user->deleteProfilePhoto()`

stubs/app/Actions/Jetstream/DeleteUserWithTeams.php
```diff
@@ -36,7 +36,7 @@ public function delete($user)
     {
         DB::transaction(function () use ($user) {
             $this->deleteTeams($user);
-
+            $user->deleteProfilePhoto();
```
Chrissi2812, line 39: Could check before the deletion

```suggestion
            if (\Laravel\Jetstream\Features::managesProfilePhotos()) {
                $user->deleteProfilePhoto();
            }
```

stubs/app/Actions/Jetstream/DeleteUser.php
```diff
@@ -14,6 +14,7 @@ class DeleteUser implements DeletesUsers
      */
     public function delete($user)
     {
+        $user->deleteProfilePhoto();
```
Chrissi2812, line 17:
```suggestion
        if (\Laravel\Jetstream\Features::managesProfilePhotos()) {
                $user->deleteProfilePhoto();
       }
```

Chrissi2812 (changes requested):
You could check if the Feature is enabled before, so no errors if it's not enabled

stubs/app/Actions/Jetstream/DeleteUser.php
lostdesign: Ohhh right, yeah just came across that feature check yesterday but didn't quite connect the dots here. Added it!

stubs/app/Actions/Jetstream/DeleteUser.php
```diff
@@ -14,6 +15,9 @@ class DeleteUser implements DeletesUsers
      */
     public function delete($user)
     {
+        if (Features::managesProfilePhotos()) {
+            $user->deleteProfilePhoto();
+        }
         $user->delete();
```
claudiodekker, line 21:
```suggestion
        }

        $user->delete();
```

stubs/app/Actions/Jetstream/DeleteUserWithTeams.php
```diff
@@ -36,7 +37,9 @@ public function delete($user)
     {
         DB::transaction(function () use ($user) {
             $this->deleteTeams($user);
-
+            if (Features::managesProfilePhotos()) {
+                $user->deleteProfilePhoto();
+            }
             $user->delete();
```
claudiodekker, line 43:
```suggestion
            }

            $user->delete();
```

stubs/app/Actions/Jetstream/DeleteUserWithTeams.php
```diff
@@ -36,6 +37,9 @@ public function delete($user)
     {
         DB::transaction(function () use ($user) {
             $this->deleteTeams($user);
+            if (Features::managesProfilePhotos()) {
```
claudiodekker, line 40:
```suggestion
            $this->deleteTeams($user);

            if (Features::managesProfilePhotos()) {
```

driesvints (approved):

claudiodekker (approved):

taylorotwell:
Personally I would move that feature check into the `deleteProfilePhoto` method so you can just always call it from the stub.

taylorotwell:
Looks like tests are failing...

lostdesign:
@taylorotwell adjusted, guess that's the way you meant? The test now fails because the tests\Fixture\User.php  doesnt have the `HasProfilePicture` trait. Is that intended? Otherwise I'd add it so the test passes 👍 
https://github.com/laravel/jetstream/blob/e37dbff40a74d40580130f15d2cf401489580d2d/tests/Fixtures/User.php#L11

taylorotwell:
You can add it to the fixture. Yes that looks fine. Thanks.

PR #404, opened 2020-10-27
[1.x] Fix Livewire Modal Closing Issue By Bumping Alpine To 2.7.3

### LINKED ISSUE #384: Jetstream modals are closing abruptly on dismiss
- Jetstream Version: 1.5.1
- Jetstream Stack: Livewire
- Laravel Version: 8.11.2
- PHP Version: 7.4.6

### Description:

When dismissing a modal by pressing the Cancel button (or the Save action), the modal is abruptly dismissed without respecting the animation. It should behave the same as when the modal is dismissed for a background click.

### Steps To Reproduce:

I slowed down the connection to show the issue.

https://www.loom.com/share/ac244139980848b087ce83a26406ef22

calebporzio:
Closes #384 

The issue where modals transition in, but not out when Livewire is the one to trigger the state change has been fixed in the latest versions of Livewire and Alpine:

Livewire: v2.3.1
Alpine: v2.7.3

The original issue should be partially fixed in older versions of Alpine (for existing installs), but a few more edge cases have been addressed that will only be fixed for people running Alpine 2.7.3 and above.

Namely, without 2.7.3 there are few transition interruption edge cases.

For example: if a user closes a modal in Jetstream, and while it's transitioning out, they somehow re-open it, it will disappear.

I'm not even sure if this case is possible to re-create in Jetstream, but just noting it here.

### DIFF
--- a/stubs/livewire/resources/views/layouts/app.blade.php
+++ b/stubs/livewire/resources/views/layouts/app.blade.php
@@ -16,7 +16,7 @@
         @livewireStyles
 
         <!-- Scripts -->
-        <script src="https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.7.0/dist/alpine.js" defer></script>
+        <script src="https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.7.3/dist/alpine.js" defer></script>
     </head>
     <body class="font-sans antialiased">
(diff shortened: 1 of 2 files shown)


taylorotwell:
You're so big brained. 🧠

PR #406, opened 2020-10-27
Delete user tokens when user is deleted

lostdesign:
Currently if a user is deleted, their api tokens will remain inside the database. The api tokens are now deleted at the same time the user will be deleted.

### DIFF
--- a/stubs/app/Actions/Jetstream/DeleteUser.php
+++ b/stubs/app/Actions/Jetstream/DeleteUser.php
@@ -15,6 +15,7 @@ class DeleteUser implements DeletesUsers
     public function delete($user)
     {
         $user->deleteProfilePhoto();
+        $user->tokens->each->delete();
         $user->delete();
     }
 }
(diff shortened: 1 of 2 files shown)


taylorotwell:
I wonder if this would be better implemented as a cascading delete foreign key constraint?

lostdesign:
Yeah, I guess cascade would be the better option here to be fair.

taylorotwell:
That's how I would personally do it.

lostdesign:
That'd mean the change has to be done in the [laravel/sanctum migration](https://github.com/laravel/sanctum/blob/268bbe1609e581c5c8352a7bda9758324d904e7e/database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php#L18)? As jetstream uses sanctum for this and the migration is there.

taylorotwell:
Ick. It's morphable too. Your PR may be the best approach.

PR #409, opened 2020-10-28
Add textarea component

ZeProf2Code:
- Added the textarea component

### DIFF
--- a/resources/views/components/textarea.blade.php
+++ b/resources/views/components/textarea.blade.php
@@ -1,0 +1,3 @@
+@props(['disabled' => false])
+
+<textarea {{ $disabled ? 'disabled' : '' }} {!! $attributes->merge(['class' => 'form-input rounded-md shadow-sm']) !!}></textarea>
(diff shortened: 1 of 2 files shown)


resources/views/components/textarea.blade.php
```diff
@@ -0,0 +1,3 @@
+@props(['disabled' => false])
+
+<textarea {{ $disabled ? 'disabled' : '' }} {!! $attributes->merge(['class' => 'form-input rounded-md shadow-sm']) !!}></textarea>
```
bonzai, line 3:
```suggestion
<textarea {{ $disabled ? 'disabled' : '' }} {!! $attributes->merge(['class' => 'form-textarea rounded-md shadow-sm']) !!}></textarea>
```

bonzai: See #353.

taylorotwell:
You can add this in your own project.

srdjanmarjanovic:
Actually, i find it useful to have textarea along with the input

GLStephen:
@taylorotwell do you mind providing any context for why you would reject textarea from the project? Jetstream has every other input type, buttons, labels, sections, modals, etc. 

Why not just put the last remaining HTML form element into the project instead of sending users on a vision quest to find it? Users will inevitably build their first, second or third form and need a textarea and start pawing through code only to find it doesn't exist.

It's not like other relative esoterica in the form world isn't already there. It's just, basically, this one missing form element. Care to explain?

troccoli:
I don't really understand either why this PR was closed, but thank you @ZeProf2Code. I was able to use your code and add a textarea Jetstream component to my project.

PR #414, opened 2020-10-29
[1.x] Update two-factor-authentication-form.blade.php

underwood:
I had an issue where I couldn't scan the barcode when setting up 2FA. Determined the issue was using darkmode tailwind css styling. The Google authenticator app and VIP Access Authenticator don't pick up the barcode if it has a dark background right up against it. Add a p-4 white background resolved the issue on both apps. Propose adding this white border out of the box to prevent this issue if the user chooses a dark color adjacent to the barcode.

### DIFF
--- a/stubs/livewire/resources/views/profile/two-factor-authentication-form.blade.php
+++ b/stubs/livewire/resources/views/profile/two-factor-authentication-form.blade.php
@@ -30,7 +30,7 @@
                     </p>
                 </div>
 
-                <div class="mt-4">
+                <div class="mt-4 p-4 w-56 bg-white">
                     {!! $this->user->twoFactorQrCodeSvg() !!}
                 </div>
             @endif
--- a/stubs/livewire/resources/views/profile/two-factor-authentication-form.blade.php
+++ b/stubs/livewire/resources/views/profile/two-factor-authentication-form.blade.php
@@ -30,7 +30,7 @@
                     </p>
                 </div>
 
-                <div class="mt-4 p-4 w-56 bg-white">
+                <div class="mt-4 dark:p-4 dark:w-56 dark:bg-white">
                     {!! $this->user->twoFactorQrCodeSvg() !!}
                 </div>
             @endif


bonzai:
Can you also add similar styles to Inertia stack?
https://github.com/laravel/jetstream/blob/1.x/stubs/inertia/resources/js/Pages/Profile/TwoFactorAuthenticationForm.vue#L34

underwood:
> Can you also add similar styles to Inertia stack?
> https://github.com/laravel/jetstream/blob/1.x/stubs/inertia/resources/js/Pages/Profile/TwoFactorAuthenticationForm.vue#L34

Good catch, I didn't think about the inertia side of things.

https://github.com/laravel/jetstream/pull/417

taylorotwell:
Can you show a screenshot of what it looks like with your code changes? The width looks good in responsive as well?

underwood:
> Can you show a screenshot of what it looks like with your code changes? The width looks good in responsive as well?

see screenshots below. It looks good in mobile, tablet, desktop.

desktop
full
mobile
mobile

taylorotwell:
Does it make the QR code look oddly shifted left in non-dark mode?

underwood:
y, the QR code wouldn't quite line up with the text.

I'm using dark: modifiers for my dark mode so we could make these only for dark:

If someone would change their background to a dark background directly without using the dark modifier then they would run in to the issue of the QR code not working.

Here are some screenshots of how it looks in light mode.
light-desktop

light-mobile

taylorotwell:
If they were to change their background they could just modify the code directly. I'm not sure Jetstream should be made to account for every possible case. The code is your code after you install Jetstream. It is no longer Jetstream's code. You should feel free to modify it as much as you want or need for your application.

underwood:
Here's another option. Centering the QRcode would make it work either way.
https://github.com/laravel/jetstream/pull/420

taylorotwell:
Can we just add the dark: modifiers here and on the Inertia PR?

underwood:
> Can we just add the dark: modifiers here and on the Inertia PR?

done and done, this works, tested in light and dark mode, that lines the QR code up with the text in light mode.

PR #420, opened 2020-10-30
Update two-factor-authentication-form.blade.php (Alternative centering the QRcode)

underwood:
Alternative to previous change, this change centers the QRcode so the extra white padding doesn't affect it's alignment relative to the text above and below in light or dark mode.

If this looks like the preferred fix, I'll add one for inertia.


underwood:
Here are some screenshots of this alternative.
dark-desktop-centered
light-mobile-centered

underwood:
just a quick note, I tightened up the padding around the QR code from 4 to 2, it still works with the authenticator apps and looks a cleaner.

PR #424, opened 2020-10-31
[1.x] Removed ui-avatars.com for GDPR compliance

mhenkel1:
To make this package compliant with GDPR rules by default (not sending the users name to a third party website), I have changed the default behaviour for inertia and livewire.

I have removed the ui-avatars.com profile photo generator.
Instead the default "profile photo" will be generated with pure tailwind and php.

We check if the user currently has an profile photo, if that is true, the image will be displayed.
If there is no image assigned to the user, a div will be generated containing the "profile tag".

The profile tag is created from the first letter of the first two words in the user name.
It will be displayed the same way as the default profile photo.

### DIFF
--- a/src/HasProfilePhoto.php
+++ b/src/HasProfilePhoto.php
@@ -72,6 +72,7 @@ trait HasProfilePhoto
         if (! Features::managesProfilePhotos()) {
             return '';
         }
+
         return 'https://ui-avatars.com/api/?name='.urlencode($this->name).'&color=7F9CF5&background=EBF4FF';
     }
 
--- a/src/HasProfilePhoto.php
+++ b/src/HasProfilePhoto.php
@@ -97,8 +97,9 @@ trait HasProfilePhoto
         $tag = '';
 
         for ($i = 0; $i < 2; $i++) {
-            if (count($segments) >= $i + 1)
+            if (count($segments) >= $i + 1) {
                 $tag .= Str::of($segments[$i])->ucfirst()->substr(0, 1);
+            }
         }
 
         return $tag;
--- a/src/HasProfilePhoto.php
+++ b/src/HasProfilePhoto.php
@@ -4,6 +4,7 @@ namespace Laravel\Jetstream;
 
 use Illuminate\Http\UploadedFile;
 use Illuminate\Support\Facades\Storage;
+use Illuminate\Support\Str;
 use Laravel\Jetstream\Features;
 
 trait HasProfilePhoto
@@ -56,17 +57,51 @@ trait HasProfilePhoto
     {
         return $this->profile_photo_path
                     ? Storage::disk($this->profilePhotoDisk())->url($this->profile_photo_path)
-                    : $this->defaultProfilePhotoUrl();
+                    : '';
     }
 
     /**
      * Get the default profile photo URL if no profile photo has been uploaded.
+     * Only return the default profile photo url if Jetstream is configured to manage profile photos.
+     * This reduces unnecessary web requests.
      *
      * @return string
      */
     protected function defaultProfilePhotoUrl()
     {
+        if (! Features::managesProfilePhotos()) {
+            return '';
+        }
         return 'https://ui-avatars.com/api/?name='.urlencode($this->name).'&color=7F9CF5&background=EBF4FF';
+    }
+
+    /**
+     * Checks if there is a profile photo.
+     *
+     * @return bool
+     */
+    public function getHasProfilePhotoAttribute()
+    {
+        return $this->profile_photo_path ? true : false;
+    }
+
+    /**
+     * Returns the profile tag, created out of the first letter of the first two words in the users name.
+     *
+     * @return string
+     */
+    public function getProfileTagAttribute()
+    {
+        $segments = Str::of($this->name)->split('/[\s ]+/');
+
+        $tag = '';
+
+        for ($i = 0; $i < 2; $i++) {
+            if (count($segments) >= $i+1)
+                $tag .= Str::of($segments[$i])->ucfirst()->substr(0,1);
+        }
+
+        return $tag;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/HasProfilePhoto.php
```diff
@@ -56,19 +57,55 @@ public function getProfilePhotoUrlAttribute()
     {
         return $this->profile_photo_path
                     ? Storage::disk($this->profilePhotoDisk())->url($this->profile_photo_path)
-                    : $this->defaultProfilePhotoUrl();
+                    : '';
```
bonzai, line 60: It's a breaking change.

src/HasProfilePhoto.php
```diff
@@ -56,19 +57,55 @@ public function getProfilePhotoUrlAttribute()
...
      * Get the default profile photo URL if no profile photo has been uploaded.
+     * Only return the default profile photo url if Jetstream is configured to manage profile photos.
+     * This reduces unnecessary web requests.
      *
      * @return string
      */
     protected function defaultProfilePhotoUrl()
     {
+        if (! Features::managesProfilePhotos()) {
+            return '';
```
bonzai, line 73: It's also a breaking change.

src/HasProfilePhoto.php
```diff
@@ -56,19 +57,55 @@ public function getProfilePhotoUrlAttribute()
...
+     * @return string
+     */
+    public function getProfileTagAttribute()
+    {
+        $segments = Str::of($this->name)->split('/[\s ]+/');
+
+        $tag = '';
+
+        for ($i = 0; $i < 2; $i++) {
+            if (count($segments) >= $i + 1) {
```
bonzai, line 101: You can count the number of segments before `for` loop.

src/HasProfilePhoto.php
mhenkel1: Thats why I have updated both livewire and inertia.
But I guess you are right, users already having jetstream installed will not get the default template updates.

I will think about something.

src/HasProfilePhoto.php
mhenkel1: True, gonna change that

mhenkel1:
@bonzai Now it shouldn't be breaking anything.

taylorotwell:
Change this in your own app if GDPR is relevant to you.

mhenkel1:
> Change this in your own app if GDPR is relevant to you.
@taylorotwell 
I already changed it. I just thought it would be a good change for Jetstream too.
Since that web request is unnecessary and could be entirely avoided by using default Tailwind design elements, I created this pull request.
And it was working and behaving like Jetstream was before my change.

It is not just relevant to gdpr, it is relevant to a lot of rules and regulations in multiple countries.
And I want you to answer this question for yourself too: 
Do you really want to send each and every user's name to a third party service, when all that could be avoided by default?

driesvints:
@mhenkel1 like Taylor said: feel free to adjust this in your own app if you don't want that.

mhenkel1:
> @mhenkel1 like Taylor said: feel free to adjust this in your own app if you don't want that.

So you rather want to force every user that uses Jetstream, having to change his template to remove a default setting, which disregards any and every privacy concerns?
Instead of just clicking the merge button and making Jetstream by default regard every users privacy?

relaypilot:
I like the way it is at the moment and that it is a default feature of Jetstream. If all the features were disabled by default, there would be no purpose in using Jetstream, as you could just start with a blank Laravel app and manually add the features that you need. The GDPR concern is something that you should worry at a personal level, in my opinion.

mhenkel1:
> I like the way it is at the moment and that it is a default feature of Jetstream. If all the features were disabled by default, there would be no purpose in using Jetstream, as you could just start with a blank Laravel app and manually add the features that you need. The GDPR concern is something that you should worry at a personal level, in my opinion.

@relaypilot @driesvints @taylorotwell 
Have you not looked at what I actually commited?
My pull request is not removing a feature, it is simply changing it in a really basic way.

It takes this: 
`profile_photo_url }}" alt="{{ Auth::user()->name }}" />`

And checks if the user has a profile photo, if it does, it will not change anything!
If the user does not have a profile photo it will do this instead:
`<div class="h-8 w-8 rounded-full object-center text-center bg-blue-100 text-indigo-700">`
 `<span class="absolute inset-0 flex items-center justify-center">{{ Auth::user()->profile_tag }}</span>`
`</div>`

As you can see nothing is removed, the feature is still there, the profile photo is still there.
The only case something is changed is, if the user does not have a profile photo, it will insert the div instead of the profile photo in 2 places:
- top right
- profile settings page

It does that for both Livewire and Inertia.

I created this pull request to simply change the default behavior and because I was told to create it when I created an issue about this behavior.

No feature is being removed and I have the feeling, that nobody except @bonzai even looked at what I did.
Feels like I was told to do it, I invested 2 hours to do it, and then everybody just reads the title and is like no its a feature we don't want to change it.

Look at what it does, try it out and then see for yourself.
There is still the users name displayed in a short form in a rounded circle with another background color if the user does not have a profile photo.

tedsecretsource:
As someone working in Europe, I can appreciate where you are coming from.

>I invested 2 hours to do it

Although this may seem like a lot of time, I can assure you it isn't for the type of change you are proposing. In your PR I don't see any tests they can use to verify your changes aren't breaking anything unseen. Also, I much prefer the more terse approach in the original code: I call an image, I get an image. I'm not fond of adding conditions to [blade templates](https://github.com/laravel/jetstream/pull/424/files#diff-02f0d0445fdcc7cb7fecff9c9117d79eb9bb8097556eabba40757a514f924805R27)

I know this is frustrating but patience and persistence is the only path to performant patches.

src/HasProfilePhoto.php
Lewiscowles1986: why is everything returning empty string?

miclaus:
@taylorotwell @driesvints 

Shouldn't this feature **at least** be mentioned in the documentation?

When thinking about scalability: most apps probably need to change it at some point.

When thinking about privacy policy: developers need to take this into account from the start.
Basically anyone from Europe using Jetstream is put at risk breaking the GDPR without even knowingly doing so – because it is not documented.

It's a nice feature, but IMO it should be changed (without any outside request) or at least be mentioned in the docs.

PR #431, opened 2020-11-05
[2.x] Adapt hasTeamRole check to only check the role even for team owners

christophrumpel:
The current `hasTeamRole` method, always returns `true` if the user is the owner. This makes sense for the `hasTeamPermission` check, but not for the role. Even the owner has just one role that should be checked.

```php
public function hasTeamRole($team, string $role)
{
        if ($this->ownsTeam($team)) {
            return true; // here we just return true for every owner no matter the given role
        }

        // Check the role
        // ...
}
```

If you show some data depending on the role (I know there are permission too but some prefer to check the role), you will run into issues where the owner will be shown things that you only wanted to be seen by a specific role.

This PR changes the `hasTeamRole` method to check the role for the given user, even for owners.

Since developers currently already use this method as it works, it makes sense to see it as a feature for the next major release.

### DIFF
--- a/src/HasTeams.php
+++ b/src/HasTeams.php
@@ -143,10 +143,6 @@ trait HasTeams
      */
     public function hasTeamRole($team, string $role)
     {
-        if ($this->ownsTeam($team)) {
-            return true;
-        }
-
         return $this->belongsToTeam($team) && optional(Jetstream::findRole($team->users->where(
             'id', $this->id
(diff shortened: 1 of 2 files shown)


taylorotwell:
It's not my intention to ever attach the owner to the team as a regular user. They don't go in the intermediate table at all. I think I will leave this as is for now.

PR #432, opened 2020-11-05
[1.x] Add account deletion feature

redidizzy:
I'm building an app that shouldn't even allow users to delete their accounts.
I thought it would be useful to add it as a feature.


taylorotwell:
Would need to go to master and you did not implement it for Inertia.

PR #433, opened 2020-11-05
[2.x] Add account deletion feature

redidizzy:
I'm building an app that shouldn't even allow users to delete their accounts.
I thought it would be useful to add it as a feature.

### DIFF
--- a/routes/inertia.php
+++ b/routes/inertia.php
@@ -24,7 +24,7 @@ Route::group(['middleware' => config('jetstream.middleware', ['web'])], function
         Route::delete('/user/profile-photo', [ProfilePhotoController::class, 'destroy'])
                     ->name('current-user-photo.destroy');
 
-        if(Jetstream::hasAccountDeletionFeatures()){
+        if (Jetstream::hasAccountDeletionFeatures()) {
             Route::delete('/user', [CurrentUserController::class, 'destroy'])
                         ->name('current-user.destroy');
         }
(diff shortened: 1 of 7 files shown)


routes/inertia.php
```diff
@@ -16,16 +16,18 @@
     Route::group(['middleware' => ['auth', 'verified']], function () {
         // User & Profile...
         Route::get('/user/profile', [UserProfileController::class, 'show'])
-                    ->name('profile.show');
+            ->name('profile.show');
```
driesvints, line 19: Please undo all styling changes in this PR.

stubs/livewire/resources/views/profile/show.blade.php
```diff
@@ -33,11 +33,13 @@
                 @livewire('profile.logout-other-browser-sessions-form')
             </div>
 
-            <x-jet-section-border />
+            @if(Laravel\Jetstream\Jetstream::hasAccountDeletionFeatures())
```
bonzai, line 36:
```suggestion
            @if (Laravel\Jetstream\Jetstream::hasAccountDeletionFeatures())
```

src/Features.php
```diff
@@ -111,4 +120,14 @@ public static function teams(array $options = [])
...
     }
+
+    /**
+     * Enable the account deletion feature.
+     *
+     * @return string
+     */
+    public static function accountDeletion()
+    {
+        return 'accountDeletion';
```
bonzai, line 131:
```suggestion
        return 'account-deletion';
```

src/Features.php
```diff
@@ -77,6 +77,15 @@ public static function sendsTeamInvitations()
         return static::optionEnabled(static::teams(), 'invitations');
     }
 
+    /** Determine if the application is using any account deletion features.
+     *
```
bonzai, line 81:
```suggestion
    /**
     * Determine if the application is using any account deletion features.
     *
```

PR #437, opened 2020-11-08
[1.x] Option for custom name field for default profile photo URL

jamesbhatta:
Right now HasProfilePhoto trait uses "name" field to generate the default profile photo.

With this change the field key can be overwritten as :

protected $defaultProfilePhotoKey = 'first_name';


taylorotwell:
Overwrite the entire method on your user model.

PR #439, opened 2020-11-08
[2.x] Fix invalid key name

bonzai:
Profile page isn't accessible because of invalid key name:

> [Vue warn]: Error in render: "TypeError: Cannot read property 'hasAccountDeletionFeatures' of undefined"


driesvints:
Thanks!

Issue #214, opened 2020-09-16
OAuth login / Socialite


barryvdh:
Besides normal login and two-factor, it might be nice to add OAuth login. Because Socialite is a first-party package, this could make sense to add by default. 

Ideas:
 - Configure socialite providers, eg Google, Facebook etc
 - Register with Socialite (define the fields)
 - ~Whitelist~ Allowlist registration for certain domains (eg. let everyone inside the company register, nobody else)
 - Login with Socialite (perhaps optionally disable normal login)
 - Connect/disconnect your account with other accounts on your profile page
 - Possibility to skip 2FA if coming from Socialite (because other accounts could handle it)

What would be needed
 - Additional table with socialite ids
 - UI for login/connect/register
 - Registering of socialite / mapping of fields

abenettt:
I really like this idea! But we should refrain from using terms such as `whitelist` and `blacklist` and use terms such as `allowlist` and `blocklist` or something similar.

FoksVHox:
> I really like this idea! But we should refrain from using terms such as `whitelist` and `blacklist` and use terms such as `allowlist` and `blocklist` or something similar.

Why?

abenettt:
@FoksVHox https://www.zdnet.com/article/linux-team-approves-new-terminology-bans-terms-like-blacklist-and-slave

FoksVHox:
> @FoksVHox https://www.zdnet.com/article/linux-team-approves-new-terminology-bans-terms-like-blacklist-and-slave

That's not a reason why we should change the words. The article just describes how it's a symbiotic change. The 'issue' is an issue that doesn't exist.

driesvints:
I closed https://github.com/laravel/jetstream/issues/84 before but I'll bring this up with Taylor.

barryvdh:
Sorry, Allowlist is better indeed. I've updated it, so can we please stay on track ;)

@driesvints Thank you. #84 is the other way around though I think? This ticket was meant being able to login using Socialite, not provide an OAuth server.

driesvints:
@barryvdh you're right 👍

NanoCellWebDesign:
@abenettt 
> I really like this idea! But we should refrain from using terms such as `whitelist` and `blacklist` and use terms such as `allowlist` and `blocklist` or something similar.

Bit excessive, can't we keep political and/or personal views out of a git issue, the OP just wants help.....

Here is some info on why it is named whitelist and blacklist:

[...]

PR #444, opened 2020-11-10
[2.x]  Socialite Support

joelbutcher:
This PR adds optional support for Laravel Socialite and all Socialite support providers, via feature flags. Support is available for both InertiaJS and Livewire. See #214 

To get started with this PR, create a fresh Laravel project (without `--jet`) and update your `composer.json` to include the repository following and require the feature branch.

```
composer require laravel/jetstream:dev-feature-jetstream-support
```

You may then run the Jetstream install command, specifying socialite support:

```
php artisan jetstream:install {stack} --socialite
```

You may remove support for Socialite by commenting out the feature in `jetstream.php`

```php
'features' => [
    // Features::socialite(['facebook' => true, 'github' => true]),
],
```

You'll need to follow the instructions for [configuring socialite providers](https://laravel.com/docs/8.x/socialite#configuration) to ensure you're application will support this feature.

Once you're all setup, you should be able to see your providers in your `login` and `register` routes views:

Screenshot 2020-11-10 at 21 06 36

Existing users that have signed up using email and password can connect social accounts in their profile settings, as well as remove them:

Screenshot 2020-11-13 at 17 04 10


resources/views/components/socialite.blade.php
```diff
@@ -0,0 +1,49 @@
...
+        </a>
+    @endif
+
+    @if(\Laravel\Jetstream\JetStream::hasSocialiteSupportFor('gitlab'))
+        <a href="{{ route('socialite.redirect', ['provider' => 'gitlab']) }}">
+            <x-jet-gitlab-icon class="h-6 w-6 mx-2" />
+        </a>
+    @endif
+
+    @if(\Laravel\Jetstream\JetStream::hasSocialiteSupportFor('bitbucket'))
```
driesvints, line 44: You can remove the leading slash for all these FQN's.

stubs/inertia/resources/js/Pages/Profile/Show.vue
```diff
@@ -20,11 +20,17 @@
...
+
+                    <remove-connected-accounts-form :accounts="connectedAccounts" class="mt-10 sm:mt-0" />
+                </div>
+
                 <jet-section-border />
 
                 <logout-other-browser-sessions-form :sessions="sessions" class="mt-10 sm:mt-0" />
 
-                <template v-if="$page.props.jetsteam.hasAccountDeletionFeatures">
+                <template v-if="$page.props.jetstream.hasAccountDeletionFeatures">
```
driesvints, line 33: I think we already got a PR for this. Might need to merge 1.x into master for this. If not, can you send this in separately to 1.x?

stubs/livewire/resources/views/profile/update-profile-information-form.blade.php
```diff
@@ -62,7 +62,11 @@
         <!-- Email -->
         <div class="col-span-6 sm:col-span-4">
             <x-jet-label for="email" value="{{ __('Email') }}" />
-            <x-jet-input id="email" type="email" class="mt-1 block w-full" wire:model.defer="state.email" />
+            @if ($emailDisabled)
+                <x-jet-input id="email" type="email" class="mt-1 block w-full" wire:model.defer="state.email" disabled />
+            @else
+                <x-jet-input id="email" type="email" class="mt-1 block w-full" wire:model.defer="state.email" />
+            @endif
```
driesvints, line 69: Leave a white line above and below this if block.

stubs/livewire/resources/views/profile/show.blade.php
```diff
@@ -29,6 +29,14 @@
                 <x-jet-section-border />
             @endif
 
+            @if (Laravel\Jetstream\Jetstream::hasSocialiteFeatures())
+                <div class="mt-10 sm:mt-0">
+                    @livewire('profile.remove-connected-accounts-form')
+                </div>
+
+                    <x-jet-section-border />
```
driesvints, line 37: Indentation is off here.

stubs/livewire/resources/views/profile/remove-connected-accounts-form.blade.php
```diff
@@ -0,0 +1,101 @@
...
+                                </div>
+
+                                <div class="text-xs text-gray-500">
+                                    Added on {{ $account->created_at }}
+                                </div>
+                            </div>
+                        </div>
+
+                        <!-- Remove Action -->
+                        @if(count($this->accounts) > 1)
```
driesvints, line 70: For consistency can you go through your PR and make sure that all `@if (` directives have a space between the `if` and `(`?

driesvints:
@joelbutcher why did you close this?

joelbutcher:
@driesvints Apologies, I rebased the branch and push it up, but from a different point in time - the git diff forced it to close for some reason. I've made a note of the requested changes and will re-open this PR

joelbutcher:
@driesvints I've addressed the changes requests in the latest. Once again apologies for the mess-up with Git!

driesvints:
@joelbutcher no worries. This is a fantastic PR tbh 👍

joelbutcher:
@driesvints Thanks!! I was debating whether or not to use you Blade Icons ;) for the SVG's, but since there's only a handful, and I wanted to maintain the brand colors, I thought it best to just include the ones that were needed!

taylorotwell:
Have you considered instead of having a separate configuration item doing this:

```php
Features::socialite(['facebook', 'github'])
```

joelbutcher:
@taylorotwell I hadn't, but it wouldn't take too much to implement this. I'm assuming developers would specify this in their `JetstreamServiceProvider`?

joelbutcher:
@taylorotwell Just pushed up that change 👍

taylorotwell:
What does the "update password" feature do when the user is authenticated with a social provider? Is the two factor management pane hidden when the user is authenticated using a social provider?

taylorotwell:
Also, what if email verification is enabled for email based users? What happens to social users?

driesvints:
And what about "reset password"? (maybe I missed the convo about that one)

joelbutcher:
None of those features have been touched as I was unsure what the desired outcome should be?

Email verification, in my opinion, should behave in the same way regardless of whether the user registered via email or socialite (I use the email returned from socialite and populate the users email from this)

taylorotwell:
We should probably just change password reset to always return some message like "If an account exists with that email address you will receive an email soon..." etc... 

Update password and two factor authentication management panels should be totally hidden if the user is authenticated socially.

driesvints:
@taylorotwell I'm not sure those two should be mutually exclusive. For example, on Laravel.io you can first register using Github but afterwards set a password if you want. Both allow you to login.

taylorotwell:
Is that how most social authed applications work? Should we just leave both update password and 2FA screens available? The update password screen would perhaps need some wording or something to indicate that if they set a password they will be able to login with both their connected accounts and their email? I dunno.

@joelbutcher said earlier that he thought 2FA should be the social service's responsibility. I'm not sure if this PR is setup to handle 2FA from a social authed account?

joelbutcher:
@taylorotwell I was going off my experience tbh, I don't think there's a defined way of doing 2FA. Regardless, the users email is always populated whether signing/logging in with email or via Socialite. The only issue there is the 2FA confirmation dialogue - the 2FA feature currently requires a user to enter password before being able to enable 2FA.

I've also found that there is an additional complication with hiding those panels as the logic becomes more complex. We've got three conditions we need to check for, making the `@if` statements look horrible 😂.

1. App supports socialite & user signed up with email
2. App supports socialite & user signed up with social provider
3. App doesn't supports socialite

At the end of the day, I'm happy, willing, and able to implement whatever the general consensus is...

taylorotwell:
Checking on GitLab... I sign in via GitHub... 

They show me a password update screen but I don't even know what my current password would be? I'm not sure if this is a bug on their end.

They also allow me to enable / disable 2FA but do not require me to confirm my password to do so.

driesvints:
> Is that how most social authed applications work?

I don't think there's a golden rule here. DigitalOcean allows only one login type: either email/password or one single social login (not multiple like this PR). Algolia allows a combination of email/password and social accounts.

Imho I think a combination makes most sense. You add different ways of logging in so you have an alternative if you ever get locked out of one.

Algolia requires your 2FA from within Algolia even if you login with a social account. I don't know the general consensus on this.

taylorotwell:
Fortify supports enabling 2FA without password confirmation via the feature flag option but the Jetstream mark-up would have to be changed to respect this I believe.

joelbutcher:
I've just tested with Medium. I already had an account through Facebook OAuth. But when signing up with email, using the same email address that's connected to Facebook, it created me a new account. Medium also don't require a password as login is all done via 'magic' email links...

taylorotwell:
It's a little odd to me to add a password to a socially connected account. Isn't that one of the main points of logging in socially? To not have to give the service your password? Does this PR allow a password authenticated user to connect a social account? I guess they would have to do that via the login screen and make sure the email on their social service matched their existing Jetstream email?

joelbutcher:
@taylorotwell I'd probably disable 2FA if the user is signed up via social OAuth, just to remove the initial headache of enabling/disabling 2FA without having a password set? Otherwise it defeats the security benefits of 2FA?

> Does this PR allow a password authenticated user to connect a social account? 
Yes it does. It's done via login/registration. The `AuthenticatesSocialite` handles this logic

joelbutcher:
One possible route to go down, is to allow the user to remove all connected social accounts, but require the user (re)set their password before removing the last one.

Currently, password reset works for all users (regardless of whether or not they have a connected social account).

joelbutcher:
@driesvints @taylorotwell The latest commit just hides both panels if the app supports socialite and the user has any connected accounts. The final thing that I needs addressing/figuring out is the password confirmation dialog for the "delete account" feature...

taylorotwell:
I'll think about the questions above. If a user is doing login via GitHub for example - how do they specify their scopes? Is it possible to allow sign-in **only** through social apps if you want to disable email sign up entirely?

joelbutcher:
@taylorotwell Scopes is something that I hadn't considered with this as I was going with the assumption of the socialite feature just being used for OAuth. 

I haven't developed the ability to remove email sign up/login, but I'll look into adding this as I think it'll be something a number of developers would like to have the option to remove

routes/inertia.php
```diff
@@ -57,4 +65,10 @@
                         ->name('team-invitations.destroy');
         }
     });
+
+    // Socialite...
+    if (Jetstream::hasSocialiteFeatures()) {
+        Route::get('/socialite/{provider}', [SocialiteController::class, 'redirectToProvider'])->name('socialite.redirect');
```
dillingham, line 71: If you loop `socialite.providers` to define the routes it does a few things:

- makes a cleaner use of route helpers
 ```
route('socialite.github')
```
```
route('socialite.redirect', ['provider' => 'github'])
```

You can set the parameter like so: `Route::get(.., ...)->defaults('provider', $provider);`

- It also 404s instead of 500s from passing a unknown provider to `Socialite::driver($anythingGoesHere)`

**Comment by dillingham** [comment]
> I guess they would have to do that via the login screen and make sure the email on their social service matched their existing Jetstream email?

If they are logged in, it should associate regardless of mismatching email
https://github.com/laravel/jetstream/issues/214#issuecomment-724103218

**Comment by joelbutcher** [comment]
@dillingham Some good points, I've just pushed a big update to the branch that allows users who registered with email to connect their to support social providers.

@taylorotwell there were a couple of missing feature flags in the inertia stack that are present in the Livewire stack "profile information" and "update passwords", so I've added them in here. Let me know if you'd prefer I take them out and make a separate PR for that.

File: src/Http/AuthenticatesSocialite.php [hunk 7]
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Find a user from a given Socialite provided user, or create a new
+     * one from the details given by the provider.
+     *
+     * @param  \Laravel\Socialite\Contracts\User  $providerUser
+     * @return \Illuminate\Contracts\Auth\Authenticatable
+     */
+    protected function getUser(SocialiteUserContract $providerUser)
+    {
```
dillingham, line 51: If they connect from settings, I think this would avoid email mismatch issues.

```suggestion
    {
        if(auth()->check()) {
            return auth()->user();
        }
```

joelbutcher:
@dillingham Nice spot - cheers! 👍 

I've just pushed a few more tweaks. I've added a new "set password" form that will show if the user hasn't got a password and socialite is enabled. Which **SHOULD** get around most of the issues relating to password protected forms (2FA, Log out of other browser sessions etc.

@taylorotwell sorry if that is overkill 😅

driesvints:
@joelbutcher I've merged 1.x into master which will fix those permission issues. Can you rebase?

driesvints:
Maybe it's also best that you squash your commits.

joelbutcher:
@driesvints All squashed 👍

taylormaguire:
> I'll think about the questions above. If a user is doing login via GitHub for example - how do they specify their scopes? Is it possible to allow sign-in **only** through social apps if you want to disable email sign up entirely?

@taylorotwell It is not uncommon for some enterprise apps to support both local authentication and social/single-sign-on auth at the same time, and there can be benefits to this. For example if the social/sso auth is down you can still authenticate locally, especially important if you are an admin on the team. SSO is really just another method of authentication, it doesn't have to be one or the other. 

you could leave all the existing password, 2fa, and user profile stuff as is, but when SSO is enabled provide some kind of notice that those settings affect Local auth only.

driesvints:
@joelbutcher please don't merge master into this branch and use rebasing instead. Otherwise we won't be able to squash commits.

joelbutcher:
@driesvints My bad, I resolved a couple conflicts in Github's UI. I've just force pushed a fix.

routes/inertia.php
brentarcane: You can still pass in parameters to the `route` helper, for example, `route('socialite.redirect', 'github')`.

taylorotwell:
@joelbutcher would you mind making a summary post of where the code stands after all these discussions, summarizing what is available / not available in various in situations such as when a user is socially authed vs email authed... how they transition between the two, etc.

joelbutcher:
@taylorotwell Sure!

### Feature:
- Socialite support feature toggle
  - Specify your supported providers by passing options in `config/jetstream.php`
```php
'features' => [
    Feature::socialite(['facebook' => true, 'github' => true]);
]
```

### Email authed users
- Connect to given providers via `user/profile'
- As user was originally created with email/password. 2FA, update password, delete account all remain.

### New socially authed users
- Register/Login perform the same action to create a user if a one with the email attached to the providers user doesn't already exist.
- Replaces 'Update password' with a 'Set password' panel.
- Hides 2FA and delete account features until the user sets a password.

### TODO/Unsure
- Password confirmation modal for other browser sessions - I feel the feature should remain, but hesitant to make **another** breaking change and remove the password confirmation if a socially authed/created user hasn't set a password yet

taylorotwell:
It is impossible to revoke browser sessions across all drivers in Laravel without a password.

joelbutcher:
It does not - I'll hide that panel if the user hasn't got a password

taylorotwell:
Are there things that could potentially be broken into sub-components?

joelbutcher:
@taylorotwell sorted it, just reduced the logic to a per component/form basis

joelbutcher:
@taylorotwell @driesvints I've just push two fixes to the logic used to authenticate a user login in via a social provider. I have a demo project available at [joelbutcher/jetstream](https://github.com/joelbutcher/jetstream-demo). There's a link to a live demo in the repo If you want to check it out!

driesvints:
@joelbutcher thanks for all your work so far 👍

driesvints:
@joelbutcher when I pressed "not now" in the facebook oauth screen I was redirected to a 500 error screen.

joelbutcher:
@driesvints Sorted 👍 

I've put a check in to handle errors and either redirect to the 'register' route show the error message, or display the 'dangerBanner' macro added in #388

database/migrations/2020_05_21_400000_create_connected_accounts_table.php
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::drop('connected_accounts');
```
ZeProf2Code, line 36: dropIfExists?

ZeProf2Code (changes requested):

ZeProf2Code:
I noticed some problems:
- You can link the same social account on two different users.
- If user is created with OAuth, no personal team will be created causing a 500 error
- If user is created with OAuth, no verification email will be sent.

stubs/livewire/resources/views/profile/connected-accounts-form.blade.php
```diff
@@ -0,0 +1,70 @@
...
+            {{ __('You are free to connect any social accounts to your profile and may remove any connected accounts at any time. If you feel any of your connected accounts have been compromised, you should disconnect them immediately and change your password.') }}
+        </div>
+
+        <div class="mt-5 space-y-6">
+            @foreach ($this->providers as $provider => $enabled)
+                @if ($account = $this->accounts->where('provider_name', $provider)->first())
+                    <x-jet-connected-account provider="{{ $account->provider_name }}" created-at="{{ $account->created_at }}">
+
+                        <x-slot name="action">
+                            @if ($this->accounts->count() > 1)
```
ZeProf2Code, line 29: `@if (! is_null($user->password) || $this->accounts->count() > 1)`

ZeProf2Code (changes requested):

stubs/livewire/resources/views/profile/connected-accounts-form.blade.php
joelbutcher: What does this resolve? This solution will have unwanted affects in that users who have signed up with OAuth, cannot connect to other providers?

joelbutcher: What does this resolve exactly? If I've signed up with Facebook, but later decided to connect Github and remove Facebook (and NOT set a password) this will not allow me to remove Facebook

ZeProf2Code: It should allow people who set a password to remove all their OAuth 2 accounts.

joelbutcher: I don't disagree, but I think that condition should be flipped so that users can remove any connected accounts, if they have more than one. An only be able to remove the last one if they've got a password set

joelbutcher:
@ZeProf2Code 

Just addressing some of these issues now 👍.

> You can link the same social account on two different users.

**Q:** Is there a problem here? If, for example, I've signed up with `name@gmail.com` and connected my Facebook account for easier login, but then create another account at, say `name@work-domain.com` then also connect Facebook just for easier login. I wouldn't associate one connected account with both users, each user would have a connected account record, thus producing two (seemingly identical) records in the database?

dillingham:
@joelbutcher hmmm 🤔 sounds like the issue would be deciding which of the two users associated with that Facebook record to authenticate when you click the facebook login icon. "That account is associated with a user" might be needed during the connect / logged in flow. Signup / login should just associate it with an email of an existing user.

```php
$existing = ConnectedAccount::firstWhere([
    'provider_id' => $providerUser->getId();
    'provider_name' => 'facebook;
]);

if($existing && Auth::check() && Auth::user()->id != $existing->user_id) {
    //  do not pass go, do not collect $200
}
```

joelbutcher:
@ZeProf2Code I believe the last two commits will address the issues you've raised.

@joelbutcher thanks for this - I've added the logic into the controller! 👍

joelbutcher:
@ZeProf2Code I believe the last commits will address the issues you've raised.

@joelbutcher thanks for this - I've added the logic into the controller! 👍

joelbutcher:
@driesvints that demo project has had a clean install if you wish to do a number of tests?

ZeProf2Code:
@joelbutcher When a user has set a password, you should let them delete all their OAuth accounts, right?

joelbutcher:
@ZeProf2Code Yup! Just sorted that - I had it in my test project but forgot to pull those changes into the fork

lexdewilligen:
@joelbutcher awesome PR! Can you rebase it and make it tailwindcss 2 ready? They've just merged #488

joelbutcher:
@lexdewilligen sure thing!

database/migrations/2020_05_21_400000_create_connected_accounts_table.php
ZeProf2Code: was it reverted?

src/Features.php
```diff
@@ -135,6 +145,21 @@ public static function termsAndPrivacyPolicy()
...
+    /**
+     * Enabled the socialite feature.
+     *
+     * @param  array  $options
+     * @return string
+     */
+    public static function socialite(array $options = [])
+    {
+        if (! empty($options)) {
+            static::$featureOptions['socialite'] = $options;
```
ZeProf2Code, line 157: config(['jetstream-options.socialite' => $options]);

joelbutcher:
@driesvints I've rebased to include the latest master. I've also squashed all the commits into one

ZeProf2Code:
Is it normal that createUsersFromProviderUsing doesn't exists anymore?

joelbutcher:
@taylorotwell @driesvints Squashing these commits screwed up the rebase - I've just reset the branch back to what it was AFTER the rebase, but BEFORE the squash. This PR is getting quite complicated now - would you prefer if I closed this an opened a new one?

> @joelbutcher Is it normal that createUsersFromProviderUsing doesn't exists anymore?

@ZeProf2Code This should have been fixed with the latest force-push.

src/Http/Controllers/SocialiteController.php
```diff
@@ -0,0 +1,136 @@
...
+            return redirect(config('fortify.home'));
+        }
+
+        return redirect(config('fortify.home'))->banner(
+            __('You have successfully connected '.ucfirst($provider).' to your account.')
+        );
+    }
+
+    /**
+     * Find a user from a given Socialite provided user, or create a new
```
joelbutcher, line 122: Code comment needs updating

src/Jetstream.php
```diff
@@ -332,6 +383,17 @@ public static function createTeamsUsing(string $class)
         return app()->singleton(CreatesTeams::class, $class);
     }
 
+    /**
+     * Register a class / callback that should be used to create users from social providers.
+     *
+     * @param  string  $class
+     * @return void
+     */
+    public static function createUsersFromProviderUsing(string $class)
```
joelbutcher, line 392: Singular "User"

src/JetstreamServiceProvider.php
```diff
@@ -93,6 +99,10 @@ public function boot()
         if (config('jetstream.stack') === 'inertia') {
             $this->bootInertia();
         }
+
+        if (Jetstream::hasSocialiteFeatures()) {
```
joelbutcher, line 103: Can be removed as

```php
Jetstream::createUsersFromProviderUsing()
```

Is now in the published `JetstreamServiceProvider` stubs

stubs/app/Actions/Jetstream/DeleteUser.php
```diff
@@ -2,10 +2,30 @@
...
+     */
+    protected $deletesTeams;
+
+    /**
+     * Create a new action instance.
+     *
+     * @param  \Laravel\Jetstream\Contracts\DeletesTeams  $deletesTeams
+     * @return void
+     */
+    public function __construct(DeletesTeams $deletesTeams)
```
joelbutcher, line 24: Changes in this file need reverting

stubs/app/Actions/Jetstream/DeleteUserWithTeams.php
```diff
@@ -56,4 +58,17 @@ protected function deleteTeams($user)
...
+
+    /**
+     * Delete the connected accounts and account associations attached to the user.
+     *
+     * @param  mixed  $user
+     * @return void
+     */
+    protected function deleteConnectedAccounts($user)
+    {
+        if (method_exists($user, 'connectedAccounts')) {
```
joelbutcher, line 70: @taylorotwell whats best practice for Laravel here? `method_exists` doesn't care about the methods visibility. Would you prefer `is_callable`?

stubs/inertia/resources/js/Jetstream/ConnectedAccount.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+    <div class="px-3 flex items-center justify-between">
+        <div class="flex items-center">
+
+            <jet-facebook-icon class="h-6 w-6 mr-2" v-if="provider === 'facebook'"/>
```
joelbutcher, line 5: @driesvints should the order of attributes here have the `v-if` first?

stubs/inertia/resources/js/Pages/Profile/ConnectedAccountsForm.vue
```diff
@@ -0,0 +1,144 @@
...
+        <template #title>
+            Connected Accounts
+        </template>
+
+        <template #description>
+            Manage and remove your connect accounts.
+        </template>
+
+        <template #content>
+            <h3 class="text-lg font-medium text-gray-900" v-if="accounts.length === 0">
```
joelbutcher, line 12: @driesvints same question here - order of attributes?

joelbutcher (commented):
Needs changes before merge

driesvints:
Hey @joelbutcher. @taylorotwell will review this when he finds the time.

database/migrations/2014_10_12_000000_create_users_table.php
```diff
@@ -18,7 +19,9 @@ public function up()
             $table->string('name');
             $table->string('email')->unique();
             $table->timestamp('email_verified_at')->nullable();
-            $table->string('password');
+            $table->string('password')->nullable(
+                Features::hasSocialiteFeatures()
```
SimBig, line 23: what happens if someone enables SocialiteFeatures after the migration did run?
Would if make sense to make this always nullable and check the nullable rule on the saving of a new user?

lexdewilligen:
I think I've also found several small issues:

- route 'user-password.set' is not in the route list
- if you've not connected another socialite service, it says "Added on Not connected", maybe just "Not connected" would be better
- Pages/Profile/Show.vue doesn't accept the hasPassword prop yet

src/Http/Middleware/ShareInertiaData.php
```diff
@@ -37,6 +37,11 @@ public function handle($request, $next)
                     'hasTermsAndPrivacyPolicyFeature' => Jetstream::hasTermsAndPrivacyPolicyFeature(),
                     'locale' => app()->getLocale(),
                     'managesProfilePhotos' => Jetstream::managesProfilePhotos(),
+                    'canUpdateProfileInformation' => Features::enabled(Features::updateProfileInformation()),
+                    'canUpdatePassword' => Features::enabled(Features::updatePasswords()),
```
lexdewilligen, line 41: The `canUpdatePassword` prop is already set above, same goes for `canUpdateProfileInformation`, `managesProfilePhotos`, `canUpdateProfileInformation`. Is this on purpose? Implementations are almost the same

lexdewilligen (changes requested):

database/migrations/2014_10_12_000000_create_users_table.php
joelbutcher: The idea with Jetstream is that it's a scaffold for new app to get up and running easily with auth. Since this feature is optional, you would need to determine whether or not your app/product/service requires socialite feature prior to release. Nots ping so, IMO, would be the improper way to start a project

joelbutcher:
@lexdewilligen done!

> And maybe this is just a problem

This sounds like you're running your local environment through `https`, but have the callback setup in Github as `http`?

stubs/inertia/resources/js/Pages/Profile/SetPasswordForm.vue
```diff
@@ -0,0 +1,74 @@
...
+
+        <template #form>
+            <div class="col-span-6 sm:col-span-4">
+                <jet-label for="password" value="New Password" />
+                <jet-input id="password" type="password" class="mt-1 block w-full" v-model="form.password" autocomplete="new-password" />
+                <jet-input-error :message="form.error('password')" class="mt-2" />
+            </div>
+
+            <div class="col-span-6 sm:col-span-4">
+                <jet-label for="password_confirmation" value="Confirm Password" />
```
bonzai, line 19:
```suggestion
                <jet-label for="password_confirmation" :value="$t('Confirm Password')" />
```

stubs/inertia/resources/js/Pages/Profile/SetPasswordForm.vue
```diff
@@ -0,0 +1,74 @@
...
+                <jet-input id="password_confirmation" type="password" class="mt-1 block w-full" v-model="form.password_confirmation" autocomplete="new-password" />
+                <jet-input-error :message="form.error('password_confirmation')" class="mt-2" />
+            </div>
+        </template>
+
+        <template #actions>
+            <jet-action-message :on="form.recentlySuccessful" class="mr-3">
+                Saved.
```
bonzai, line 27:
```suggestion
                {{ $t('Saved.') }}
```

stubs/inertia/resources/js/Pages/Profile/SetPasswordForm.vue
```diff
@@ -0,0 +1,74 @@
...
+            </jet-action-message>
+
+            <jet-button :class="{ 'opacity-25': form.processing }" :disabled="form.processing">
+                Save
```
bonzai, line 31:
```suggestion
                {{ $t('Save') }}
```

stubs/inertia/resources/js/Pages/Profile/ConnectedAccountsForm.vue
```diff
@@ -0,0 +1,144 @@
...
+                return null;
+            },
+
+            removeConnectedAccount(id) {
+                this.form.post(`/user/connected-account/${ id }`, {
+                    preserveScroll: true
+                }).then(response => {
+                    if (!this.form.hasErrors()) {
+                        this.confirmingRemove = false;
+                    }
```
bonzai, line 139: https://inertiajs.com/manual-visits#promise-deprecation

```suggestion
                    onSuccess: () => {
                        if (! this.form.hasErrors()) {
                            this.confirmingRemove = false;
                        }
                    },
```

stubs/inertia/resources/js/Pages/Profile/SetPasswordForm.vue
```diff
@@ -0,0 +1,74 @@
...
+            Set Password
+        </template>
+
+        <template #description>
+            Ensure your account is using a long, random password to stay secure.
+        </template>
+
+        <template #form>
+            <div class="col-span-6 sm:col-span-4">
+                <jet-label for="password" value="New Password" />
```
bonzai, line 13:
```suggestion
                <jet-label for="password" :value="$t('New Password')" />
```

stubs/inertia/resources/js/Pages/Profile/UpdateProfileInformationForm.vue
```diff
@@ -50,8 +50,8 @@
 
             <!-- Email -->
             <div class="col-span-6 sm:col-span-4">
-                <jet-label for="email" :value="$t('Email')" />
-                <jet-input id="email" type="email" class="mt-1 block w-full" v-model="form.email" />
+                <jet-label for="email" value="Email" />
```
bonzai, line 53:
```suggestion
                <jet-label for="email" :value="$t('Email')" />
```

src/Http/Livewire/ConnectedAccountsForm.php
```diff
@@ -0,0 +1,113 @@
...
+    public function removeConnectedAccount($accountId)
+    {
+        DB::table('connected_accounts')
+            ->where('user_id', Auth::user()->getAuthIdentifier())
+            ->where('id', $accountId)
+            ->delete();
+
+        $this->confirmingRemove = false;
+
+        $this->banner('Account removed');
```
bonzai, line 84:
```suggestion
        $this->banner(__('Account removed.'));
```

bonzai (changes requested):

lexdewilligen:
@joelbutcher awesome, thank you for the update. The `hasAccountDeletionFeatures` is also a duplicate.

src/Http/Controllers/Inertia/UserProfileController.php
```diff
@@ -21,6 +21,24 @@ public function show(Request $request)
...
+            'connectedAccounts' => $request->user()->connectedAccounts
+                ->map(function ($account) {
+                    return (object) [
+                        'id' => $account->id,
+                        'provider_name' => $account->provider_name,
+                        'created_at' => (new \DateTime($account->created_at))->format('d/m/Y H:i'),
+                    ];
+                }),
+            'hasPassword' => ! is_null($request->user()->password),
+            'socialiteProviders' => [
```
Garanaw, line 33: I have seen this array repeated a few times through the PR. Isn't it better to move it to `src/Features` instead? Something like

```
public static function mapToEnabledSocialProviders(): array
{
    return  [
        'facebook' => Jetstream::hasSocialiteSupportFor('facebook'),
        'google' => Jetstream::hasSocialiteSupportFor('google'),
         'twitter' => Jetstream::hasSocialiteSupportFor('twitter'),
         'linkedin' => Jetstream::hasSocialiteSupportFor('linkedin'),
         'github' => Jetstream::hasSocialiteSupportFor('github'),
         'gitlab' => Jetstream::hasSocialiteSupportFor('gitlab'),
         'bitbucket' => Jetstream::hasSocialiteSupportFor('bitbucket'),
    ];
}
```

Then, it can simply be called as `Features::mapToEnabledSocialProviders()` from anywhere avoiding the repetition and having a single source of truth

joelbutcher: Yup, I'll add something like this in later today. Cheers!

stubs/inertia/resources/js/Jetstream/ConnectedAccount.vue
```diff
@@ -0,0 +1,60 @@
...
+
+        <slot name="action"></slot>
+    </div>
+</template>
+
+<script>
+    import JetFacebookIcon from '@/Socialite/FacebookIcon';
+    import JetGoogleIcon from '@/Socialite/GoogleIcon';
+    import JetTwitterIcon from '@/Socialite/TwitterIcon';
+    import JetLinkedinIcon from '@/Socialite/LinkedinIcon';
```
irakan, line 36: LinkedinIcon should be LinkedInIcon

stubs/inertia/resources/js/Jetstream/Socialite.vue
```diff
@@ -0,0 +1,65 @@
...
+            </a>
+        </div>
+    </div>
+</template>
+
+<script>
+import JetFacebookIcon from '@/Socialite/FacebookIcon';
+import JetGoogleIcon from '@/Socialite/GoogleIcon';
+import JetTwitterIcon from '@/Socialite/TwitterIcon';
+import JetLinkedinIcon from '@/Socialite/LinkedinIcon';
```
irakan, line 45: LinkedinIcon should be LinkedInIcon

src/Http/Controllers/Inertia/UserProfileController.php
abr4xas: But,  if I only want to activate just one, how would I do it?

joelbutcher: You define it in your Jetstream config like so:

```php
'features' => [
    Jestream::socialite(['facebook' => true, 'github' => true]),
],
```

Any providers that are set to `true` will be displayed as options on your `login`, `register` routes as well as in the `user/profile` route.

You will need to define the tokens used in your `services.php` config in your Laravel project. 

An example project can be found [here](https://github.com/joelbutcher/jetstream-demo)

abr4xas: Thats nice!! Thanks!!

routes/inertia.php
```diff
@@ -64,4 +74,13 @@
...
+        Route::get('/socialite/{provider}/callback', [SocialiteController::class, 'handleProviderCallback'])->name('socialite.callback');
```
dillingham, line 81: Maybe `/oauth/{provider}` & `oauth.callback` 🤔, maybe a worthy refinement, maybe not.
Reasoning: `socialite` is the brand of the package, `oauth` is the functionality.

joelbutcher: @dillingham Thanks for this, just pushed it up! 👍

taylorotwell:
Sorry to ask this again @joelbutcher but can you give a bullet point summary of the features... what is disabled when authenticating socially? can users add a email / password even if they authenticate socially? if they do, what happens? all of those little details are really making this a bear.

joelbutcher:
@taylorotwell sure thing! I'll walk you through the flow:

1. Code deployed & migrations run - users table created with nullable password column
2. User registration via OAuth provider, creating a user with the users name & email address gathered from the providers profile data. Password is set to `null`
3. User navigates to `user/profile`
    - All forms requiring a password (2FA, account deletion etc.) are hidden
    - Users can remove all but one OAuth provider, IF they haven't yet set a password
    - Update your password form is replaced with a "set password" form
4. User sets password
    - All password-protected forms (2FA, Account deletion etc.) become available on refresh.
5. User connects another account
    - Irrespective of the email on the 2nd providers profile, the account may be connected to the user (If that providers profile ID is already associated with another user in the DB, a danger banner is shown after the callback, indicating that the account cannot be connected due to being already used by another user).


That should have covered everything. Login/Register via OAuth are effectively wired the exact same way behind-the-scenes.

stubs/inertia/resources/js/Jetstream/ConnectedAccount.vue
```diff
@@ -0,0 +1,60 @@
...
+    import JetGithubIcon from '@/Socialite/GithubIcon';
+    import JetGitLabIcon from '@/Socialite/GitLabIcon';
+    import JetBitbucketIcon from '@/Socialite/BitbucketIcon';
+
+    export default {
+        props: {
+            provider: String,
+            createdAt: {
+                type: String,
+                defaut: null,
```
lexdewilligen, line 46: Small typo: defaut => default

lexdewilligen (changes requested):

irakan:
A question: Shouldn't this PR be in Fortify instead of Jetstream? because:

> Laravel Fortify is a frontend agnostic authentication backend for Laravel. Fortify powers the registration, authentication, and two-factor authentication features of Laravel Jetstream.

joelbutcher:
Possibly, but it does require frontend elements. Especially the management panel in the profile route.

nicotravassos:
@joelbutcher it does make sense if this feature was part of Fortify. Currently Fortify has no frontend assets all the frontend assets reside within Jetstream and the backend auth logic sit within Fortify.

joelbutcher:
@nicotravassos I'm not sure I agree. Fortify is supposed to be a lightweight authentication package for Laravel, whereas Jetstream is application scaffolding, it just utilises fortify for authentication. This PR doesn't involve changing or adding any functionality to fortify, it just utilises the authentication mechanics and login pipeline. None of the logic to do with associating a user to a provider is the responsibility of fortify. 

I would agree with you if your point was to integrate socialite with Laravel Passport. I would also agree that, in that scenario, w custom `SocialGrantType` class and `SocialiteUserResolver` class would need to be created.

nicotravassos:
@joelbutcher I agree 👍🏻

taylorotwell:
@joelbutcher I think we talked about this before but is it not a security vulnerability to link accounts by email on multiple providers since some of these services may not verify emails on sign up. I could gain access to any account that way if that person doesn't already have an account on the provider?

nicotravassos:
@taylorotwell will jetstream v2 only be released once this pull request has been added or is there an ETA for V2 👯

taylorotwell:
@nicotravassos well, this PR is not *guaranteed* to get added. That's why we're still reviewing it 😄  ... but yes we need a decision on this PR or this feature in general for v2 before shipping. I would like to ship it before end of year.

dillingham:
If they are already logged in, unconfirmed email doesn't pose a security issue. So maybe just when there is an existing account.. requiring email confirmation.. Or maybe a config of services to require confirmation due to absence of. `'confirm_signup' => ['facebook']`

This is good idea regardless of multiple oauths for password resets.

nicotravassos:
Thanks @taylorotwell

joelbutcher:
@taylorotwell I think there was a discussion about that yeah, I can't remember what the conclusion was...

I'm not sure I follow your example. In my head, dropping the link between a provider and a user (using the email address on the provider) would require a paradigm shift in the way authentication in Jestream works, perhaps using a unique username instead of an email? (Forgive me if I'm wrong - it's late here in the U.K!)

If you could give me a scenario where you think a users account could be compromised, hit me up with the user flow and I'll try and "break" this PR and do what I can to resolve the problems

One approach I can think of off the top of my head? would be to generate a secure password string for users that are registering with a socialite provider for the first time and send this to the associated email address. That way, the users email address can be verified via a verification link sent in the same email?

I could alter the functionality so that linking users to multiple providers is only done from the connected accounts form in the `user/profile` route. But I think this would involve removing the uniqueness of an email to a user?

Let me know what you guys think needs to happen here and I'd be more than happy to ensure that the PR gets into an appropriate state to merge it in before you release Jetstream v2! 💪🏼

dillingham:
@joelbutcher from what I understood.. lets say I made a custom example.com provider for socialite.. and example.com lets users create an account with any email and doesn't verify the ownership of that email.. if you the real Joe Butcher had an account authenticated with Facebook on jetstream.app with `joebutcher@gmail.com`, I could create an account with `joebutcher@gmail.com` on example.com and then oAuth on jetstream.app and gain access as you, because example.com didn't verify.

jwahdatehagh:
In my mind this is quite simple:

If during signup via oAuth an email is returned that already exists, the user should get an _"An account with this email address already exists - please sign in through your initial provider"_-error **unless** they are already authenticated as the user associated to that email address.

Or am i missing something?

joelbutcher:
@jwahdatehagh I'm certainly a fan of that method! 

Going way back to Taylor's Spotify and my Medium example's registering with different provider associated with the same email address creates you a separate account every time. I'm not a huge fan of this, but it is one way around the problem. The trade off there is that it removes the uniqueness of an email address in the database, with would break the current login convention used by Jestream. 

@taylorotwell I think a good solution here would be to have registration & login check for an existing user with that email and check for any associated connected account records with the provider and provider ID. If none exists, return the error that @jwahdatehagh suggested. That way linking additional providers to an existing account. The only obvious issue then is that if a user is on the register route and looking to sign up, the error message needs to be different to the login route (as they know an account doesn't exist, that's why they're on the register route).

irakan:
- What happens when the provided user have an empty email field? 

Some providers like `Facebook` allows users to create an account using a phone number instead of an email.

Garanaw:
I see a couple issues with creating different accounts for the same email depending on the used IDP: 

- As @joelbutcher said, it removes the uniqueness of the email in the databse and breaks the current mechanism
- Any email address should only belong to a single person, therefore, IMO, a single user, no matter what IDP it comes from.

Also, a personal case I faced a few days ago, after clearing all my cookies/cache: I have an account on Medium which I created with Facebook. Now, when I tried to login after clearing my browser, they had an issue in that action (login users with some IDPs was returning a 500 error). I couldn't use my Facebook account to login, so I tried my Google account, but it wasn't returning errors, it just wouldn't let me login, so I had to go and reset my password just to be able to login with my email.

My point after telling this experience is that I'm still the same person, who have different accounts in different IDPs, all of them belonging to me, and these IDPs usually apply their own security steps (at least the ones added to this PR). As a user, I would like to be able to login to the same site using any of my IDPs, as they all represent me. **If, for whatever reason** (an internal error, a API's change of version, errors in the DNSs, or any possible attack) **the initial IDP blocks the user's authentication, login in to the site will also be blocked**. Users want things easy.

As a suggestion, I'd say that, if a user has an account associated to X IDP (let's say Facebook for the sake of the example), and then tries to login with Y IDP (Google for example), an extra security step should be applied (something like confirming the user's identity with any of the already associated IDPs, or maybe a 2fa, which can be enabled by default when registering with any IDP), but still allow that person to login to the same account if that security step passes (and also link that IDP to the same account behind the scenes?). This way, the databases will remain clean of extra users's accounts, and each user will have their IDPs bounded to a single account, thus representing the same person from different possible sources, and also @taylorotwell 's concern about gaining access to other user's account can also be covered.

@irakan has also a point there: some IDPs allow accounts without email, here I see a few options:
- Add a phone number as nullable unique field in the `users` table so it can be used in those cases
- Dissallow the account creation/login if an email is not provided by the selected IDP
- Add a `key` field in the `socialite.php` config file to represent the column/columns that will be used to log in users
- Use Fortify's config keys `username` and `email` to represent the possible columns to check when login in with an IDP (maybe group them in an array, so more can be added -phone-).

Thoughts?

dillingham:
@jwahdatehagh your suggestion is exactly what I meant and what my previous code example does 👍

dillingham:
@irakan good point, maybe just validate email presence: 

```php
if(! $providerUser->getEmail()) {
    return redirect(config('fortify.home'))->dangerBanner(
        __('Login with :provider failed due to no email.", ['provider' => $original->provider])
    );
```

joelbutcher:
@irakan 

> Some providers like Facebook allows users to create an account using a phone number instead of an email.

I thought Facebook allow creating an account with phone numbers for messenger only? Meaning that OAuth for that account via Facebook wouldn't work?

dillingham:
@joelbutcher confirmed, just created a facebook account using only a google voice number. `facebook.com/example.dillingham`. if you want & enable fb on your example install of jetsteam, I can authenticate it for you.

irakan:
> @irakan good point, maybe just validate email presence:
> 
> ```
> if(! $providerUser->getEmail()) {
>     return redirect(config('fortify.home'))->dangerBanner(
>         __('Login with :provider failed due to missing email.", ['provider' => $provider])
>     );
> }
> ```

This is the easy way to do it or maybe if the email is not present the user should be redirected to a "email confirmation" screen where the user enters an email and confirm it before moving forward.

This is not for `Facebook` only, also `Twitter` allow users to create an account using a phone number instead of an email.

ZeProf2Code:
This is not how it must be done. All real-world apps ask the users to connect to the their account before linking.

irakan:
Here's how Facebook and Twitter APIs response regarding to the email field.

[Facebook](https://developers.facebook.com/docs/graph-api/reference/user)
> The User's primary email address listed on their profile. This field will not be returned if no valid email address is available.

[Twitter](https://developer.twitter.com/en/docs/twitter-api/v1/accounts-and-users/manage-account-settings/api-reference/get-account-verify_credentials)
> If the user does not have an email address on their account, or if the email address is not verified, null will be returned.

So this means the email field will be returned only when its verified. So when creating a user, the `email_verified_at` field should be filled.

bilfeldt:
There are two distinct situations in my view:

1. **Connecting an OAuth provider to an existing account:** This is done when the user is logged in and should not require an email (or that it matches the users email for that matter)
2. **Registering a new user account using an OAuth provider:**
    A. **Email exists:** Redirect to login with an error _"An account with this email address already exists - please sign in first before linking this provider to the account"_
    B. **All required information available (including email) and email does not exist already:** Create account and connect to the provider
    C. **Not all required information available (like email):** Show extra registration step to manually input the missing data before creating user.

The above solves: 
- Security issue with OAuth providers (potentially) not verifying email - raised by @taylorotwell [here](https://github.com/laravel/jetstream/pull/444#issuecomment-739575222)
- Issue with OAuth providers that do not require email (like Facebook and Twitter) - raised by @irakan [here](https://github.com/laravel/jetstream/pull/444#issuecomment-743755445)
- Issue with missing required user data that could have been added to the registration - example could be to require a  telephone number doing signup. 

One drawback is that an existing user that signed up either manually or using provider A cannot login using another provider without having connected that provider to the account first (while logged in).

routes/inertia.php
```diff
@@ -64,4 +74,13 @@
...
         }
     });
+
+    // Socialite...
+    if (Jetstream::hasSocialiteFeatures()) {
+        Route::get('/oauth/{provider}', [SocialiteController::class, 'redirectToProvider'])->name('oauth.redirect');
+        Route::get('/oauth/{provider}/callback', [SocialiteController::class, 'handleProviderCallback'])->name('oauth.callback');
+
+        // Set user password
+        Route::put('/user/set-password', [PasswordController::class, 'store'])->name('user-password.set');
```
dillingham, line 84: Maybe this could turn into a more general concept.

```
Route::get('complete-profile', [CompleteProfileController::class, 'show'])->name('complete-profile.set');
```
```
Route::put('complete-profile', [CompleteProfileController::class, 'update'])->name('complete-profile.set');
```
Maybe session flags (missing_email, missing_password) set during auth flow determine fields shown.

Maybe a middleware to redirect back to this route if incomplete

joelbutcher:
> if you want & enable fb on your example install of jetsteam, I can authenticate it for you.

@dillingham  done!

joelbutcher:
@bilfeldt Thanks for that - really helpful to have people looking over this and making those sorts of suggestions/recommendations! 🎉

joelbutcher:
@taylorotwell Don't wait for this PR to release Jetstream v2. Also, If you guys don't necessarily want to include this in Jetstream, I'm happy to release this as a package to go alongside Jetstream, but it definitely won't be ready for v2 release

dillingham:
To me, the 2 found edge cases are small enough in their impact that they don't require `perfect` solutions now and can be refined later. The majority benefit hugely outweighs the tiny edge cases. Plus, the simple solutions are totally reasonable.

- oauth without an email gets told to login with a service that has an email (docs: oauth **requires** email)
- oauth login with provider that doesn't match first provider gets told to use original provider.

```php
if(Auth::check()) {
    return; // only applies to login/register
}

if(! $providerUser->getEmail()) {
    return redirect(config('fortify.home'))->dangerBanner(
        __('Your :provider is missing an email.", ['provider' => $provider])
    );
}

$original = ConnectedAccount::query()
    ->where('email', $providerUser->getEmail())
    ->orderBy('created_at')
    ->first();

if($original && $original->provider != $provider) {
     return redirect(config('fortify.home'))->dangerBanner(
        __('Must login with :provider", ['provider' => $original->provider])
    );
}
```

dillingham:
Another edge case 😅.. updating profile email would create a new account in future login attempts.

Maybe store `oauth_id` on `users` table that references the original connected account.

src/Http/Controllers/SocialiteController.php
```diff
@@ -0,0 +1,136 @@
...
+
+        if ($existing && Auth::check() && Auth::user()->id != $existing->user_id) {
+            return redirect(config('fortify.home'))->dangerBanner(
+                __('That account is already associated with a user. Please use a different account.')
+            );
+        }
+
+        $user = $existing
+            ? $existing->user
+            : ($this->getUser($providerUser) ?? $this->createsUser->create($provider, $providerUser));
```
dillingham, line 106:
```php
if (Auth::check()) {
    $user = Auth::user();
} else if($existing) {
    $user = $existing->user;
} else {
    $user = $this->createsUser->create($provider, $providerUser);
}
```

Avoids the email check of `getUser()`

src/Http/Controllers/SocialiteController.php
```diff
@@ -0,0 +1,136 @@
...
+
+        $this->connectToProvider($user, $provider, $providerUser);
```
dillingham, line 108: This duplicates the query for finding `$existing`
Maybe should be combined with `createsUser`

src/Http/Controllers/SocialiteController.php
dillingham: Also `$account` might be more readable than `$existing`

database/migrations/2014_10_12_000000_create_users_table.php
dillingham: @joelbutcher we would have to note to migrate after setting up config in docs. 2fa is has a separate migration for their table adjustments, maybe be good to follow suit.

joelbutcher:
@taylorotwell I've just done quite a big refactor to the controller logic, taking into account all the above scenarios. Here's the lowdown:

## Registration

1. Registering for the first time with a provider that has an email address and a user for that email doesn't exist in the database, creates a new user and logs them in.
2. Registering with a provider that already exists, returns to `/register` with the following error:

Screenshot 2020-12-17 at 11 39 41

3. Registering with a provider with no email associated with the account returns to `/register` with the following error:

Screenshot 2020-12-17 at 11 39 28


## Login

1. Attempting to log in using a provider that doesn't exist for the account ID and provider name, returns to `/login` with the following error:

Screenshot 2020-12-17 at 11 39 53

2. Attempting to log in using a provider that does exist for the account ID and provider name, fetches the user the account is associated with in the DB and logs them in, redirecting them to the `fortify.home` route.

## Authenticated (connecting to an existing, logged-in user)

1. Attempting to connect an account that is already associated with another user, redirects to the `profile.show` route with the following danger banner error message:

Screenshot 2020-12-17 at 11 40 53


# Feature Recap

All important logic for OAuth registration, login and connection can be found in `\Laravel\Jetstream\Http\Controllers\SocialiteController@handleProviderCallback` (method name subject to change, if requested)

- Alters the `create_users_table` migration to set the password field as nullable, if the Jetstream is installed with the `--socialite` option.
- Registering via OAuth creates a new user using the email address on the provider account (provider accounts with no email address associated to them cannot register)
    - If Jetstream is installed with `--socialite` AND `--teams` options, registering also creates a personal team.
- If you've registered via OAuth and NOT set a password, all forms that require a password in `profile.show` are hidden/disabled (Delete Account, 2FA, Log out Browser Sessions etc.).
- A "Set Password" form replaces the "Update Password" form.
- Once a user has set a password using the new "Set Password" form, all previously hidden/disabled forms are available (Delete Account, 2FA, Log out Browser Sessions etc.)

## Update
- Users can update their email at any time, regardless of whether they've sign up with an OAuth provider.

irakan:
@joelbutcher keep up the good work.. one thing I think is missing here is the email verification...

@taylorotwell already mentioned this:
> Also, what if email verification is enabled for email based users? What happens to social users?

When creating a user I think the `email_verified_at` field should be filled since the social platform has verified it already for us...

joelbutcher:
@irakan yup, all done ✅

taylorotwell:
I just can't really take on the maintenance of this right now. Sorry - I may try to implement it myself for some future 3.x release.

joelbutcher:
@taylorotwell well I'd be lying if I said I wasn't a tad annoyed considering the number of people who have helped me get this PR to the stage it's in, the discussions taken place etc. But I understand the maintenance burden this also poses for you and the team. 

If it's okay with you guys, I think I'll look into sticking this into a package to compliment Jetstream.

Thanks for taking the time out for the discussions on this, valuable questions asked and lessons learned whilst developing this.

jwahdatehagh:
@joelbutcher - your patience in this was astounding. Hats off to you - and of course go ahead and stick it into a package - i wouldn't think you need anyones permission to do that 😅

joelbutcher:
@jwahdatehagh thanks! A lot of the patience was only face value - it is quite frustrating to work for a month on something that ends up not being used, or effectively being told the work isn't good enough.

A couple of changes I'm thinking of making before I release this as a package include:

- Moving the controller logic to a published action. 
- Utilising Authorization exceptions better.

What are peoples thoughts?

driesvints:
@joelbutcher We're very sorry about this but as Taylor said this would place quite a big maintenance burden on us. The reason why it took this long was because we were genuinely considering this to be merged into the upcoming 2.x release until the very end. 

> it is quite frustrating to work for a month on something that ends up not being used, or effectively being told the work isn't good enough.

Unfortunately this is also part of open source. We as maintainers need to decide how much work we want to take upon us while also still being able to put in work into all our other projects and apps. This means making these tough decisions. Also: nobody said your own isn't good enough. Imo this PR looks good and definitely could be released as a third party package.

As a closing note I want to say to everyone who's disappointed that some PRs aren't getting merged: Jetstream is completely open source. Feel free to fork the package and start a "Jetstream Extended" package yourself if you're willing to take up the maintenance work of all the extra features.

FoksVHox:
@driesvints Could you elaborate on this?
> this would place quite a big maintenance burden on us. The reason why it took this long was because we were genuinely considering this to be merged into the upcoming 2.x release until the very end.

Personally, right now I don't see how this could be "a big maintenance burden" as it's an integration with a 1st party package.

driesvints:
@FoksVHox this adds about 1900 extra lines of code.

FoksVHox:
> this adds about 1900 extra lines of code.

@driesvints While that being true, I doubt that those lines need to be changed frequently, but anyhow thanks for the reply.

dillingham:
@joelbutcher while Im sure it stung, it wasn't for nothing. Alot was discovered and I think this will be a very popular package. I for one am very looking forward to it and happy to contribute as best I can.

@taylorotwell Im not envious of having to make the hard call. But it does seem like tons of edge cases and might be rushed for an upcoming release. Tons of blow back potential on twitter if things werent smooth enough lol

@driesvints if jetstream was a little more extensible, might not need to maintain a fork. For example, I think the only thing this oAuth package would need is for Jetstream to have some blade `@stack` for views (https://github.com/laravel/ideas/issues/2439) and a `Jetstream.booting((Vue) => {})` to allow adding components to the Jetstream Vue instance (Like nova does). This would make it easy to displace the maintenance burden while also testing these features in the wild and discovering bugs before making them official.

driesvints:
@dillingham I'm not sure if that's a viable solution as you cannot guarantee the order of elements from the views you're stacking to/from. Also: it's not guaranteed that the source code from the parts you're extending from remains the same.

dillingham:
@driesvints fair point. In the event of collision and the order not being exact, the consumer can opt into manually adding components and maybe packages have a flag `config('jetstream-oauth.automatic')`. To the second point, yeah would have to avoid getting too specific with the stacks. But `login-form.after` would always be there, as would `profile-panels.after`

Just food for thought, Im sure adding a few components to views wouldn't be that complicated for users.

joelbutcher:
@taylorotwell @driesvints do you guys have any plans to open GitHub Discussions on this repo? Would give a central location for everyone to talk about this feature. 

Side note: I've done the Livewire version for Jetstream v1.6, but I'm getting an issue in Inertia stating that `props` is undefined when attempting to retrieve `$page.props.jetstream` and `$page.props.user` - any ideas? 

Thanks in advance!

driesvints:
No plans for discussions sorry. We tested it on the framework repo but felt it wasn't exactly working in the way we'd like. We won't be rolling it out on other repos.

The inertia error is probably because you updated to a newer inertia version that 1.x doesn't supports. V2 will fix this.

joelbutcher:
Just issued the first pre-release version v0.0.1 of this package [here](https://github.com/joelbutcher/socialstream).

I intend to ship v1 on Monday and work on v2 for Jetstream v2 after that. If you have any issues, hit me up in Discord or add the issue to the repo

driesvints:
Great work on that @joelbutcher 👍

MohannadNaj:
Just sharing: I was listening to Taylor's Q&A at PHPKonf 2020 (the video was uploaded at Dec 28, 2020), where [he referred to this PR -at 00:20:40-](https://youtu.be/Ympk3_WfKoc?t=1240) as potentially getting merged. The Socialstream package looks really great, can't wait to start using it.. Thank you @joelbutcher for the great work!

Issue #445, opened 2020-11-11
Inertia Stack doesn't respect fortify config


ManojKiranA:
- Jetstream Version: 1.6.0
- Jetstream Stack: Inertia
- Laravel Version: 8.14.0
- PHP Version: 7.4.11
- Database Driver & Version:  --

### Description:
Even after disabling the following features it doesn't hiding from profile 

```
Features::updateProfileInformation(),
Features::updatePasswords(),
```

### Steps To Reproduce:

Just try disabling the above features and refresh the profile page

driesvints:
Correct. I'll send in a PR to fix these. Thanks for reporting.

PR #451, opened 2020-11-12
[1.x] Fix feature toggles in Inertia

driesvints:
Fixes https://github.com/laravel/jetstream/issues/445

### DIFF
--- a/stubs/inertia/resources/js/Pages/Profile/Show.vue
+++ b/stubs/inertia/resources/js/Pages/Profile/Show.vue
@@ -8,19 +8,23 @@
 
         <div>
             <div class="max-w-7xl mx-auto py-10 sm:px-6 lg:px-8">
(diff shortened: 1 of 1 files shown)


ManojKiranA:
Hi @driesvints ,

I think currently we are not sharing those config to Inertia

[ShareInertiaData.php#L23](https://github.com/laravel/jetstream/blob/master/src/Http/Middleware/ShareInertiaData.php#L23)

Which may cause 

`Cannot read property 'canManageUpdateProfileInformation' of undefined"`

and also the keys which we don't share to frontend

driesvints:
That's odd because these worked for me. I did have to recompile my assets.

driesvints:
I've sent in a fix for this: https://github.com/laravel/jetstream/pull/454

PR #456, opened 2020-11-14
[1.x] Updating response status code to prevent method not allowed

parsingeye:
Return back with 303 when requesting with DELETE  and PUT to prevent 405 on the subsequent request.
`back()` returns 302 by default.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303
https://inertiajs.com/redirects

### DIFF
--- a/src/Http/Controllers/Inertia/ApiTokenController.php
+++ b/src/Http/Controllers/Inertia/ApiTokenController.php
@@ -62,7 +62,7 @@ class ApiTokenController extends Controller
             'abilities' => Jetstream::validPermissions($request->input('permissions', [])),
         ])->save();
 
-        return back();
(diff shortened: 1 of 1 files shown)


bonzai (approved):

taylorotwell:
Per Inertia documentation:

> If you're using one of our official server-side adapters, redirects will automatically be converted.

parsingeye:
That's true, now I see that redirects are converted in the code https://github.com/inertiajs/inertia-laravel/blob/master/src/Middleware.php#L121
For some reason, it still didn't work for me. I'm gonna dig a bit deeper into this.

taylorotwell:
I don't mind adding them here.

PR #461, opened 2020-11-15
Use local route methods

emargareten:
This PR replaces all usage of global (window) method `route()` to use the local method defined in mixin in app.js


taylorotwell:
There is no explanation as to why this is needed or what it fixes.

PR #467, opened 2020-11-16
[2.x] Add translation support to Inertia

### LINKED ISSUE #232: Add translation support for Inertia stack
At moment most of components in the *livewire* stub have translation support. This seems more easy because livewire components use Blade syntax `{{ __('...') }}`. But it is not the same with Inertia that uses Vue components.

I'd like a way to use the same translation files existing in `resources/lang`  (`.php` and `.json`) in Vue components provided by Inertia stack.

bonzai:
This pull request adds translation support to Inertia stack using the [vue-i18n](https://kazupon.github.io/vue-i18n/) package, which supports many features out-of-the-box: [formatting](https://kazupon.github.io/vue-i18n/guide/formatting.html), [pluralization](https://kazupon.github.io/vue-i18n/guide/pluralization.html), [datetime](https://kazupon.github.io/vue-i18n/guide/datetime.html) and [number](https://kazupon.github.io/vue-i18n/guide/number.html) localization.

I'm also using the `vue-i18n-loader` package to support [i18n custom blocks](https://kazupon.github.io/vue-i18n/guide/sfc.html) inside the single file components. It's possible to remove this dependency and only use [`Component based localization`](https://kazupon.github.io/vue-i18n/guide/component.html), but I decided not to.

Livewire already supports localization (see #101). Closes #232.

### Defining translations

Translations can be defined in many ways.

#### 1. Globally

We can define all translations for our app inside the `resources/js/app.js` file:

```js
const i18n = new VueI18n({
	// ...
    messages: {
        en: {
            hello: "hello world!"
        },
        ja: {
            hello: "こんにちは、世界！"
        },
    },
});
```

#### 2. Locally

We can use [i18n custom blocks](https://kazupon.github.io/vue-i18n/guide/sfc.html). This allows us to define translations inside the single file component:

```vue
<i18n src="./common/locales.json"></i18n>

<i18n>
{
    "en": {
        "hello": "hello world!"
    },
    "ja": {
        "hello": "こんにちは、世界！"
    }
}
</i18n>
```

Or we can use component based localization:

```vue
<script>
export default {
    i18n: {
        messages: {
            "en": {
                "hello": "hello world!"
            },
            "ja": {
                "hello": "こんにちは、世界！"
            }
        },
    }
}
</script>
```


drjdr:
Suggestion: Can we use .json files in a similar structure as laravel localization php files?
```
/resources
    /lang
        /en
            messages.json
        /es
            messages.json

```

bonzai:
You can use something like this if you need it, but keep in mind that `vue-i18n` has a different syntax (e.g. placeholders):

```js
// app.js
import en from 'path/to/resources/lang/en/messages.json'
import es from 'path/to/resources/lang/es/messages.json'

const i18n = new VueI18n({
    messages: {
        en,
        es,
    },
})
```

taylorotwell:
There was recently some fairly major shifting around of AppLayout.vue so this will need conflicts resolved. I'm generally fine with the idea though.

Enrica-r:
@bonzai Thank you for your extensive work and your documentation. Until now it's half way to have same functionality as the livewire (blade) stack has. To define messages via app.js or import it in app.js is ok for small applications or the "hello world" example. Same for <i18n> tag in vue.js files. But if you think about of a larger application or a web sites with several languages jetstream should not import all language definition files.

Even for one language could the import to be too large. On "vue-i18n" documentation is lazy load described. see [doc](https://kazupon.github.io/vue-i18n/guide/lazy-loading.html). It's done via a function "loadLanguageAsync" and a helper array "loadedLanguages". After every response app.js calls this function. If messages for this local have not loaded yet the it imports the language.json.

This load is better than hard coded message but disadvantages are:
- not compatible with Laravel translation definitions (/resources/lang/{lang}/*.php)
- loading a json file directly is no cached by the browser or if cached in .htacess then it's not versioned

I recommend to write an artisan command which scans the lang folder including json and php and which generate a js file per language containing one Javascript object or a json. Each php file generates a sub json. Eg. lang/en/auth.php --> {"auth":{"failed":"These...", "password":"The..."}}. An existing "en.json" is merged on top level. This js file (eg. en.js) is then importet by webpack as documented in "loadLanguageAsync".

With this further implementiation the Inertia stack has similar behaviour then normal Laravel.

I18n locale is set in your app.js via props from Laravel. For first load is that ok, but if locale changes on server the content of tag "<html lang="en"> will not be changed because of "app.blade.php" won't be reloaded. Therefore the entry app.js should check if locale has changed if yes then "loadLanguageAsync" (see above) and set lang={new} with (document.querySelector('html').setAttribute('lang', locale).

Thank you.
Regards Enrica

GENL:
I recommand to use  [Matice]( https://github.com/GENL/matice)

Issue #466, opened 2020-11-16
Team Switcher Menu


drjdr:
https://github.com/laravel/jetstream/blob/f053c5f9937bb42ced5dd069049ab48f9ade84ea/stubs/inertia/resources/js/Layouts/AppLayout.vue#L32

@taylorotwell currently the AppLayout.vue has `<div>` inside of `<button>`. I am making a PR to correct this. 
While at it, I was wondering what you think to have a separate `Team Switcher Menu`. (see screenshot below) 
+ The team name is always visible in the nav bar, which avoids a lot of confusion.
+ Menu items are more logically put in accordance to user vs team. (e.g. future Spark `Billing` vs `Team Billing`)
+ Only showing when hasTeamFeatures

<img width="1286" alt="Screenshot 2020-11-16 at 17 46 59" src="https://user-images.githubusercontent.com/55947988/99282438-f7287900-2833-11eb-9099-0c2a085b9970.png">

PR #468, opened 2020-11-16
#466 Team switcher

drjdr:
#466  implementation
- Fixes Button / div html5 validation
- please feel free to adapt to desired styling


stubs/inertia/resources/js/Layouts/AppLayout.vue
```diff
@@ -1,6 +1,6 @@
 <template>
     <div>
-        <jet-banner />
```
bonzai, line 3: I think you should revert any formatting changes, as they are not related to this PR.

Issue #309, opened 2020-10-05
TeamFactory when using the --teams option


tomb1n0:
First of all, sorry if this is in the wrong repo!

It would be nice to have a TeamFactory out of the box when using the `--teams` option. I appreciate this could easily be implemented by the developer, but this would provide a nice starting point for testing. You could immediately get started with writing tests around Team-based functionality. 

On the same note, it might be nice if the UserFactory returned a user that had a personal team by default, as this is what happens when a new user registers. Maybe have a `->withTeam()` method on the factory?

MCA332VSE:
I NEED TO OFFER A BOUNTY ON THE GROUP WHOS BEEN HACKING ME EITH 50%RECOVERED ASSETS AND LIQUID PROFITS FOR THERE INFORMATION OR ANY RETRIBUTION THEY OBTAIN

ZeProf2Code:
This issue can be closed: #469

PR #469, opened 2020-11-16
[2.x] TeamFactory

drjdr:
Implements #309 

UserFactory 
- also creates a personal team
- overwrites original laravel UserFactory
TeamFactory
- creates a personal_team as default

Only installs when --teams option is used.

### DIFF
--- a/database/factories/TeamFactory.php
+++ b/database/factories/TeamFactory.php
@@ -25,7 +25,7 @@ class TeamFactory extends Factory
         return [
             'name' => $this->faker->unique()->company,
             'user_id' => User::factory(),
-            'personal_team' => false,
+            'personal_team' => true,
         ];
     }
 }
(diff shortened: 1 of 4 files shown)


taylorotwell:
If you use this user factory to create a user... a team is created... does that team's user_id then create **another** user... such that creating 1 user with the factory creates an extra one every time?

drjdr:
**UserFactory.php** : 
`$user = User::factory()->create();` will use the configure() `has` below. The state specified here has a user_id specified which will overwrite the TeamFactory.php default ` 'user_id' => User::factory(), ' `.
This results in 1 user created + 1 personal team. 

```
    public function configure()
    {
        return $this->has(
            Team::factory()
                ->state(function (array $attributes, User $user) {
                    return ['name' => $user->name.'\'s Team', 'user_id' => $user->id, 'personal_team' => true];
                })
        );
    }
```

**TeamFactory.php** 
If no user_id is specified:
This will create a User + personal user team + a non-personal team for this new user.
If a user_id is specified:
team is attached to this user_id.

PR #481, opened 2020-11-20
[2.x] Move markdown files from root into resources/markdown

dbakan:
Currently, the markdown files `terms.md` and `policy.md` are published into the project's root directory.

This PR moves them into `resources/views/markdown` instead (also their stubs). Maybe somebody has an even better idea for their location, but personally I think the root directory is not ideal (esp. if you have many localized versions per file).

I also extracted the logic to lookup the localized versions (e.g. `terms.de.md`) into `Jetstream::localizedMarkdownPath($name)`. If this new function does not find its files in the new location, it checks the old location next.


taylorotwell:
What if the resources/views/markdown directory doesn't exist? Your Install command appears like it would fail? Did you check that?

driesvints:
These aren't views so I don't think it makes much sense to move them into the views directory.

dbakan:
@taylorotwell Good point, sorry, I missed that. Just added the checks. I tried to add tests, but couldn't figure out, how to do it best.

@driesvints I am open to any other suggestion of course. I just don't like them to lay around in the root directory.

driesvints:
@dbakan `resources/markdown` is where I'd put them.

dbakan:
Sounds good to me. I will change it. And also fix the wrong order of my `ensureDirectoryExists()`.

dbakan:
@driesvints Changed the paths to `resources/markdown`. BTW Is a good idea for me to force push such changes? Or would you prefer to keep the commit history complete until the PR gets merged/closed?

driesvints:
@dbakan it's fine.

Issue #482, opened 2020-11-20
Uncaught ReferenceError: route is not defined


TechhDan:
- Jetstream Version: 1.6.1
- Jetstream Stack: Inertia
- Laravel Version: 8.15.0
- PHP Version: 7.4.12
- Database Driver & Version: mysql v8.0.20

Description: Login page displays error on the web console. This message disappears once a user is logged in.
`Uncaught ReferenceError: route is not defined at Module../resources/js/app.js`

Steps To Reproduce:
1. laravel new sample
2. composer require laravel/jetstream
3. php artisan jetstream:install inertia (or with --teams)
4. npm install && npm run dev
5. php artisan migrate
6. php artisan serve
7. Visit Login, Registration, or Forgot Password page to view error on console.

![UncaughtReferenceErrorRoute](https://user-images.githubusercontent.com/11691784/99823456-69de7080-2b22-11eb-9e27-7a17c5b953a5.png)

driesvints:
What version of laravel-jetstream in your package.json are you using?

TechhDan:
I'm using laravel-jetstream version 0.0.3

andreliem:
Having same issue as well, same versions.

andreliem:
But I don't have the issue on another app with a older version

```
  "inertiajs/inertia-laravel": "^0.2.4",
  "laravel/jetstream": "^1.5",
```

```
    "laravel-jetstream": "^0.0.3",
```

philmmoore:
I had this issue myself when installing Jetstream with Inertia for the first time where there were a few hiccups along the way  i.e. the installer didn't work as expected because of my local environment so I had to install a few dependencies manually.

For me once I'd created an account and was taken to /dashboard the issue was clearer since I could only see @routes output in the browser. This is because the @routes blade directive wasn't available from the `tightenco/ziggy` and should've been added by the installer (but it wasn't because of those hiccups) https://github.com/laravel/jetstream/blob/daba98419164024e2c071f5547dd5569aba69d98/src/Console/InstallCommand.php#L249

If that's the same for you just do `composer require tightenco/ziggy` and it should resolve your issue.

rokde:
And the calls to `route('logout').url()` are wrong. There is no object you can call `.url()` at. Just remove and you are fine. It is in multiple Stub Templates in Jetstream wrong. So replace all occurrences to get it back to work.

TechhDan:
[...]

PR #485, opened 2020-11-24
[1.x] Use Alpine directly for guest layout

driesvints:
This reverts a small part of https://github.com/laravel/jetstream/pull/450. What happened was that because the guest layout isn't an SPA in Inertia the console starts to throw errors because it cannot resolve routes. This is technically solved in Jetstream v2 where the guest layout isn't copied anymore and all views are Inertia scripts. But for Jetstream 1.x we'll need to reference Alpine directly instead of loading in the entire `app.js` on the guest layout.

I have to note that nothing was actually broken, just the console which was showing the errors from the screenshots in the issue below.

Fixes https://github.com/laravel/jetstream/issues/482

### DIFF
--- a/stubs/resources/views/layouts/guest.blade.php
+++ b/stubs/resources/views/layouts/guest.blade.php
@@ -14,7 +14,7 @@
         <link rel="stylesheet" href="{{ mix('css/app.css') }}">
 
         <!-- Scripts -->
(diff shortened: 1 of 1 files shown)


driesvints:
Also, afaik the guest views don't use Alpine so not sure if it's needed at all?

PR #489, opened 2020-11-25
Publish "RemoveTeamMember" to allow it to be modified

joelbutcher:
This PR publishes the `RemoveTeamMember` action, allowing develops to tap into the functionality of what happens when you remove a team member.

### Example problem

_For the purposes of this PR, we consider an application where users can create either personal or team "reports". When a user creates a team report, they also own that report, only they can remove it_

1. User 1 (admin) invites User 2 (editor) to a team.
2. User 2 creates a team report.
3. User 1 removes user 2 from the team, but the report remains part of the team and owned by the user 2.

In this scenario, the remove team member functionality should allow developers to determine what happens to the report that user 2 is the owner of. (For example, programmatically make user 1 (the admin, and remover of user 2) the new owner of that report.


joelbutcher:
Fixing the tests now...

taylorotwell:
Send to `master` branch please.

taylorotwell:
Although I'm not sure this is needed? We fire a TeamMemberRemoved event that you can listen for to do whatever you want when a team member is removed. You receive the team instance and the member that was removed.

joelbutcher:
@taylorotwell Will send to master.

You don't receive the user to issued the transfer, which is quite essential to know IMO

PR #490, opened 2020-11-25
[2.x] Publish "RemoveTeamMember" to allow it to be modified

joelbutcher:
This PR publishes the `RemoveTeamMember` action, allowing develops to tap into the functionality of what happens when you remove a team member.

### Example problem

_For the purposes of this PR, we consider an application where users can create either personal or team "reports". When a user creates a team report, they also own that report, only they can remove it_

1. User 1 (admin) invites User 2 (editor) to a team.
2. User 2 creates a team report.
3. User 1 removes user 2 from the team, but the report remains part of the team and owned by the user 2.

In this scenario, the remove team member functionality should allow developers to determine what happens to the report that user 2 is the owner of. (For example, programmatically make user 1 (the admin, and remover of user 2) the new owner of that report.

Duplicate of #489 (closed as PR'd into wrong branch)


taylorotwell:
Why don't you just add the user that initiated the removal to the event so that we don't have to do all this publishing and adding contracts, etc.

joelbutcher:
@taylorotwell Because the event is fired as the last thing? 

You are publishing actions to do with inviting team members and a bunch of other team stuff (presumably to allow developers to customise them as they need? If not then I don't see the need for those either).

Would it not make sense to publish this action as well as it's essentially the inverse and developers may have all sorts of models (within a non-personal team context) associated to the the user being removed, that they would want to handle BEFORE the initial action to remove a user from a team is complete (i.e. when the event is fired). 

You could, for example, fire an event/notification to the user being removed that they need to action a transfer of the things they own to the admin (via a signed URL).

Issue #492, opened 2020-11-26
ConfirmPassword.vue calls url() on route()


vincenzoraco:
- Jetstream Version: 1.6.2
- Jetstream Stack: Inertia
- Laravel Version: 8.16.0
- PHP Version: 8.0.0-dev
- Database Driver & Version: Mysql 8

### Description:
The method `startConfirmingPassword` on the `Jetstream\ConfirmsPassword.vue` component is calling a `url()` method from the Ziggy `route()` function. This method is not available (or at least not documented) and it throws an error:

```
app.js:51694 [Vue warn]: Error in v-on handler: "TypeError: route(...).url is not a function"
```

### Steps To Reproduce:
Install Laravel with Jetstream
Create user account
Go to profile page
Click the `Enable` button from Two Factor Authentication section
It should prompt you to verify your password but it shows a console error.

vincenzoraco:
I've found that even `Layouts\AppLayout.vue` has a method (logout) that calls the url() method; created a pull request for that too.

bonzai:
I did a clean installation of Jetstream 1.6 (I followed your steps) and it's working fine for me:

![Zrzut ekranu 2020-11-26 o 02 15 41](https://user-images.githubusercontent.com/98191/100296732-4afe3500-2f8d-11eb-9801-96c672fc0e12.png)

Can you show your `composer.json`?

vincenzoraco:
That is weird, I called the same function on the terminal and threw the error above.

I also started a new clean installation

My composer.json is

```
"require": {
        "php": "^7.3|^8.0",
        "fideloper/proxy": "^4.4",
        "fruitcake/laravel-cors": "^2.0",
        "guzzlehttp/guzzle": "^7.0.1",
        "inertiajs/inertia-laravel": "^0.2.4",
        "laravel/framework": "^8.12",
        "laravel/jetstream": "^1.6",
        "laravel/sanctum": "^2.6",
        "laravel/tinker": "^2.5",
        "tightenco/ziggy": "^1.0"
    },
    "require-dev": {
        "facade/ignition": "^2.5",
        "fakerphp/faker": "^1.9.1",
        "mockery/mockery": "^1.4.2",
        "nunomaduro/collision": "^5.0",
        "phpunit/phpunit": "^9.3.3"
    },
```

bonzai:
> "tightenco/ziggy": "^1.0"

Did you change this line by yourself? Because Jetstream installs an older version (0.9.4):

[...]

PR #494, opened 2020-11-26
[1.x] Remove route().url() method call

vincenzoraco:
Fixes https://github.com/laravel/jetstream/issues/492


driesvints:
This is because you update Ziggy to v1 manually. Ziggy v1 will be supported in the upcoming Jetstream v2 release.

PR #504, opened 2020-12-01
[1.x] Rename path to prefix for route config

unlikenesses:
This changes the config key used to determine the prefix for Jetstream routes from `path` to `prefix`. This keeps it consistent with the same config key used in Fortify (see https://github.com/laravel/fortify/blob/1.x/src/FortifyServiceProvider.php#L128).


driesvints:
This is a breaking change on a stable branch. Best to be sent to master.

taylorotwell:
You can make it not breaking by supporting both - using one as a fallback to the other.

unlikenesses:
> 
> 
> You can make it not breaking by supporting both - using one as a fallback to the other.

I've made it fall back to `path`.

Issue #503, opened 2020-12-01
The logout feature does not work [BUG]


stephan0810:
- Jetstream Version: v1.6.2
- Jetstream Stack: Inertia
- Laravel Version: Laravel Framework 8.16.1
- PHP Version: 7.4.9
- Database Driver & Version: MySQL:5.7.26

### Description:

Previously, the logout feature worked fine using jetstream, but after upgrading the version, this behavior is no longer possible.
I haven't edited the code for jetstream logout, what kind of problem can I expect?

### Steps To Reproduce:

- install jetstream and choose inertiajs
- Register account and login
- Click the logout button in the header

bistory:
Check your version of Ziggy, it should be 0.9.4 for Jetstream 1.6.x

driesvints:
Hi there,

Thanks for reporting but it looks like this is a question which can be asked on a support channel. Please only use this issue tracker for reporting bugs with the library itself. If you have a question on how to use functionality provided by this repo you can try one of the following channels:

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://webchat.freenode.net/?nick=laravelnewbie&channels=%23laravel&prompt=1)

However, this issue will not be locked and everyone is still free to discuss solutions to your problem!

Thanks.

amineCHATATE:
With Jetstream we can use this code :

`<form method="POST" action="{{ route('logout') }}">
       @csrf
       <x-jet-dropdown-link href="{{ route('logout') }}" onclick="event.preventDefault(); this.closest('form').submit();">
                <i class="fa fa-sign-out"></i>{{ __('Logout') }}
        </x-jet-dropdown-link>
</form>`

amineCHATATE:
this this is used by jetstream by default

YisusYaro:
You need to use a form with method POST. Here an example.

<pre><code>
&lt;form method="POST" name="logout" action="{{ route('logout') }}">
      @csrf
      &lt;a class="dropdown-item" href="javascript:document.logout.submit()">Salir&lt;/a>
&lt;/form>
</code></pre>

geekfahim:
Here's the code for logout. I had added some conditions too. If anyone need feel free to use it.
```
        @if (auth()->id())
[...]

PR #505, opened 2020-12-03
Depends on Ziggy ^1.0.3 and fixed inertia stack still using .url()

bistory:
This PR fixes bug #503
Ziggy 1.0 was released and deprecated route(...).url() calls in favor of route(...).


driesvints:
This will be supported in v2

PR #506, opened 2020-12-03
[2.x] Enable assets versioning only in production

slavarazum:
Quote from Laravel Docs, which is true:

> Because versioned files are usually unnecessary in development, you may instruct the versioning process to only run during npm run production

Link - https://laravel.com/docs/8.x/mix#versioning-and-cache-busting

The same in laravel-mix docs - https://laravel-mix.com/docs/5.0/versioning#optionally-versioning-files

Related to - https://github.com/laravel/jetstream/pull/502

### DIFF
--- a/stubs/inertia/webpack.mix.js
+++ b/stubs/inertia/webpack.mix.js
@@ -17,5 +17,8 @@ mix.js('resources/js/app.js', 'public/js')
         require('tailwindcss'),
         require('autoprefixer'),
     ])
-    .webpackConfig(require('./webpack.config'))
-    .version();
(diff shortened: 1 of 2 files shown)


driesvints (approved):

nicotravassos:
Thanks @slavarazum but this should be applied to the livewire stack as well? 😃

taylorotwell:
Livewire stack too?

slavarazum:
It makes sense for both stacks even for any custom stack 😉.
Just have updated a PR.

nicotravassos:
Thanks @slavarazum.

PR #508, opened 2020-12-04
[1.x] Add vertical scroll to modal.blade component

marcoT89:
This enables the vertical scroll for long modals, still working for small ones and still centralized.
Also reduces a little bit the way of setting the `$maxWidth` variable. It's not a php8 pattern matching
but it's very similar.

### DIFF
--- a/resources/views/components/modal.blade.php
+++ b/resources/views/components/modal.blade.php
@@ -3,24 +3,13 @@
 @php
 $id = $id ?? md5($attributes->wire('model'));
 
-switch ($maxWidth ?? '2xl') {
-    case 'sm':
-        $maxWidth = 'sm:max-w-sm';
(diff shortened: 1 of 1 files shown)


driesvints:
@marcoT89 a quick gif/vid of the new behavior is always useful here. Good refactor to that lookup-map btw 👍

taylorotwell:
This PR doesn't update the Inertia version as well.

marcoT89:
I'm in a different timezone @taylorotwell that's why I couldn't respond so quickly. So be patient. I've opened a new [PR](https://github.com/laravel/jetstream/pull/512) with the changes to inertia as well. Please let me know if there's anything more to change and I'll do the changes asap. And thanks for the awesome package and framework.

PR #511, opened 2020-12-04
[2.x] Move Livewire views

driesvints:
This PR moves all views who are Livewire specific into the Livewire directory.


stubs/livewire/resources/views/layouts/guest.blade.php
```diff
@@ -14,7 +14,7 @@
         <link rel="stylesheet" href="{{ mix('css/app.css') }}">
 
         <!-- Scripts -->
-        <script src="https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.7.3/dist/alpine.js" defer></script>
+        <script src="{{ mix('js/app.js') }}" defer></script>
```
driesvints, line 17: I've re-added this now that Inertia doesn't uses this view anymore in v2.

PR #512, opened 2020-12-05
[1.x] Add vertical scroll to modal for blade and inertia

marcoT89:
This enables the vertical scroll for long modals, still working for small ones and still centralized.
Also reduces a little bit the way of setting the `$maxWidth` variable. It's not a php8 pattern matching
but it's very similar.

Before:
Screen Recording 2020-12-04 at 10 09 28 PM

After:
Screen Recording 2020-12-04 at 10 12 28 PM

### DIFF
--- a/resources/views/components/modal.blade.php
+++ b/resources/views/components/modal.blade.php
@@ -3,24 +3,13 @@
 @php
 $id = $id ?? md5($attributes->wire('model'));
 
-switch ($maxWidth ?? '2xl') {
-    case 'sm':
-        $maxWidth = 'sm:max-w-sm';
-        break;
-    case 'md':
-        $maxWidth = 'sm:max-w-md';
-        break;
-    case 'lg':
-        $maxWidth = 'sm:max-w-lg';
-        break;
-    case 'xl':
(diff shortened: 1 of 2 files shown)


resources/views/components/modal.blade.php
```diff
@@ -3,24 +3,13 @@
...
-        $maxWidth = 'sm:max-w-2xl';
-        break;
-}
+$maxWidth = [
+    'sm' => 'sm:max-w-sm',
+    'md' => 'sm:max-w-md',
+    'lg' => 'sm:max-w-lg',
+    'xl' => 'sm:max-w-xl',
+    '2xl' => 'sm:max-w-2xl',
+][$maxWidth ?? '2xl'];
```
bonzai, line 12: Please revert this change, because it's a breaking change.

If someone uses a name that is not defined, he will get an exception:

```blade
<x-jet-modal :maxWidth="XX" />

{{-- ErrorException --}}
{{-- Undefined array key "XX" --}}
```

nicotravassos:
Thank you @marcoT89

taylorotwell:
I've applied these changes in master branch.

PR #514, opened 2020-12-06
[2.x] Allows to use a custom TeamInvitation Model

RomainMazB:
Like all other models in Jetstream, TeamInvitation should be easily overridable.

A basic use case occurs when we don't want to send an email invitation but notify the user with another method (in-app notification for example).
In that case, we should be able to remove the email field and add another field to identify the user and send him the invitation the way we want using the already overridable `InvitesTeamMembers` contract.

*In the same time I fixed a (not-mine) typo in -probably- copy/pasted comment*


joelbutcher (approved):

joelbutcher:
@RomainMazB Big fan of publishing models for packages such as Jetstream. IMO the developer needs to have as much control of all the moving parts as possible, so the more flexible a framework/package is, the better!

Let's hope Taylor agrees 👍

PR #520, opened 2020-12-09
Fix avatar aspect ration in resposive navigation menu

raynirola:
Added `object-cover` in profile avatar `img` under  _Responsive Navigation Menu_


taylorotwell:
Fixed on master branch.

PR #548, opened 2020-12-18
[2.x] Dispatch team events

joelbutcher:
This PR dispatches the `TeamCreated` event after creating a team via the `TeamCreated` action. It also dispatches `TeamUpdated` event when a team's name is updated via the `UpdateTeamName` action.


bonzai:
I think the events are already dispatched:

https://github.com/laravel/jetstream/blob/master/stubs/app/Models/Team.php#L40-L42

joelbutcher:
@bonzai Indeed they are, my mistake, Thanks for pointing that out!

joelbutcher:
@bonzai random question (this part of events is lost on me). How does Laravel know how to resolve the correct `$team` model instance for those events?

PR #549, opened 2020-12-18
[2.x] Fix RegistrationTest.php

ZeProf2Code:
- Fixes RegistrationTest if terms are enabled.


joelbutcher:
@ZeProf2Code This test wasn't failing.

I would suggest adding a new test method (`test_new_users_can_register_with_terms`?) and ensuring that the terms feature is enabled in Jetstream before making the request.

driesvints:
I agree with @joelbutcher

PR #553, opened 2020-12-20
Update banner so that the navigation does not "jump" in height on slow connections

eporsche:
There is a short delay when the page loads and the banner gets hidden.
This has the effect that the navigation "jumps" in height.
X-cloak is used to hide the element until alpine is finished loading. Alpine will the automatically delete the x-cloak attribute from the element.

https://github.com/alpinejs/alpine#x-cloak


taylorotwell:
Add this to your own application if you want it. This would put x-cloak in both stack's CSS.

PR #559, opened 2020-12-21
Determine if the current stack is officially supported

TitasGailius:
## Problem

I'm creating a custom Jetstream stack.

Everything looks great so far except that Jetstream always tries to register livewire or inertia routes.

## Solution

This PR makes Jetstream register routes only when the current stack is officially supported.

### DIFF
--- a/src/Jetstream.php
+++ b/src/Jetstream.php
@@ -77,6 +77,23 @@ class Jetstream
     public static $inertiaManager;
 
     /**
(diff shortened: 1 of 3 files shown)


taylorotwell:
No plans to accommodate custom Jetstream stacks. Then I have to think about breaking changes against those stacks, etc. Build your own starter kit using Fortify as the auth backend.

PR #561, opened 2020-12-23
[2.x] Inertia: Remove moment.js dependency

claudiodekker:
While cleaning up the `app.js`, I noticed that the `moment` library was included, but was only being used once.

Because Inertia gets it's state from the back-end, we can instead let Carbon do this conversion for us and pass it through to the front-end and remove the (fairly large) `moment` dependency entirely.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -273,7 +273,6 @@ EOF;
                 '@tailwindcss/forms' => '^0.2.1',
                 '@tailwindcss/typography' => '^0.3.0',
                 'laravel-jetstream' => '^1.0.0',
-                'moment' => '^2.26.0',
                 'portal-vue' => '^2.1.7',
                 'postcss-import' => '^12.0.1',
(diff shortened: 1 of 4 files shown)


reinink:
Love this change, simply because Moment.js is *huge*. I personally removed it from my own SaaS app recently for the same reason. Read more about the issues with Momment [here](https://github.com/you-dont-need/You-Dont-Need-Momentjs).

taylorotwell:
One benefit of using Moment is the ability to use the user's local browser timezone when rendering / manipulating dates. Does this still display accurately?

reinink:
@taylorotwell Great point, I don't think this does. Let us investigate. 👍

claudiodekker:
@taylorotwell I don't think it'd matter in this case.. The server stores it's own timestamp in the database, and on retrieval it diffs against it's the timestamp it has at that point, so relatively the string would be identical, right?:

Screenshot 2020-12-23 at 14 59 57

reinink:
@claudiodekker Yup, you're right. Since the diff is done between two UTC dates, you'll end up with identical results either way. Good catch. 👍

PR #562, opened 2020-12-23
[1.x] Added authentication guard

thecaliskan:
Hi;

I use this system with Nova. Two guards are required in this scenario. If there is a possibility of define, it should be. No time should be wasted to find.

Needs to be standard in config files on Laravel Official Projects.

Thanks


taylorotwell:
Really should be `sanctum` as the default value.

thecaliskan:
Default value changed `sanctum`.

driesvints:
@thecaliskan this seems to break the test suite. Can you fix these breakages?

thecaliskan:
@driesvints  I am fixed. Default guard should be web.

taylorotwell:
No it shouldn't.

thecaliskan:
Why ?

PR #563, opened 2020-12-23
[2.x] Upgrade Inertia

claudiodekker:
This PR upgrades Jetstream to support and make use of the features shipped with `@inertiajs/inertia-vue` version `0.5.0`.

Specifically, it migrates away from [`jetstream-js`](https://github.com/laravel/jetstream-js) in favor of Inertia's own form helper, removes explicit 'error bag' definitions where possible, and also fixes a couple other minor issues found during this migration.

### DIFF
--- a/stubs/inertia/resources/js/Jetstream/ConfirmsPassword.vue
+++ b/stubs/inertia/resources/js/Jetstream/ConfirmsPassword.vue
@@ -23,7 +23,7 @@
             </template>
 
             <template #footer>
-                <jet-secondary-button @click.native="confirmingPassword = false">
+                <jet-secondary-button @click.native="closeModal">
                     Nevermind
                 </jet-secondary-button>
 
@@ -74,6 +74,10 @@
         },
 
         methods: {
+            closeModal() {
+                this.confirmingPassword = false
+                this.form.password = '';
+            },
             startConfirmingPassword() {
                 this.form.error = '';
 
@@ -82,11 +86,8 @@
                         this.$emit('confirmed');
                     } else {
                         this.confirmingPassword = true;
-                        this.form.password = '';
 
-                        setTimeout(() => {
-                            this.$refs.password.focus()
-                        }, 250)
+                        setTimeout(() => this.$refs.password.focus(), 250)
                     }
                 })
             },
--- a/stubs/inertia/resources/js/Jetstream/ConfirmsPassword.vue
+++ b/stubs/inertia/resources/js/Jetstream/ConfirmsPassword.vue
@@ -79,6 +79,7 @@
                 this.form.password = '';
                 this.form.error = '';
             },
+
             startConfirmingPassword() {
                 axios.get(route('password.confirmation')).then(response => {
                     if (response.data.confirmed) {
--- a/stubs/inertia/resources/js/Jetstream/ConfirmsPassword.vue
+++ b/stubs/inertia/resources/js/Jetstream/ConfirmsPassword.vue
@@ -4,7 +4,7 @@
             <slot />
         </span>
 
-        <jet-dialog-modal :show="confirmingPassword" @close="confirmingPassword = false">
+        <jet-dialog-modal :show="confirmingPassword" @close="closeModal">
             <template #title>
                 {{ title }}
             </template>
@@ -77,10 +77,9 @@
             closeModal() {
                 this.confirmingPassword = false
                 this.form.password = '';
+                this.form.error = '';
             },
             startConfirmingPassword() {
-                this.form.error = '';
-
                 axios.get(route('password.confirmation')).then(response => {
                     if (response.data.confirmed) {
                         this.$emit('confirmed');
@@ -97,12 +96,9 @@
 
                 axios.post(route('password.confirm'), {
                     password: this.form.password,
-                }).then(response => {
-                    this.confirmingPassword = false;
-                    this.form.password = '';
-                    this.form.error = '';
+                }).then(() => {
                     this.form.processing = false;
-
+                    this.closeModal()
                     this.$nextTick(() => this.$emit('confirmed'));
                 }).catch(error => {
                     this.form.processing = false;
--- a/stubs/inertia/resources/js/Jetstream/ConfirmsPassword.vue
+++ b/stubs/inertia/resources/js/Jetstream/ConfirmsPassword.vue
@@ -66,13 +66,10 @@
         data() {
             return {
                 confirmingPassword: false,
-
-                form: this.$inertia.form({
+                form: {
                     password: '',
                     error: '',
-                }, {
-                    bag: 'confirmPassword',
-                })
+                },
             }
         },
 
--- a/stubs/inertia/resources/js/Jetstream/ConfirmsPassword.vue
+++ b/stubs/inertia/resources/js/Jetstream/ConfirmsPassword.vue
@@ -103,6 +103,7 @@
                 }).catch(error => {
                     this.form.processing = false;
                     this.form.error = error.response.data.errors.password[0];
+                    this.$refs.password.focus()
                 });
             }
         }
--- a/stubs/inertia/resources/js/Pages/API/ApiTokenManager.vue
+++ b/stubs/inertia/resources/js/Pages/API/ApiTokenManager.vue
@@ -230,7 +230,6 @@
                     onSuccess: () => {
                         this.apiTokenRecentlyCreated = true
                         this.displayingToken = true
-
                         this.createApiTokenForm.reset()
                     }
                 })
--- a/stubs/inertia/resources/js/Pages/API/ApiTokenManager.vue
+++ b/stubs/inertia/resources/js/Pages/API/ApiTokenManager.vue
@@ -34,7 +34,7 @@
             </template>
 
             <template #actions>
-                <jet-action-message :on="apiTokenRecentlyCreated" class="mr-3">
+                <jet-action-message :on="createApiTokenForm.recentlySuccessful" class="mr-3">
                     Created.
                 </jet-action-message>
 
@@ -214,7 +214,6 @@
 
                 deleteApiTokenForm: this.$inertia.form(),
 
-                apiTokenRecentlyCreated: false,
                 displayingToken: false,
                 managingPermissionsFor: null,
                 apiTokenBeingDeleted: null,
@@ -225,9 +224,7 @@
             createApiToken() {
                 this.createApiTokenForm.post(route('api-tokens.store'), {
                     preserveScroll: true,
-                    before: () => (this.apiTokenRecentlyCreated = false),
                     onSuccess: () => {
-                        this.apiTokenRecentlyCreated = true
                         this.displayingToken = true
                         this.createApiTokenForm.reset()
                     }
--- a/stubs/inertia/resources/js/Pages/API/ApiTokenManager.vue
+++ b/stubs/inertia/resources/js/Pages/API/ApiTokenManager.vue
@@ -15,7 +15,7 @@
                 <div class="col-span-6 sm:col-span-4">
                     <jet-label for="name" value="Name" />
                     <jet-input id="name" type="text" class="mt-1 block w-full" v-model="createApiTokenForm.name" autofocus />
-                    <jet-input-error :message="createApiTokenForm.error('name')" class="mt-2" />
+                    <jet-input-error :message="createApiTokenForm.errors.name" class="mt-2" />
                 </div>
 
                 <!-- Token Permissions -->
@@ -34,7 +34,7 @@
             </template>
 
             <template #actions>
-                <jet-action-message :on="createApiTokenForm.recentlySuccessful" class="mr-3">
+                <jet-action-message :on="createApiTokenFormRecentlySuccessful" class="mr-3">
                     Created.
                 </jet-action-message>
 
@@ -68,12 +68,13 @@
 
                                 <div class="flex items-center">
                                     <div class="text-sm text-gray-400" v-if="token.last_used_ago">
-                                        Last used {{ fromNow(token.last_used_ago) }}
+                                        Last used {{ token.last_used_ago }}
                                     </div>
 
                                     <button class="cursor-pointer ml-6 text-sm text-gray-400 underline"
-                                                @click="manageApiTokenPermissions(token)"
-                                                v-if="availablePermissions.length > 0">
+                                        @click="manageApiTokenPermissions(token)"
+                                        v-if="availablePermissions.length > 0"
+                                    >
                                         Permissions
                                     </button>
 
@@ -205,20 +206,15 @@
                 createApiTokenForm: this.$inertia.form({
                     name: '',
                     permissions: this.defaultPermissions,
-                }, {
-                    bag: 'createApiToken',
-                    resetOnSuccess: true,
                 }),
 
                 updateApiTokenForm: this.$inertia.form({
                     permissions: []
-                }, {
-                    resetOnSuccess: false,
-                    bag: 'updateApiToken',
                 }),
 
                 deleteApiTokenForm: this.$inertia.form(),
 
+                createApiTokenFormRecentlySuccessful: false,
                 displayingToken: false,
                 managingPermissionsFor: null,
                 apiTokenBeingDeleted: null,
@@ -228,11 +224,14 @@
         methods: {
             createApiToken() {
                 this.createApiTokenForm.post(route('api-tokens.store'), {
+                    errorBag: 'createApiToken',
                     preserveScroll: true,
+                    before: () => (this.createApiTokenFormRecentlySuccessful = false),
                     onSuccess: () => {
-                        if (! this.createApiTokenForm.hasErrors()) {
-                            this.displayingToken = true
-                        }
+                        this.createApiTokenFormRecentlySuccessful = true
+                        this.displayingToken = true
+
+                        this.createApiTokenForm.reset()
                     }
                 })
             },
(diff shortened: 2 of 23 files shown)


stubs/inertia/resources/js/Pages/Teams/DeleteTeamForm.vue
```diff
@@ -81,7 +77,7 @@
 
             deleteTeam() {
                 this.form.delete(route('teams.destroy', this.team), {
-                    preserveScroll: true
+                    errorBag: 'deleteTeam'
```
claudiodekker, line 80: I removed `preserveScroll` here on purpose, because from my testing it actually didn't make sense (due to the redirect that happens to another page)

stubs/inertia/resources/js/Pages/Profile/TwoFactorAuthenticationForm.vue
```diff
@@ -152,7 +152,7 @@
             disableTwoFactorAuthentication() {
                 this.disabling = true
 
-                this.$inertia.delete('/user/two-factor-authentication', {
+                this.$inertia.delete('/user/two-factor-authentication', {}, {
```
claudiodekker, line 155: Inertia now takes an argument body for it's `delete` method, making the signature match the other methods. 
Since we're not posting any data here, I'm just passing in an empty object.

stubs/inertia/resources/js/Pages/API/ApiTokenManager.vue
```diff
@@ -34,7 +34,7 @@
             </template>
 
             <template #actions>
-                <jet-action-message :on="createApiTokenForm.recentlySuccessful" class="mr-3">
+                <jet-action-message :on="createApiTokenFormRecentlySuccessful" class="mr-3">
```
reinink, line 37: I wonder, should this local variable just be called `apiTokenRecentlyCreated`? Seems overly verbose. 🤷‍♂️ 
```suggestion
                <jet-action-message :on="apiTokenRecentlyCreated" class="mr-3">
```

stubs/inertia/resources/js/Pages/API/ApiTokenManager.vue
```diff
@@ -228,11 +224,14 @@
...
+                    errorBag: 'createApiToken',
                     preserveScroll: true,
+                    before: () => (this.createApiTokenFormRecentlySuccessful = false),
                     onSuccess: () => {
-                        if (! this.createApiTokenForm.hasErrors()) {
-                            this.displayingToken = true
-                        }
+                        this.createApiTokenFormRecentlySuccessful = true
+                        this.displayingToken = true
+
```
reinink, line 233: Personal preference...don't think you *need* the space here.
```suggestion
```

stubs/inertia/resources/js/Pages/Auth/Login.vue
```diff
@@ -68,28 +68,20 @@
...
             submit() {
-                this.form.post(this.route('login'), {
-                    onSuccess: () => {
-                        this.remember = false
-                    }
-                })
+                this.form.post(this.route('login'))
             }
         }
     }
```
reinink, line 87: Oh my, the simplification here! 😍

stubs/inertia/resources/js/Pages/Auth/Login.vue
```diff
@@ -68,28 +68,20 @@
...
-        },
-
-        watch: {
-            remember(value) {
-                this.form.remember = value ? 'on' : ''
+                    remember: false
+                }).transform(data => ({
+                    ... data,
+                    remember: this.form.remember ? 'on' : ''
+                }))
```
reinink, line 78: Personally I think it's better to put the `transform` in the `submit` handler. Just because it has more to do with sending the data to the server.

stubs/inertia/resources/js/Pages/Auth/Register.vue
```diff
@@ -79,18 +79,15 @@
...
         },
 
         methods: {
             submit() {
                 this.form.post(this.route('register'), {
-                    onSuccess: () => {
-                        this.form.password = ''
-                        this.form.password_confirmation = ''
-                    }
+                    onFinish: () => this.form.reset('password', 'password_confirmation'),
```
reinink, line 90: 😍

reinink, line 90: Curious, why do we need to reset these fields at all? Doesn't the user get redirected after they register? 🤔

stubs/inertia/resources/js/Pages/Auth/ResetPassword.vue
```diff
@@ -68,10 +68,7 @@
         methods: {
             submit() {
                 this.form.post(this.route('password.update'), {
-                    onSuccess: () => {
-                        this.form.password = ''
-                        this.form.password_confirmation = ''
-                    }
+                    onFinish: () => this.form.reset('password', 'password_confirmation'),
```
reinink, line 71: Shouldn't this only happen on success?

stubs/inertia/resources/js/Pages/Auth/VerifyEmail.vue
```diff
@@ -18,7 +18,7 @@
                     Resend Verification Email
                 </jet-button>
 
-                <inertia-link :href="route('logout')" method="post" class="underline text-sm text-gray-600 hover:text-gray-900">Logout</inertia-link>
+                <inertia-link :href="route('logout')" method="post" as="button" class="underline text-sm text-gray-600 hover:text-gray-900">Logout</inertia-link>
```
reinink, line 21: I haven't had a chance to install and test this yet, but can you confirm that this hasn't messed up the style of the Logout button? In Ping CRM I had to add `w-full text-left` to get it render properly, just because of default button browser styles.

stubs/inertia/resources/js/Pages/Profile/LogoutOtherBrowserSessionsForm.vue
```diff
@@ -48,7 +48,7 @@
                     Logout Other Browser Sessions
                 </jet-button>
 
-                <jet-action-message :on="form.recentlySuccessful" class="ml-3">
+                <jet-action-message :on="formRecentlySuccessful" class="ml-3">
```
reinink, line 51: Again, I'd be inclined to use a more meaningful name here, such as `successfullyLogoutOtherBrowserSessions`.
```suggestion
                <jet-action-message :on="successfullyLogoutOtherBrowserSessions" class="ml-3">
```

stubs/inertia/resources/js/Pages/Profile/UpdatePasswordForm.vue
```diff
@@ -12,24 +12,24 @@
...
                 <jet-label for="password_confirmation" value="Confirm Password" />
                 <jet-input id="password_confirmation" type="password" class="mt-1 block w-full" v-model="form.password_confirmation" autocomplete="new-password" />
-                <jet-input-error :message="form.error('password_confirmation')" class="mt-2" />
+                <jet-input-error :message="form.errors.password_confirmation" class="mt-2" />
             </div>
         </template>
 
         <template #actions>
-            <jet-action-message :on="form.recentlySuccessful" class="mr-3">
+            <jet-action-message :on="formRecentlySuccessful" class="mr-3">
```
reinink, line 32: Another one of these possible renames. 😎 
```suggestion
            <jet-action-message :on="successfullyUpdatedPassword" class="mr-3">
```

stubs/inertia/resources/js/Pages/Teams/TeamMemberManager.vue
```diff
@@ -59,7 +59,7 @@
                 </template>
 
                 <template #actions>
-                    <jet-action-message :on="addTeamMemberForm.recentlySuccessful" class="mr-3">
+                    <jet-action-message :on="addTeamMemberFormRecentlySuccessful" class="mr-3">
```
reinink, line 62: Another possible naming change.
```suggestion
                    <jet-action-message :on="successfullyAddedTeamMember" class="mr-3">
```

stubs/inertia/resources/js/Pages/Teams/TeamMemberManager.vue
```diff
@@ -351,7 +342,9 @@
             },
 
             leaveTeam() {
-                this.leaveTeamForm.delete(route('team-members.destroy', [this.team, this.$page.props.user]))
+                this.leaveTeamForm.delete(route('team-members.destroy', [this.team, this.$page.props.user]), {
```
reinink, line 345: Missing second argument?
```suggestion
                this.leaveTeamForm.delete(route('team-members.destroy', [this.team, this.$page.props.user]), {}, {
```

reinink:
This looks awesome @claudiodekker! The only thing I'm left wondering is whether any of the `errorBag` stuff is even required anymore, given that the errors are all automatically scoped to the individual forms? But maybe we do, simply because they are defined server-side as well, and I'm not sure we can remove them from there, as that code might be shared with Livewire.

One think I'm pretty sure we can remove is the shared `errorBags` data, as this is all handled through the `errors` prop now:

https://github.com/laravel/jetstream/blob/bbaa2cb2b355a0c3aa57f5e510dc41bf4b9ef301/src/Http/Middleware/ShareInertiaData.php#L53-L57

stubs/inertia/resources/js/Pages/Profile/TwoFactorAuthenticationForm.vue
joelbutcher: @claudiodekker  should this route be updated to use the named route provided by Ziggy?

claudiodekker: @joelbutcher I looked into this, but because this route is coming from Fortify and isn't named, this isn't possible. Not a bad idea, though!

stubs/inertia/resources/js/Pages/Auth/Register.vue
claudiodekker: Primarily the reasoning here is that if there's errors on the page, you do want to reset the passwords fields.
Furthermore, it's never a bad idea to clear this kind of sensitive data when you can, as you don't want to accidentally retain them in history/storage.

stubs/inertia/resources/js/Pages/Auth/ResetPassword.vue
claudiodekker: That'd change the behaviour, as with the 'old' version of Inertia, `onSuccess` was also triggered on validation errors etc.. This way, the behaviour stays the same, all while retaining the benefits I mentioned in my earlier comment

stubs/inertia/resources/js/Pages/Auth/VerifyEmail.vue
claudiodekker: Yep, looks fine 👍

stubs/inertia/resources/js/Pages/Teams/TeamMemberManager.vue
claudiodekker: This is part of the Inertia form, which only takes options as the 2nd argument ;)

reinink: IGNORE ME CLAUDIO 🤦

stubs/inertia/resources/js/Pages/Profile/TwoFactorAuthenticationForm.vue
joelbutcher: @claudiodekker ah okay, it might be worth a quick PR into Fortify to change this

taylorotwell:
Nice! ❤️ 

One thing I'm curious about is the removal of `recentlySuccessful`. That was pretty convenient compared to adding a separate "success" boolean variable for each form like in this PR. Is there a specific reason that wasn't ported over?

claudiodekker:
Haha, thanks! 

Yeah, @reinink  and I had discussed it, and almost added it yesterday, but basically we found it difficult to really put a finger on when a form is considered recently submitted, and at what point it's modified too much (e.g. resetting fields or clearing errors) and should no longer be considered as such.

I think I can also speak for the both of us when I say that we do see the use cases & added value of something like this, but because we really wanted to get this release out of the door and didn't want to last-minute ship something we'd end up regretting, we decided to simply shelve the idea for now and to implement it with local state instead.

Merry Christmas! 🎄

bonzai:
There are still files that use `form.recentlySuccessful`:
- [UpdateProfileInformationForm.vue#L60](https://github.com/claudiodekker/jetstream/blob/upgrade-inertia/stubs/inertia/resources/js/Pages/Profile/UpdateProfileInformationForm.vue#L60)
- [UpdateTeamNameForm.vue#l41](https://github.com/claudiodekker/jetstream/blob/upgrade-inertia/stubs/inertia/resources/js/Pages/Teams/UpdateTeamNameForm.vue#L41)

claudiodekker:
Thanks @bonzai! 👍 

@taylorotwell Only after @bonzai's comment just now I noticed that `recentlySuccessful` in `jetstream-js` automatically resets after 2 seconds, which is something that I hadn't noticed before, and I don't think @reinink had either. I'll discuss this with him tomorrow, as this means that the implementation suddenly becomes very simple and makes a lot of sense.

taylorotwell:
@claudiodekker Thanks. Yeah, it was nice for transitions and such on those Saved indicators that fade out in Jetstream.

claudiodekker:
@taylorotwell Done! We've added `recentlySuccessful` as of https://github.com/inertiajs/inertia/pull/374 and have implemented those changes here.

[At the same time](https://github.com/laravel/jetstream/pull/563/commits/7057ea3785fcc87b9799a1f89c382371846f032e)  I made a few changes to some sensitive forms, to directly clear passwords from the page entirely on the close of a modal, instead of only when that model gets re-opened. 👍

bonzai:
[ConfirmsPassword.vue#L7](https://github.com/laravel/jetstream/blob/7057ea3785fcc87b9799a1f89c382371846f032e/stubs/inertia/resources/js/Jetstream/ConfirmsPassword.vue#L7):

```diff
-<jet-dialog-modal :show="confirmingPassword" @close="confirmingPassword = false">
+<jet-dialog-modal :show="confirmingPassword" @close="closeModal">
```

[DeleteUserForm.vue#L23](https://github.com/laravel/jetstream/blob/7057ea3785fcc87b9799a1f89c382371846f032e/stubs/inertia/resources/js/Pages/Profile/DeleteUserForm.vue#L23):

```diff
-<jet-dialog-modal :show="confirmingUserDeletion" @close="confirmingUserDeletion = false">
+<jet-dialog-modal :show="confirmingUserDeletion" @close="closeModal">
```

[LogoutOtherBrowserSessionsForm.vue#L57](https://github.com/laravel/jetstream/blob/7057ea3785fcc87b9799a1f89c382371846f032e/stubs/inertia/resources/js/Pages/Profile/LogoutOtherBrowserSessionsForm.vue#L57):

```diff
-<jet-dialog-modal :show="confirmingLogout" @close="confirmingLogout = false">
+<jet-dialog-modal :show="confirmingLogout" @close="closeModal">
```

stubs/inertia/resources/js/Jetstream/ConfirmsPassword.vue
```diff
@@ -66,17 +66,18 @@
...
+                },
             }
         },
 
         methods: {
+            closeModal() {
+                this.confirmingPassword = false
+                this.form.password = '';
+            },
             startConfirmingPassword() {
```
bonzai, line 81:
```suggestion
            },

            startConfirmingPassword() {
```

stubs/inertia/resources/js/Pages/Profile/DeleteUserForm.vue
```diff
@@ -75,33 +75,27 @@
...
                 })
             }
         },
 
         methods: {
+            closeModal() {
+                this.confirmingUserDeletion = false
+                this.form.reset()
+            },
             confirmUserDeletion() {
```
bonzai, line 88:
```suggestion
            },

            confirmUserDeletion() {
```

stubs/inertia/resources/js/Pages/Profile/LogoutOtherBrowserSessionsForm.vue
```diff
@@ -113,33 +113,27 @@
...
                 })
             }
         },
 
         methods: {
+            closeModal() {
+                this.confirmingLogout = false
+                this.form.reset()
+            },
             confirmLogout() {
```
bonzai, line 126:
```suggestion
            },

            confirmLogout() {
```

claudiodekker:
Thanks a lot for the review(s), @bonzai! 💪 👍

markvesterskov:
@claudiodekker I'm checking this pull request out, its awesome! 

Shouldnt the Fortify actions be updated to not use error bags as well? eg. `https://github.com/laravel/jetstream/blob/master/stubs/app/Actions/Fortify/UpdateUserProfileInformation.php`

claudiodekker:
@markvesterskov I'd like that, but you have to remember that Jetstream is just one implementation of Fortify. If we were to remove the error bags from there, it'd break any other possible implementations that exist and rely on them being set. 

This is also why some Inertia forms/requests in this PR still use a manually defined error bag, even after this PR.

markvesterskov:
> @markvesterskov I'd like that, but you have to remember that Jetstream is just one implementation of Fortify. If we were to remove the error bags from there, it'd break any other Jetstream-like packages that may exist and rely on them being set.
> 
> This is also why some Inertia forms/requests in this PR still use a manually defined error bag, even after this PR.

@claudiodekker Makes great sense, thanks for the reply, and thanks for the awesomeness with Jetstream and Inertia - I'm working on a new project using the master branch and it is a pleasure to work with.

PR #574, opened 2021-01-01
Add @can check to mobile menu

stancl:
Adds `@can` check to the mobile menu.

This check is present on the desktop menu:

https://github.com/stancl/jetstream/blob/5c09ad50e4371610ebbc336aa0b6f9a3d5c8eccf/stubs/livewire/resources/views/navigation-dropdown.blade.php#L71-L75

The current behavior on mobile is that user is shown a link in the menu, he can visit the "create team" site, and only when he submits the form he gets a 403.


taylorotwell:
Please send to master branch.

PR #581, opened 2021-01-04
[2.x] Test update for Registration,Two Factor Authentication and Account De…

FaizAhmadSE:
…letion features


driesvints:
You own these files. Just remove the tests which don't apply to your app.

PR #587, opened 2021-01-05
Automatically select API token when it's created

stancl:
The user can now just CTRL+C.

Screen Shot 2021-01-05 at 21 40 33

Let me know if the template needs any changes.


stancl:
(Found that it needs one more change, so I'm flagging this as a draft, will push a commit shortly)

stancl:
The `@keydown.prevent` was preventing ctrl+c as well. Changed it to `readonly` so it's not disabled & can be selected, but keyboard shortcuts still work 👍🏻

taylorotwell:
Please use 2.x branch.

PR #588, opened 2021-01-05
Add --local-composer flag to the install command

miclaus:
Using this flag will use the local composer.phar binary installed at the base path instead of the global composer binary command.

This is useful for systems where composer is not installed globally.

Flag Usage:
```
php artisan jetstream:install livewire --local-composer

php artisan jetstream:install livewire --teams --local-composer
```

Executes e.g. `php composer.phar require livewire/livewire:^2.0 laravel/sanctum:^2.6` 
instead of `composer require livewire/livewire:^2.0 laravel/sanctum:^2.6`

Not sure about the command flag description though. 
Happy for suggestions.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -16,7 +16,8 @@ class InstallCommand extends Command
      * @var string
      */
     protected $signature = 'jetstream:install {stack : The development stack that should be installed}
-                                              {--teams : Indicates if team support should be installed}';
+                                              {--teams : Indicates if team support should be installed}
+                                              {--local-composer : Indicates that the local composer.phar should be use instead of the gobal composer}';
 
     /**
      * The console command description.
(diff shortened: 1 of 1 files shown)


taylorotwell:

miclaus:
@taylorotwell The proposed flag is not necessary to be merged for the jetstream:install command.

However, developers using a local composer.phar binary instead of a globally installed composer binary cannot currently run the jetsteam:install command successfully, without a globally installed composer binary.

It would be best if the install command detects automatically whether to use the globally installed composer binary or (if available) a local composer.phar binary to require packages.

Thanks for your time.

driesvints:
@miclaus when I said we could consider this I meant that an argument could be added to inculde the full pathname to the composer binary, not just a flag that assumes a phar. If you want to retry this please try with that in mind.

PR #593, opened 2021-01-06
[2.x] Fix registration test

fdalcin:
Currently when Terms and Privacy feature is enabled the `test_new_users_can_register` fails because Jetstream will require the accepted `terms`.

The field could be passed either way, but I opted for verifying if the feature is enabled to be more explicit.


ZeProf2Code:
See #549

fdalcin:
> See #549

Oh, my bad. Thanks.

PR #599, opened 2021-01-06
[2.x] Banner should be hidden in default

RiKap:
There is a small design bug.
When you open you website (or refresh) it jumps a little bit as you can see on this gif.
2021-01-06 at 17

This PR will fix that problem.

### DIFF
--- a/resources/views/components/banner.blade.php
+++ b/resources/views/components/banner.blade.php
@@ -9,7 +9,8 @@
                     message = event.detail.message;
                     show = true;
                 });
-            ">
+            "
+            style="display: none;">
     <div class="max-w-screen-xl mx-auto py-2 px-3 sm:px-6 lg:px-8">
         <div class="flex items-center justify-between flex-wrap">
             <div class="w-0 flex-1 flex items-center min-w-0">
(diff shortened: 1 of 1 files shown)


resources/views/components/banner.blade.php
```diff
@@ -9,7 +9,8 @@
                     message = event.detail.message;
                     show = true;
                 });
-            ">
+            "
+            style="display: none;">
```
driesvints, line 13: Please use a tailwind class here.

RiKap: It is not possible to use `hidden` class without another changes as it will be hidden forever.

To make `class="hidden"` work, I will need to update `:class` attribute and add `, hidden: !show || !message`. Otherwise apline.js does not know, that he should remove `hidden` class.

If it is ok, I will update this PR, just let me know.

joelbutcher: @RiKap in app.cas add the following:

```css
[x-cloak] {
    display: none;
}
```

You can then use this as an HTML attribute! 😎

driesvints: @RiKap ah okay, I was confusing this with Vue 3's behavior for a sec 😅 Just leave this be for now.

PR #604, opened 2021-01-06
[2.x] Updated UPGRADE.md

daviddarke:
Include missing `InviteTeamMember` action reference in the upgrade doc. This step was missing when I just upgraded.

Even though it was easy to figure what was going on, I'm sure this is a required step and might catch people out.


taylorotwell:
That won't work because views would have to be updated, etc.

The upgrade guide is **only** meant to inform people how to keep using people's existing Jetstream 1.x feature set on a 2.x backend.

PR #607, opened 2021-01-07
[2.x] Add --composer option to the install command

miclaus:
Adds the optional `--composer` option to the install command. This option may receive a path pointing to a specific composer binary. As proposed in PR #588.

```
php artisan jetstream:install livewire --composer=composer.phar

php artisan jetstream:install livewire --teams --composer=composer.phar
```

This is useful for systems where no global composer binary is installed.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -16,7 +16,8 @@ class InstallCommand extends Command
      * @var string
      */
     protected $signature = 'jetstream:install {stack : The development stack that should be installed}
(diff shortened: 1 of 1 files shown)


taylorotwell:
Why do you not have Composer globally available?

miclaus:
@taylorotwell personally, I used to run into permission issues where I was forced to use `sudo` (which I later learned should be avoided, https://getcomposer.org/root)

But I also found it practical to have it included for developers who don't necessarily use Composer.

PR #595, opened 2021-01-06
[2.x] Improve registration test

joelbutcher:
Previous PR's (#549 #593) have been attempted to fix this test as it currently fails for developers when their application's `jetstream.features` config contains the`terms` feature.

This PR improves and updates the current test by ensuring the terms feature is not in the `jetstream.features` config.

An additional test has been added to test registration with terms enabled.

PR #609, opened 2021-01-08
Fix RegistrationTest, account for case with terms feature on

drehimself:
If you turn on the `Features::termsAndPrivacyPolicy()` feature in Jetstream, the `RegistrationTest` doesn't account for this case and will fail. Fixed by passing through the `terms` field if the feature is on.

### DIFF
--- a/stubs/tests/RegistrationTest.php
+++ b/stubs/tests/RegistrationTest.php
@@ -4,6 +4,7 @@ namespace Tests\Feature;
 
 use App\Providers\RouteServiceProvider;
 use Illuminate\Foundation\Testing\RefreshDatabase;
+use Laravel\Jetstream\Jetstream;
 use Tests\TestCase;
 
 class RegistrationTest extends TestCase
@@ -24,6 +25,7 @@ class RegistrationTest extends TestCase
             'email' => 'test@example.com',
             'password' => 'password',
             'password_confirmation' => 'password',
+            'terms' => Jetstream::hasTermsAndPrivacyPolicyFeature() ? 'on' : null,
         ]);
 
         $this->assertAuthenticated();


ZeProf2Code:
Ses #595

driesvints:
Hey @drehimself. The test suite is meant to mimic an app in its current working state. It doesn't makes sense to adopt an configurable state in the test suite since you're app will always either have this enabled or disabled. You own this file so feel free to adjust it if you turn on the terms feature.

Illizian:
> Hey @drehimself. The test suite is meant to mimic an app in its current working state. It doesn't makes sense to adopt an configurable state in the test suite since you're app will always either have this enabled or disabled. You own this file so feel free to adjust it if you turn on the terms feature.

Could you clarify what you mean by this? There already exists a range of tests that are toggled by Feature Flags! (See. `test_api_token_permissions_can_be_updated`, `test_email_verification_screen_can_be_rendered`), I'm glad I spotted this issue before adding my own PR but it'd be good to know so as to avoid wasted effort in the future.

driesvints:
@Illizian we started accepting the PRs with feature flags recently because we got tired of having to explain over and over and closing PRs.

Illizian:
Cool, I'll pop a PR in shortly then 😏😋

driesvints:
These still don't make any sense. Your app will either be in one state or the other. Never in an ability to toggle between states. Just delete/add what you need or don't need.

Illizian:
Yeah, fair point. It's a weird concept that the tests are provided IMO.

I'm still not sure if jetstream is a package or a boilerplate - if that
makes sense? It kinda skirts both realms.

I came here off the back of existing skipped tests (FWIW).

driesvints:
@Illizian it's a scaffold tool that mostly scaffolds your app and provides a few internal classes to handle those parts out of the box. But all of the UI is modifiable after it's scaffolded.

Illizian:
Ok, thanks for the clarification! I rest my sarcasm haha!

I didn't realize the existing skips were added under duress and it makes
sense if you position jetstream as a boilerplate.

There's still some vendor publishes etc that pull it back into package
territory though.

driesvints:
Yeah, we can't really stop that from happening I guess :-/

PR #618, opened 2021-01-08
[2.x] Change current team of user when creating a new additional team

OscarKolsrud:
Current behaviour is when a new team is created you are redirected back to the dashboard of the team that you were in before creating the new. In almost all cases a users next action would be directed against the new team.

This PR solves that problem by modifying the CreateTeam class and updating the current_team_id param on the user model of the user creating the team. This way when the user is redirected back they are acting in the context of the newly created team.

### DIFF
--- a/stubs/app/Actions/Jetstream/CreateTeam.php
+++ b/stubs/app/Actions/Jetstream/CreateTeam.php
@@ -32,9 +32,7 @@ class CreateTeam implements CreatesTeams
             'personal_team' => false,
         ]);
 
-        $user->forceFill([
-            'current_team_id' => $team->id,
-        ])->save();
+        $user->switchTeam($team);
 
         return $team;
     }
--- a/stubs/app/Actions/Jetstream/CreateTeam.php
+++ b/stubs/app/Actions/Jetstream/CreateTeam.php
@@ -33,7 +33,7 @@ class CreateTeam implements CreatesTeams
         ]);
 
         $user->forceFill([
-            'current_team_id' => $team->id
+            'current_team_id' => $team->id,
         ])->save();
 
         return $team;
--- a/stubs/app/Actions/Jetstream/CreateTeam.php
+++ b/stubs/app/Actions/Jetstream/CreateTeam.php
@@ -32,10 +32,9 @@ class CreateTeam implements CreatesTeams
             'personal_team' => false,
         ]);
 
-        $user->forceFill({
+        $user->forceFill([
             'current_team_id' => $team->id
-        })->save();
-        $user->save();
+        ])->save();
 
         return $team;
     }
--- a/stubs/app/Actions/Jetstream/CreateTeam.php
+++ b/stubs/app/Actions/Jetstream/CreateTeam.php
@@ -27,9 +27,14 @@ class CreateTeam implements CreatesTeams
 
         AddingTeam::dispatch($user);
 
-        return $user->ownedTeams()->create([
+        $team = $user->ownedTeams()->create([
             'name' => $input['name'],
             'personal_team' => false,
         ]);
+
+        $user->current_team_id = $team->id;
+        $user->save();
+
+        return $team;
     }
 }
--- a/stubs/app/Actions/Jetstream/CreateTeam.php
+++ b/stubs/app/Actions/Jetstream/CreateTeam.php
@@ -32,7 +32,9 @@ class CreateTeam implements CreatesTeams
             'personal_team' => false,
         ]);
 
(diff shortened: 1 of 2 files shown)


OscarKolsrud:
Hmm, the tests are logically enough failing because its trying to assert that its the same team even though its been changed in the create function. Any suggestions on how to best solve this? Changing the test?

joelbutcher (changes requested):

stubs/app/Actions/Jetstream/CreateTeam.php
```diff
@@ -27,9 +27,14 @@ public function create($user, array $input)
 
         AddingTeam::dispatch($user);
 
-        return $user->ownedTeams()->create([
+        $team = $user->ownedTeams()->create([
             'name' => $input['name'],
             'personal_team' => false,
         ]);
+
+        $user->current_team_id = $team->id;
```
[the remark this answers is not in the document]
OscarKolsrud: Great thanks!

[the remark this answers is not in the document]
joelbutcher: @OscarKolsrud my apologies, was typing on my phone, the curly braces should be square brackets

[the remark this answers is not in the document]
OscarKolsrud: Yep, just noticed! Fixed in a seperate commit :) In my opinion line 150 from TeamBehaviourTest can be removed since there really is no need for it. Just tested this fix on a inertia stack and everything seems to update like it should in the UI

OscarKolsrud:
Okay, I have gone ahead and manually tested both the Inertia stack and the Livewire stack that these changes do not break how the f.ex team switcher updates. I think that line 150 of the TeamBehaviorTest is unneeded and can be removed to allow for this change to proceed.

tests/TeamBehaviorTest.php
```diff
@@ -147,8 +147,6 @@ public function test_user_does_not_need_to_refresh_after_switching_teams()
 
         $anotherTeam = $action->create($user, ['name' => 'Test Team']);
 
-        $this->assertTrue($user->isCurrentTeam($personalTeam));
```
joelbutcher, line 150: Why not just move this line to before the previous line is executed?

tests/TeamBehaviorTest.php
```diff
@@ -147,8 +147,6 @@ public function test_user_does_not_need_to_refresh_after_switching_teams()
...
-
         $user->switchTeam($anotherTeam);
```
joelbutcher, line 152: This line can be removed now.

```suggestion

```

stubs/app/Actions/Jetstream/CreateTeam.php
```diff
@@ -27,9 +27,15 @@ public function create($user, array $input)
 
         AddingTeam::dispatch($user);
 
-        return $user->ownedTeams()->create([
+        $team = $user->ownedTeams()->create([
             'name' => $input['name'],
             'personal_team' => false,
         ]);
+
+        $user->forceFill([
```
joelbutcher, line 35: Rethinking this for the failed tests:

```suggestion
        $user->switchTeam($team);
```

OscarKolsrud: Trying now!

OscarKolsrud: @joelbutcher That seems to solve the failing tests! But lastly i cant wrap my head around why StyleCI is bugging us. That file looks perfectly fine to me. Do you see any thing?

joelbutcher: @OscarKolsrud it's the white space after 

```php
         $personalTeam->forceFill(['personal_team' => true])->save();

```

OscarKolsrud: Fixed! All tests passing! I think this is now ready to merge :)

dillingham:
Maybe out of the scope of this PR, 

I was thinking about doing this + redirecting to the team edit screen in my project.

Naturally would want to add members after creating a team.

OscarKolsrud:
> Maybe out of the scope of this PR,
> 
> I was thinking about doing this + redirecting to the team edit screen in my project.
> 
> Naturally would want to add members after creating a team.

Cool idea! Maybe i will make a PR later that adds a config option of where to redirect after team creation, dont think its smart to make this PR very big! Could see tons of situations where that would be useful. For now maybe add the change yourself (On the livewire controller I imagine), or maybe you could possibly even make use of a middleware?

PR #627, opened 2021-01-11
[2.x] Make text translatable in confirm-password livewire view

ClaudioEyzaguirre:


driesvints:
There's some changes to the registration tests in here which shouldn't be here. Feel free to resubmit with just the translation changes. Thanks.

PR #629, opened 2021-01-11
[2.x] Skip Registration tests when registration feature is not enabled

ClaudioEyzaguirre:

### DIFF
--- a/stubs/tests/RegistrationTest.php
+++ b/stubs/tests/RegistrationTest.php
@@ -4,6 +4,7 @@ namespace Tests\Feature;
 
 use App\Providers\RouteServiceProvider;
 use Illuminate\Foundation\Testing\RefreshDatabase;
+use Laravel\Fortify\Features;
 use Tests\TestCase;
 
 class RegistrationTest extends TestCase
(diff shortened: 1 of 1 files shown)


driesvints:
This has been attempted before and was rejected. If you don't have the functionality enabled in your app, remove the tests.

ClaudioEyzaguirre:
@driesvints Yes, delete the tests is an option, but sometimes you need to temporarily disable user registration while early adopters (created manually) are testing your app. Then when you enable user registration again tests will be there :)

driesvints:
You can always copy them back in.

joelbutcher:
Or... add these conditions into the tests when they're published? The use-case you've given is extremely niche to warrant a change like this.

driesvints:
It doesn't makes sense that end-app tests cover two different scenarios when only one scenario is the actual behavior of the app. You own these files, feel free to do with them as you please.

driesvints:
Ah sorry @joelbutcher, I misunderstood you there.

joelbutcher:
No problems!

ClaudioEyzaguirre:
Thanks for your time @driesvints! As @joelbutcher said I always can modify this tests...that's why Jetstream is awesome :heart:

PR #630, opened 2021-01-12
[2.x] Jetstream ignoring sessions from session connection in config

sooand:
When using sessions in a separate database connection, the sessions table is not found when querying or deleting sessions. This pull request takes the session.connection supplied in the config, so the session management for Jetstream can follow the users configs and/or .env values for session connection.

The changes has been made both for Inertia and Livewire stack. 

Should be no breaking changes as it uses the same way Laravel/framework gets the session connection in the session manager: https://github.com/laravel/framework/blob/28e0f61bba1cbc727383d183b7116eb723de7660/src/Illuminate/Session/SessionManager.php
line: 99

### DIFF
--- a/src/Http/Controllers/Inertia/OtherBrowserSessionsController.php
+++ b/src/Http/Controllers/Inertia/OtherBrowserSessionsController.php
@@ -43,7 +43,7 @@ class OtherBrowserSessionsController extends Controller
             return;
         }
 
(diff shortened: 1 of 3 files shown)


thyseus:
+1 thanks a lot - i just encountered this issue and was going to make a PR. I am 9 days too late- its your fame :) @sooand

PR #631, opened 2021-01-12
[2.x] Use Aliases feature of Laravel Mix

studnitz:
Laravel Mix 6.0 comes with [Webpack Aliases](https://laravel-mix.com/docs/6.0/aliases
) right baked in, thus removing the need for a separate `webpack.config.js` and one less config file to worry about.

### DIFF
--- a/stubs/inertia/webpack.config.js
+++ b/stubs/inertia/webpack.config.js
@@ -1,9 +1,0 @@
-const path = require('path');
-
-module.exports = {
-    resolve: {
-        alias: {
-            '@': path.resolve('resources/js'),
-        },
-    },
-};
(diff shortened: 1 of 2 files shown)


driesvints:
I guess the choice would have to be made of providing a webpack config file out of the box or not but I do agree this simplifies things.

studnitz:
I'm always happy when I don't have to deal with a webpack config file. Not exposing the webpack config file could also encourage users to look first if Laravel Mix can do the job instead of trying to make the webpack config work.

bonzai:
There is a reason for a separate config - #328.

studnitz:
Hmm, yeah that would make the life of PHPStorm users a bit harder, I guess. In that case this should be revisited, when PHPStorm does not need that workaround anymore.

PR #636, opened 2021-01-15
[2.x] Allow configuration of profile photo disk

alexleeelkins:
This change will allow the user to specify on which disk profile photos should be stored via the Jetstream config file.

Currently, the only disk that will be used is `public` (unless you add `VAPOR_ARTIFACT_NAME` to the `.env` file [sort of an override]).

I don't think it's uncommon for all images to be stored on S3 (or some other disk) for apps, so I think this could benefit a number of users.

Please let me know if anything can be improved. I didn't explicitly add the key to the config file so as to not clutter it.

Thanks!

### DIFF
--- a/src/HasProfilePhoto.php
+++ b/src/HasProfilePhoto.php
@@ -76,6 +76,6 @@ trait HasProfilePhoto
      */
     protected function profilePhotoDisk()
     {
-        return isset($_ENV['VAPOR_ARTIFACT_NAME']) ? 's3' : 'public';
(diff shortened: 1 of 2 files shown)


driesvints:
What prevents you from just overriding the method on the model?

alexleeelkins:
> What prevents you from just overriding the method on the model?

Nothing, and that would also accomplish the goal, though it seemed a little funky that it was hardcoded to use a specific disk.

driesvints:
Hmm, I don't think we should add an extra config option just for this if you can simple overwrite the method.

taylorotwell:
The idea is that you would override and customize the method per the documentation.

PR #639, opened 2021-01-16
Fixed the missing border and padding on input.

juno-w:
Fix missing border and padding after "form-input" class removed from v2.x


bonzai:
Are you sure these properties are missing?

Jetstream uses `@tailwindcss/forms` package, so the only time you won't see them is if you don't add `type` attribute to your `input` element:

https://github.com/tailwindlabs/tailwindcss-forms/blob/master/src/index.js

jamesbhatta:
I was also not getting padding in any input element through out all the forms. Disabling cache and refreshing the page helped me out.

PR #640, opened 2021-01-18
[2.x] Fix overriding TeamInvitation model in application

eudj1n:
Hi! If we'll change table name in own TeamInvitation model in Laravel project (child model, based on Laravel\Jetstream\TeamInvitation), we got error while trying to delete and/or accept invitations:

`Base table or view not found: 1146 Table 'team_invitations' doesn't exist.`

This pull allows to use project model instead jetstream model to perform these actions.

### DIFF
--- a/src/Http/Controllers/TeamInvitationController.php
+++ b/src/Http/Controllers/TeamInvitationController.php
@@ -7,7 +7,7 @@ use Illuminate\Http\Request;
 use Illuminate\Routing\Controller;
 use Illuminate\Support\Facades\Gate;
 use Laravel\Jetstream\Contracts\AddsTeamMembers;
-use Laravel\Jetstream\TeamInvitation;
+use Laravel\Jetstream\Jetstream;
 
 class TeamInvitationController extends Controller
 {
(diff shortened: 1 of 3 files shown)


src/Http/Controllers/TeamInvitationController.php
```diff
@@ -7,19 +7,21 @@
...
     /**
      * Accept a team invitation.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @param  \Laravel\Jetstream\TeamInvitation  $invitation
+     * @param  int  $invitation
      * @return \Illuminate\Http\RedirectResponse
      */
-    public function accept(Request $request, TeamInvitation $invitation)
+    public function accept(Request $request, $invitationId)
```
driesvints, line 21: Changing the method signature here is a breaking change.

src/Http/Controllers/TeamInvitationController.php
```diff
@@ -38,11 +40,13 @@ public function accept(Request $request, TeamInvitation $invitation)
      * Cancel the given team invitation.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @param  \Laravel\Jetstream\TeamInvitation  $invitation
+     * @param  int  $invitationId
      * @return \Illuminate\Http\RedirectResponse
      */
-    public function destroy(Request $request, TeamInvitation $invitation)
+    public function destroy(Request $request, $invitationId)
```
driesvints, line 46: Changing the method signature here is a breaking change.

joelbutcher:
I'm curious to know the reason why you'd want to change the table name?

eudj1n:
> I'm curious to know the reason why you'd want to change the table name?

I have a complex SaaS platform with multiple modules, and each module has own table prefix (similar to class namespaces), for example - sys for system tables, ref for references, etc.

src/Http/Controllers/TeamInvitationController.php
taylorotwell: It isn't possible to override these controllers.

taylorotwell:
This doesn't fully update all the references to TeamInvitation. Not sure I have any plans to change this right now.

PR #641, opened 2021-01-18
Skip DeleteAccountTest if feature is not enabled

arubacao:
The tests for `DeleteAccountTest` are always executed. Therefore tests fail when Account Deletion Features are disabled.

With this PR DeleteAccountTests are skipped when the feature is deactivated. 

Screenshot 2021-01-18 at 16 39 07


driesvints:
Just delete the tests if you don't need them.

arubacao:
I delete all tests that I don't need. This PR just makes the stubs more consistent.

PR #642, opened 2021-01-18
[2.x] Validation message for existing invitation

eckelarsson:
When an email has already been invited but the same email address also gets invited to another team, but not yet accepted the past invitation, the error message implies that the email is already added to the current team. So the error message could be a bit confusing for the end-user. So I think it's better to inform the email address already has a pending invitation to avoid confusion.


driesvints:
We can't change this as this will break all apps' copy who have already translated this.

eckelarsson:
Ah, I see, that is a big change..
Tho, the error message is wrong in this scenario. So the change would be to do a new check that gives the right error message if that scenario as described above happens.

taylorotwell:
Translate it to something else in your app.

PR #650, opened 2021-01-24
[2.x] Hide AppLayout header if no slot defined in Page component

nicolasbeauvais:
On Jetstream Inertia when creating a Page component with no #header slot, the AppLayout display an empty unnecessary header instead of no header.

Before:
Opera Snapshot_2021-01-24_133749_next-ci localhost

After:
Opera Snapshot_2021-01-24_133921_next-ci localhost

### DIFF
--- a/stubs/inertia/resources/js/Layouts/AppLayout.vue
+++ b/stubs/inertia/resources/js/Layouts/AppLayout.vue
@@ -215,7 +215,7 @@
             </nav>
 
             <!-- Page Heading -->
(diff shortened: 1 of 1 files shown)


taylorotwell:
Same change needs to be made to Livewire stack.

taylorotwell:
I added it to the Livewire stack.

nicolasbeauvais:
Thanks for merging and replicating the change to the Livewire stack :+1:

PR #652, opened 2021-01-26
Replacing the Name type with the Email type for Member Email input to Add Team Member form.

VKambulov:
I believe that replacing the Name with Email for the form Add Team Member would be appropriate in this case.


driesvints:
Seems like something went wrong with branching here.

PR #655, opened 2021-01-28
[2.x] Check for registration feature in test stub

TheoKouzelis:
Would you consider changes like this where the feature tests check that the relevant Fortify feature is enabled? 

This change would stop published tests from failing when users disable some of the features in Fortify.

To clean up the above code I could create a helper function called "hasAccountRegistrationFeatures" in the Fortify to follow the convention I've seen in the Jetstream repo


driesvints:
Remove the tests if you don't have the feature enabled.

PR #661, opened 2021-02-04
[2.x] Allow custom classes on the dropdown class

natecorkish:
To keep code duplication to a minimum, people re-use components from jet stream, for example the dropdown, currently the dropdown only has a pre-set alignment, a person may want to show the dropdown on the left-side on mobile phones, but to the right on a web browser. Currently they don't have an option for that. So; with this PR they could do the following:

```
<x-jet-dropdown align="other" width="60" dropdownClasses="left-0 md:right-0">
    // ....
</x-jet-dropdown>
```


PH7-Jack:
that way, if not passed the **dropdownClasses** by parameter will give variable error

PR #662, opened 2021-02-06
[2.x] add "prop" helper in inertia, so "$page.props.user.roles" can be "prop('user.roles')"

KMLX28:
Instead of `$page.props...` this pull request adds a little helper `prop()` to get any props with behavior similar to `config()` in Laravel.

This helper also support dot notation: 
```
prop('user.roles')
```
and default value:
```
prop('user', 'guest')
```
Reasons: 
1. It's a cleaner way to access props.
2. avoid `Cannot read property "..." of undefined` errors when using `$page.props...`

Notes: 
1. In my projects I put it in helpers.js file, but here I put it in app.js until we find a better place.
2. I initialize a variable for vue app to use it in the helper.


taylorotwell:
You can add this to your own application.

PR #666, opened 2021-02-09
[3.x] Upgrade VueJS to version 3

octoper:
Upgrade Inertia stack from VueJS v2 to VueJS v3

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -268,7 +268,7 @@ EOF;
         // Install NPM packages...
         $this->updateNodePackages(function ($packages) {
             return [
-                '@inertiajs/inertia' => '^0.8.2',
+                '@inertiajs/inertia' => '^0.8.4',
                 '@inertiajs/inertia-vue3' => '^0.3.5',
                 '@tailwindcss/forms' => '^0.2.1',
                 '@tailwindcss/typography' => '^0.3.0',
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -269,16 +269,15 @@ EOF;
         $this->updateNodePackages(function ($packages) {
             return [
                 '@inertiajs/inertia' => '^0.8.2',
-                '@inertiajs/inertia-vue' => '^0.5.4',
+                '@inertiajs/inertia-vue3' => '^0.3.5',
                 '@tailwindcss/forms' => '^0.2.1',
                 '@tailwindcss/typography' => '^0.3.0',
-                'portal-vue' => '^2.1.7',
                 'postcss-import' => '^12.0.1',
                 'tailwindcss' => '^2.0.1',
                 'autoprefixer' => '^10.0.2',
-                'vue' => '^2.5.17',
-                'vue-loader' => '^15.9.6',
-                'vue-template-compiler' => '^2.6.10',
+                'vue' => '^3.0.5',
+                '@vue/compiler-sfc' => '^3.0.5',
+                'vue-loader' => '^16.1.2',
             ] + $packages;
         });
 
--- a/stubs/inertia/resources/js/Jetstream/Dropdown.vue
+++ b/stubs/inertia/resources/js/Jetstream/Dropdown.vue
@@ -29,6 +29,8 @@
 </template>
 
 <script>
+import {onMounted, onUnmounted, ref} from "vue";
+
     export default {
         props: {
             align: {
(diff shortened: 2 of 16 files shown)


driesvints:
@octoper fyi: this PR won't be reviewed until it's out of draft

octoper:
Yeah I know that's why I marked it as daft 😅

octoper:
Yeah I know that's why I marked it as daft 😅

octoper:
cc @driesvints Ready for review!

driesvints:
@octoper can you please fix the StyleCI warnings so everything's green?

octoper:
@driesvints Done!

driesvints:
@reinink does this look good to you?

stubs/inertia/resources/js/Jetstream/Dropdown.vue
```diff
@@ -55,10 +55,6 @@
                 }
             }
 
-            this.$once('hook:destroyed', () => {
-                document.removeEventListener('keydown', closeOnEscape)
-            })
-
```
taylorotwell, line 61: So where does this get removed now?

octoper: Whoops totally forgot about that, I will add it tommorow, this hook removed at first because the depreciation of `$once` function
but this willbe moved in the `beforeUnmount` lifecycle hook without the `$once` function

stubs/inertia/resources/js/Jetstream/Modal.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <portal to="modal">
+    <Teleport to="#modal">
```
reinink, line 2: I haven't tested this, but do you really need the `<div id="modal"></div>`? I think you can just do `to="body"`. Also, I'm pretty sure even Vue uses the lowercase version of `<teleport>` in their docs, so that might be more idiomatic.
```suggestion
    <teleport to="body">
```

stubs/inertia/resources/js/Jetstream/Modal.vue
```diff
@@ -25,11 +25,13 @@
                 </transition>
             </div>
         </transition>
-    </portal>
+    </Teleport>
```
reinink, line 28:
```suggestion
    </teleport>
```

stubs/inertia/resources/views/app.blade.php
```diff
@@ -19,5 +19,8 @@
     </head>
     <body class="font-sans antialiased">
         @inertia
+
+        <!-- Modal Teleport -->
+        <div id="modal"></div>
```
reinink, line 24: Not required, as per comment above.
```suggestion
```

stubs/inertia/resources/js/Pages/Auth/Register.vue
```diff
@@ -30,7 +30,7 @@
             <div class="mt-4" v-if="$page.props.jetstream.hasTermsAndPrivacyPolicyFeature">
                 <jet-label for="terms">
                     <div class="flex items-center">
-                        <jet-checkbox name="terms" id="terms" v-model="form.terms" />
+                        <jet-checkbox name="terms" id="terms" v-model:checked="form.terms"/>
```
reinink, line 33: Missing space.
```suggestion
                        <jet-checkbox name="terms" id="terms" v-model:checked="form.terms" />
```

src/Console/InstallCommand.php
```diff
@@ -269,16 +269,15 @@ protected function installInertiaStack()
         $this->updateNodePackages(function ($packages) {
             return [
                 '@inertiajs/inertia' => '^0.8.2',
```
reinink, line 271: Might be worth bumping this to `0.8.4`, given the recent axios vulnerability (see [here](https://github.com/inertiajs/inertia/releases/tag/v0.8.4)).
```suggestion
                '@inertiajs/inertia' => '^0.8.4',
```

octoper:
Fixed everything @reinink and @taylorotwell mentioned

stubs/inertia/resources/js/Jetstream/Dropdown.vue
```diff
@@ -55,13 +55,19 @@
...
             document.addEventListener('keydown', closeOnEscape)
         },
 
+        unmounted() {
+            const closeOnEscape = (e) => {
```
reinink, line 62: Hmm, does redefining the entire function actually work when removing it? I don't think so. I think you need the reference to the original function to remove it.

Here's what I believe is the proper Vue 3 way to handle this. Remove both the `mounted()` and `unmounted()` methods, and then add a new `setup()` method instead. Something like this:

```js
import { onMounted, onUnmounted } from 'vue'

setup () {
  const closeOnEscape = (e) => {
      if (this.open && e.keyCode === 27) {
          this.open = false
      }
  }
  
  onMounted(() => document.addEventListener('keydown', closeOnEscape))
  onUnmounted(() => document.removeEventListener('keydown', closeOnEscape))
},
```

Again, this is untested, so definitely do test it, but this should work. 👍

stubs/inertia/resources/js/Jetstream/Modal.vue
```diff
@@ -71,10 +73,6 @@
             }
 
             document.addEventListener('keydown', closeOnEscape)
-
-            this.$once('hook:destroyed', () => {
```
reinink, line 75: However we solve this in the `Dropdown.vue` component, needs to also be done here.

taylorotwell:
@octoper is this PR ready to go from your perspective now?

octoper:
Yeah it's ready from me @taylorotwell

khatabwedaa:
when this will be merged?

taylorotwell:
Thanks!

AbdelElrafa:
@taylorotwell will you be tagging a release soon so that this can be used, or should I just target the branch?

driesvints:
@AbdelElrafa releases are always done on Tuesdays. You should only ever target a dev branch for development and not for production releases.

PR #667, opened 2021-02-09
[2.x] Fix create user for api

gildastema:
I noticed that when I run tests. With the option team deactivates its generates an exception when creating the user.

        $this->actingAs($user = User::factory()->withPersonalTeam()->create());


driesvints:
Update your tests for the new behavior

PR #668, opened 2021-02-09
[2.x] Fix overflow hidden issue with overlaping elements inside form section

altercode:
Hi,

I found issue with form section when it has small height and we have bigger elements inside it that overextends
the height of the form section. For example datapicker, dropdown etc. I attached screen below to show the issue.

My pull request fix that issue by slightly change the classes in form section components.

I have modified inertia and livewire component.

issue

### DIFF
--- a/resources/views/components/form-section.blade.php
+++ b/resources/views/components/form-section.blade.php
@@ -8,18 +8,16 @@
 
     <div class="mt-5 md:mt-0 md:col-span-2">
         <form wire:submit.prevent="{{ $submit }}">
-            <div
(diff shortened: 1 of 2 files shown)


taylorotwell:
Doesn't make sense to me to only have the top corners rounded when there are actions?

taylorotwell:
An type="date" input looks fine to me on Chrome...

image

altercode:
But for example custom datepicker has the behavior like on my screenshot. My pr remove the overflow-hidden from paren div and ads rounded corners at top left, top right and at the bottom of the form section. Even if there is no actions, rounded corners will be present at the bottom.

taylorotwell:
Got it.

PR #670, opened 2021-02-09
Code of Conduct Policy Acceptance

blakekrone:
This PR introduces a Code of Conduct acceptance to the existing Terms of Use and Privacy Policy feature of Jetstream. 

This policy mimics the existing policies entirely and utilizes the same markdown process. 

No additional form fields are submitted and all tests follow the same feature tests that exist as this is integrated into the Features::termsAndPrivacyPolicy() configuration. I elected not to change the name, however it might be advisable to change the name to simply usagePolicies() to better capture potential future policies that might be required by developers.

### DIFF
--- a/routes/inertia.php
+++ b/routes/inertia.php
@@ -3,6 +3,7 @@
 use Illuminate\Support\Facades\Route;
 use Laravel\Jetstream\Http\Controllers\CurrentTeamController;
 use Laravel\Jetstream\Http\Controllers\Inertia\ApiTokenController;
+use Laravel\Jetstream\Http\Controllers\Inertia\CodeOfConductController;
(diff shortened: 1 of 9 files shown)


taylorotwell:
You can add this to your own application if you wish. For many applications (for example, an accounting / bookkeeping application) this type of field doesn't make sense. For some (conference sign-up website), it would.

blakekrone:
I respectfully disagree @taylorotwell that it doesn’t apply to other applications. Any application that accepts comments, notes, etc is at a risk of conduct issues within an organization and should be easy for folks to implement policies. Organizations are responsible for all methods of communications  in today’s culture. I could expand the PR to be a separate policy if desired but feel it is something that should be in the core framework to encourage more usage and visibility.

PR #674, opened 2021-02-12
translate validation errors

jimmitjoo:
Discovered that errors was not translated in Jetstream views.


jimmitjoo:
Found out that this was a problem with Fortify, that does not match the default translation keys in Laravel.

In Laravel: "The :attribute must be at least :min characters."
In Fortify: "The :attribute must be at least :length characters."

PR #685, opened 2021-02-18
[2.x] Return null for profile photo url if `managesProfilePhotos` is disabled

owenconti:
This PR changes the `getProfilePhotoUrlAttribute()` to return `null` if the application does not have profile photos enabled.

The reason for this is in the Spark billing portal, the profile photo is display as initials, even when the application has profile photos disabled:

Screen Shot 2021-02-18 at 8 27 15 AM


taylorotwell:
Might be a breaking change for some users that they don't want. I think you could overwrite this method in your own model?

PR #691, opened 2021-02-26
Update confirms-password.blade.php

acacha:
Dialog will be closed instantly by Javascript, so in slow connections a performance boost will be seen by final users improving UX.


taylorotwell:
Will just keep it as is for now. You can modify this in your own templates if you want.

PR #695, opened 2021-02-28
Prevent network request when pressing escape to close modal

freekmurze:
This PR contains a micro-optimisation, but a nice one.

Right now, if you close the modal with `esc` a network request is being made. 

By using `defer`, no network request will be made when hitting escape. I've confirmed this improved behaviour in one of my projects.

Livewire docs on `defer` are here: https://laravel-livewire.com/docs/2.x/alpine-js#extracting-blade-components

### DIFF
--- a/resources/views/components/modal.blade.php
+++ b/resources/views/components/modal.blade.php
@@ -14,7 +14,7 @@ $maxWidth = [
 
 <div
     x-data="{
-        show: @entangle($attributes->wire('model')),
+        show: @entangle($attributes->wire('model')).defer,
         focusables() {
             // All focusable element types...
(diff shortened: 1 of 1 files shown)


Loots-it:
Wouldn't it be an idea to open and close the modal without any network requests using alpine?

https://laravel-livewire.com/docs/2.x/alpine-js

I did this in my project by adding an outer div to the modal view which only has the x-data and x-init of the original modal where all the 'this' references are replaced by 'this.childNodes[1]'. Then I added one div child to this div where the user of the modal can add a 'handler' to open the modal, this handler only needs '@click="show = true"' to work.

PR #697, opened 2021-03-01
Skip TwoFactorAuthenticationSettingsTest if not enabled

julionc:
Currently, the `TwoFactorAuthenticationSettingsTest` is always executed, but fails when Two-Factor Authentication feature is disabled.

This PR is just a fix to make sure (stubs) Test suite works as expected.


driesvints:
Delete the tests if you're not using 2FA.

PR #698, opened 2021-03-03
Test team features only when active

tedsecretsource:
Add a condition in the Inertia and Livewire tests to only test team features if Teams have been enabled. At present, some tests are failing because they expect teams to always be enabled.

This is my first, proper, PR for Laravel, so, if there is a better way to address this issue (like making `withPersonalTeam()` return null if the teams feature isn't enabled), I'm more than happy to head down that path, just point me in that direction!

### DIFF
--- a/stubs/tests/inertia/ApiTokenPermissionsTest.php
+++ b/stubs/tests/inertia/ApiTokenPermissionsTest.php
@@ -18,7 +18,7 @@ class ApiTokenPermissionsTest extends TestCase
             return $this->markTestSkipped('API support is not enabled.');
         }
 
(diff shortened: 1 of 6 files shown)


tedsecretsource:
@driesvints I see you've been instructing people to just delete the tests they aren't using. To be clear, in this case, I am using these tests, just not with the teams feature(s). Also, for future reference, I see tests/P7.3 - L^8.0 is failing. Is this something I can resolve and if so, any pointers on how?

driesvints:
@tedsecretsource tests are failing because of new GH Actions ubuntu builds. I'll fix them today.

PR #701, opened 2021-03-07
[2.x] Add an error message output when submitting a form

VKambulov:
Sometimes forms are large and when submitting them with incorrect data, we do not receive an error message near the form submit button, as it happens when the form is successfully submitted. Adding an error message output on form submission will help when submitting large and long forms that may not fit on the page, or when scrolling the page below invalid fields and clicking the submit button. It will also be convenient on mobile devices, especially with small screens where even a relatively small form will no longer fit on the screen.
Another solution would be to add `recentlyError` to the **Inertia** package, but the developer can not always output messages only from the **Inertia** form, so an additional option could be to transfer his own properties of success (`onSuccess`) or failure (`onError`) into the `ActionMessage` component and tracking them there so that it looks like passing the `recentlySuccessful` property from the **Inertia** form. For compatibility, and also, if the developer wants to display only a message about the success of the **Inertia** form submission, it is possible to transfer only the `on` property as it was before without any changes.

For example, there if we type wrong name in form on **Profile** page, we not getting error. We only need to guess that something was wrong.
without_error (iPhone 5_SE)

But, if show error message, everything will become clear.
with_error (iPhone 5_SE)

Perhaps my code needs to be corrected, since I don't have much experience with this.

### DIFF
--- a/stubs/inertia/resources/js/Jetstream/ActionMessage.vue
+++ b/stubs/inertia/resources/js/Jetstream/ActionMessage.vue
@@ -1,15 +1,61 @@
 <template>
-    <div>
-        <transition leave-active-class="transition ease-in duration-1000" leave-from-class="opacity-100" leave-to-class="opacity-0">
-            <div v-show="on" class="text-sm text-gray-600">
-                <slot />
-            </div>
-        </transition>
+    <div class="text-sm">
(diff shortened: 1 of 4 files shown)


taylorotwell:
Feel free to change your own Jetstream application templates to whatever use case you need regarding error messages.

PR #708, opened 2021-03-17
Reduce amount of queries for Teams feature

lexdewilligen:
The inertia middleware which is used to share data to the frontend currently runs 3 queries to load all teams and the current team. 

1. Load all owned teams
2. Merge all other teams 
3. Load the currentTeam

However, this PR removes the last query since the currentTeam is one of the teams that's loaded in allTeams. This might just be a slight performance improvement! I hope it's worth merging into Jetstream.

### DIFF
--- a/src/Http/Middleware/ShareInertiaData.php
+++ b/src/Http/Middleware/ShareInertiaData.php
@@ -42,7 +42,9 @@ class ShareInertiaData
                 }
 
                 if (Jetstream::hasTeamFeatures() && $request->user()) {
-                    $request->user()->currentTeam;
+                    $request->user()->setRelation('currentTeam', $request->user()->allTeams()->filter(function ($team) use ($request) {
+                        return $team->id === $request->user()->current_team_id;
+                    })->first());
                 }
 
                 return array_merge($request->user()->toArray(), array_filter([


src/Http/Middleware/ShareInertiaData.php
```diff
@@ -42,7 +42,9 @@ public function handle($request, $next)
                 }
 
                 if (Jetstream::hasTeamFeatures() && $request->user()) {
-                    $request->user()->currentTeam;
+                    $request->user()->setRelation('currentTeam', $request->user()->allTeams()->filter(function ($team) use ($request) {
+                        return $team->id === $request->user()->current_team_id;
```
matthieumota, line 46: Hi,

I think that PR cause a bug due to ```current_team_id``` field who can be null. Just try to install a fresh Laravel with Jetstream and you're unable to login (white page due to JS error and team).

It's because field is refresh with ```switchTeam``` method. Before PR, ```currentTeam()``` was called and check that.

Maybe you can try that 👍 
```
$request->user()->setRelation('currentTeam', $request->user()->allTeams()->filter(function ($team) use ($request) {
    if (is_null($request->user()->current_team_id)) {
        $request->user()->switchTeam($request->user()->personalTeam());
    }

    return $team->id === $request->user()->current_team_id;
})->first());
```

src/Http/Middleware/ShareInertiaData.php
```diff
@@ -42,7 +42,9 @@ public function handle($request, $next)
...
+                    })->first());
```
matthieumota, line 47: Hi,

I think that PR cause a bug due to current_team_id field who can be null. Just try to install a fresh Laravel with Jetstream and you're unable to login (white page due to JS error and team).

It's because field is refresh with switchTeam method. Before PR, currentTeam() was called and check that.

Maybe you can try that 👍

```
if (is_null($request->user()->current_team_id)) {
    $request->user()->switchTeam($request->user()->personalTeam());
}

return $team->id === $request->user()->current_team_id;
```

driesvints: Thanks, reverted.

driesvints:
@lexdewilligen this broke login/registration for Inertia, see https://github.com/laravel/jetstream/pull/708#discussion_r600939435

I've reverted this and tagged a new release. Next time, please try to test you change thoroughly before sending this in.

lexdewilligen:
@driesvints I did test it thoroughly, but I must have missed it! I'm very sorry this has happened!

PR #711, opened 2021-03-18
[2.x] Add a tag dropdown link

dillingham:
PR adds `<a>` tag to the `<jet-dropdown-link>` component.

When adding spark billing, it would be cleaner to:
```html
<jet-dropdown-link as="a" :href="route('spark.portal')">
    Billing
</jet-dropdown-link>
```
compared to manually adding the tailwind classes to `<a>` to avoid inertia navigation.
```html
<a :href="route('spark.portal')" class="block px-4 py-2 text-sm leading-5 text-gray-700 hover:bg-gray-100 focus:outline-none focus:bg-gray-100 transition duration-150 ease-in-out">
    Billing
</a>
```
Used `as="a"` to stay in line with `as="button"`


driesvints:
`as="link"` might be more descriptive?

taylorotwell:
@driesvints it's an inertia-link by default so as=link might sound like inertia link.

PR #712, opened 2021-03-18
[2.x] Button Disabled opacity-25

fbrissi:
Add disabled attribute  in `danger-button.blade.php` and `secondary-button.blade.php` the same way that int the `button.blade.php`


taylorotwell:
You would need to update the `inertia` stack as well.

PR #713, opened 2021-03-18
Add Feature Hide Sections

fbrissi:
I think it's a good idea to allow sections to be hidden.

image

image


taylorotwell:
This can be added to your own application if you wish.

fbrissi:
Right @taylorotwell, I already implemented it in my application.
I thought it would be a good idea to have this in the core.
Thanks.

PR #715, opened 2021-03-19
[2.x] Button component href handling

dillingham:
PR adds `href` support to `<jet-button>` to allow using button style with links.

Prevents needing to create another button style definition.

```html
<jet-button :href="route('projects.create')">
    New
</jet-button>
```
Otherwise would need to duplicate the tailwind classes:
```html
<inertia-link :href="route('projects.create')" class="inline-flex items-center px-4 py-2 bg-gray-800 border border-transparent rounded-md font-semibold text-xs text-white uppercase tracking-widest hover:bg-gray-700 active:bg-gray-900 focus:outline-none focus:border-gray-900 focus:shadow-outline-gray transition ease-in-out duration-150">
    New
</inertia-link>
```

Also allows external links using the `as="a"` convention in `<jet-nav-link>`

```html
<jet-button :href="order.stripe_link" as="a">
    View Invoice
</jet-button>
```

jetstream button


dillingham:
I almost think this might be button scope creep.. but PRing it incase not :)

PR #720, opened 2021-03-21
Added "SwitchTeam" action - Allows to modify behaviour if user switches team.

camya:
**PR "SwitchTeam" Jetstream/Action**

This PR allows developers to change the "team switch" implementation in their app by modifying `app/Actions/Jetstream/SwitchTeam.php`. 

A team switch could be triggerd by the user by selecting a Team in the Teams dropdown.

_Usecase 1:_ Redirect to the team's landing page after the user switched the team.
_Usecase 2:_ Show the form of the team, if some required informations for the Team are missing from the user.

The feature is available in new Jetstream installations automatically without any required modification.

For exising projects, a simple migration path is explained below. 

Following a short example.

**Example for the "SwitchTeam" Jetstream/Action**

Redirect to the route "teams.show" of the selected team.

```
// app/Actions/Jetstream/SwitchTeam.php
public function switch($user, int $teamId)
{
    // Get team by id, switch user's current team, than redirect.

    return redirect(route('teams.show', $team));
}
```
**Migration path - Existing Jetstream Projects**

1.) artisan: Publish "jetstream-actions" with artisan command

`php artisan vendor:publish --tag=jetstream-actions`

2.) Manually update JetstreamServiceProvider, add Jetstream::switchTeamUsing()

```
// app/Providers/JetstreamServiceProvider.php - boot()
Jetstream::switchTeamUsing(SwitchTeam::class);
```

**Troubleshooting**

Problem: Error message "`Target [Laravel\Jetstream\Contracts\SwitchesTeams] is not instantiable.`"

Solution: This occurs in existing Jetstream installations only. The Jetstream::switchTeamUsing(SwitchTeam::class) line is missing in the JetstreamServiceProvider. (See "Migration path")

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -447,6 +447,7 @@ EOF;
         copy(__DIR__.'/../../stubs/app/Actions/Jetstream/InviteTeamMember.php', app_path('Actions/Jetstream/InviteTeamMember.php'));
         copy(__DIR__.'/../../stubs/app/Actions/Jetstream/RemoveTeamMember.php', app_path('Actions/Jetstream/RemoveTeamMember.php'));
         copy(__DIR__.'/../../stubs/app/Actions/Jetstream/UpdateTeamName.php', app_path('Actions/Jetstream/UpdateTeamName.php'));
+        copy(__DIR__.'/../../stubs/app/Actions/Jetstream/SwitchTeam.php', app_path('Actions/Jetstream/SwitchTeam.php'));
 
         copy(__DIR__.'/../../stubs/app/Actions/Fortify/CreateNewUserWithTeams.php', app_path('Actions/Fortify/CreateNewUser.php'));
 
--- a/src/Contracts/SwitchesTeams.php
+++ b/src/Contracts/SwitchesTeams.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Laravel\Jetstream\Contracts;
+
+interface SwitchesTeams
+{
+    /**
+     * @param mixed $user
+     * @param int   $teamId
+     *
+     * @return mixed
+     */
+    public function switch($user, int $teamId);
+}
--- a/src/Http/Controllers/CurrentTeamController.php
+++ b/src/Http/Controllers/CurrentTeamController.php
@@ -8,7 +8,6 @@ use Laravel\Jetstream\Contracts\SwitchesTeams;
 
 class CurrentTeamController extends Controller
 {
-
     /**
      * Switches the authenticated user's current team.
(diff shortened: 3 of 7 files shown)


camya:
**Reason why the integration tests not successful:** 

This PR requires some modifications for existing projects. See "Migration path - Existing Jetstream Projects".

The error "Response status code [500] is not a redirect status code." is throw, because the Jetstream::switchTeamUsing(SwitchTeam::class) line is missing in the JetstreamServiceProvider. (See Troubleshooting)

```
Laravel\Jetstream\Tests\CurrentTeamControllerTest::test_can_switch_to_team_the_user_belongs_to
Response status code [500] is not a redirect status code.
```

driesvints:
There's several breaking changes here and tests are indeed not passing yet as you indicate. Please send this to master with passing tests.

camya:
At the moment it's not possible to add new Jetstream/Actions without making a breaking change because of the required modification of the  app/Providers/JetstreamServiceProvider.php file. I'm still open for suggestions how to implement the feature without a change in Jetstream. :)

**Idea:**

It's possible to `str_replace()` the line "Jetstream::switchTeamUsing(SwitchTeam::class);" to the JetstreamServiceProvider.php in order to make it work, but this has to be added into the publishing mechanism in my opinion. (php artisan vendor:publish --tag=jetstream-actions) Such a feature is not available in "vendor:publish" by now. Should be some kind of callback like 

```
// publishes(array $paths, $groups = null, $callback = null)  in vendor/laravel/framework/src/Illuminate/Support/ServiceProvider.php

// Implementation
$this->publishes([
    __DIR__.'/../routes/'.config('jetstream.stack').'.php' => base_path('routes/jetstream.php'),
], 'jetstream-routes', function($when, array $items) {
    switch ($when) {
        case 'pre':
            // Before
            // - Prepare something before any file is updated.
            // - Use this to check, if some items of $items already exits.
            break;
        case 'post':
            // After
            // - Modify service provider here after all files are published
            // - Clean up
            break;
    }
});
```

The Jetstream installation does this inside the InstallCommand.php using methods like "installServiceProviderAfter()". It searches for a string in the service provider php file and appends another string.

**Conclusion**

Because I can't easily implement the idea described above, I have to roll back this PR. :( What a pity.

It looks like the only chance to add simple requirements like redirects is to fork the Jetstream project completely.

**Question**  

@taylorotwell & @driesvints I really like Jetstream and the Tall stack. Is there any chance that Jetstream/Actions are extended and made more configurable in JetStream 3.0, or is it better to just fork the Jetstream project in order to implement features like this  in the own project?

Best, Andreas

PR #722, opened 2021-03-22
add auto-close option

PH7-Jack:
add prop to block auto close when click inside dropdown


taylorotwell:
You are free to add this to your own project.

PR #723, opened 2021-03-23
[2.x] Load modal slot on show

dillingham:
**Problem:**

Components within modals render when `<jet-modal :show="false">` currently.

**Effects:** 

- modals with components that request data on mount.. fires on page load
- modals with forms partially completed, closed, reopened.. not fresh form

**Solution:**

Adding `v-if` to the `<slot>` insures the the `v-show` still transitions but slot contents render each `open`


driesvints:
Does this also happens on the Inertia side of things?

PR #726, opened 2021-03-24
[2.x] Get team and user by their ID when using team feature

rinodrummer:
One of the things I most love about Laravel is the fact that when operating with models you can usually provide the model object or its ID and Laravel does all the magic.

Working with Livewire, it may be really common the interaction with another model's ID instead of the whole model.

I've implemented two private methods that may increase the DX by searching automatically the User or the Team model when trying to interact with each other.

Tell me if this PR needs a bit of work on it or it will work just like this.

### Likely Improvements

I haven't spent too much time checking the existence of the model because I've seen that the code already relies on the developer, but if needed it may be adjusted by replacing the `find()` with the `findOrFail()` one.

### DIFF
--- a/src/HasTeams.php
+++ b/src/HasTeams.php
@@ -228,7 +228,8 @@ trait HasTeams
      * @param  mixed  $team
      * @return void
      */
-    private function getTeamById(&$team) {
+    private function getTeamById(&$team)
+    {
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #733, opened 2021-03-26
Enhance tests

trifisch:
Jetstream with Inertia in team mode generally expects users to have at least one team (the personal team). Not all test classes did follow this rule and created users without team even if the team feature was enabled.

We faced a problem in the `share()` method of `HandleInertiaRequests` where we wanted to get information from the user's team and share it in the request. Running the unit tests then resulted in an error, because for some unit tests the user did come without a team. 

To avoid this I propose to check if the team feature is enabled in the tests and add a team if necessary. I did not want to skip the tests because they are quite useful.

The changes have been incorporated in Inertia and Livewire test stubs. I have checked them with a fresh Jetstream installation and everything works.

Please let me know if I have to change anything.

### DIFF
--- a/stubs/tests/inertia/DeleteAccountTest.php
+++ b/stubs/tests/inertia/DeleteAccountTest.php
@@ -17,7 +17,10 @@ class DeleteAccountTest extends TestCase
             return $this->markTestSkipped('Account deletion is not enabled.');
         }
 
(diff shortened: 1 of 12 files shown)


taylorotwell:
All my tests pass on a fresh Inertia app. If you want to customize your tests further feel free to modify your own tests.

PR #734, opened 2021-03-26
Remove unnecessary classes

juno-w:
`transition` utility-class has `duration-150` and `ease-in-out` properties by default. Therefore single `transition` utility-class produce the same thing as before unless different duration and easing being used.


driesvints:
There's lots of unrelated style changes here. Please resubmit this without those.

PR #735, opened 2021-03-26
[2.x] - Remove unnecessary classes

juno-w:
`transition` utility-class has `duration-150` and `ease-in-out` properties by default. Therefore single `transition` utility-class produce the same thing as before unless different duration and easing being used.

### DIFF
--- a/resources/views/components/banner.blade.php
+++ b/resources/views/components/banner.blade.php
@@ -26,7 +26,7 @@
             <div class="flex-shrink-0 sm:ml-3">
                 <button
                     type="button"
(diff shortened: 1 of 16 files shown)


juno-w:
@driesvints Sorry, this PR just removing the classes unlike the previous one.

juno-w:
I have checked Tailwind v1.x `transition` class doesn't have `transition-duration` and `transition-timing-function` properties included by default.

driesvints:
@juno-w this package uses Tailwind v2, not v1

PR #744, opened 2021-04-02
withPersonalTeam factory method sets current team on the user

grantjanecek:
I have added this in my current project to lighten the load of adding a user with a personal team and then setting the current team. The scenarios where you want the personal team set as the current team seem to out weigh the opposite. 

This will assist in testing applications that set the team member id based on the authenticated user. This stub is only added in a new application so breaking changes should not be an issue.


taylorotwell:
Feel free to add to your own project.

PR #746, opened 2021-04-02
Fix RegistrationTest failure due to duplicate email addresses

Loupeznik:
This PR fixes the RegistrationTest which fails by default due to having the same email address for the dummy user as Livewire's ProfileInformationTest. This should be the case in the Intertia variant of the test as well.

```php
// stubs/tests/livewire/ProfileInformationTest.php
    public function test_profile_information_can_be_updated()
    {
        $this->actingAs($user = User::factory()->create());

        Livewire::test(UpdateProfileInformationForm::class)
                ->set('state', ['name' => 'Test Name', 'email' => 'test@example.com']) // THIS
                ->call('updateProfileInformation');

        $this->assertEquals('Test Name', $user->fresh()->name);
        $this->assertEquals('test@example.com', $user->fresh()->email);
    }

// stubs/tests/inertia/RegistrationTest.php
    public function test_new_users_can_register()
    {
        $response = $this->post('/register', [
            'name' => 'Test User',
            'email' => 'test@example.com', // THIS
            'password' => 'password',
            'password_confirmation' => 'password',
            'terms' => Jetstream::hasTermsAndPrivacyPolicyFeature(),
        ]);

        $this->assertAuthenticated();
        $response->assertRedirect(RouteServiceProvider::HOME);
    }
```

The test would fail due to not being able to register second user with the same email address.

**The dummy email in RegistrationTest is changed by this PR.**

### DIFF
--- a/stubs/tests/RegistrationTest.php
+++ b/stubs/tests/RegistrationTest.php
@@ -22,7 +22,7 @@ class RegistrationTest extends TestCase
     {
         $response = $this->post('/register', [
             'name' => 'Test User',
-            'email' => 'test@example.com',
+            'email' => 'testuser@example.com',
             'password' => 'password',
             'password_confirmation' => 'password',
             'terms' => Jetstream::hasTermsAndPrivacyPolicyFeature(),


taylorotwell:
Shouldn't your database be reset between each test?

taylorotwell:
Please provide exact steps to recreate this on a fresh Jetstream installation.

Loupeznik:
The database gets reset before running the whole test suite with *php artisan test* but not before this particular test. When the Livewire's test completes, the *users* table record doesn't get deleted and the record which is supposed to be created by this test cannot be inserted as the email column is set to be unique.

Steps to reproduce:
1. laravel new test-project (via Laravel installer)
2. Composer require laravel/jetstream
3. php artisan jetstream:install livewire
4. Install and build npm packages
5. Create the database and migrate
6. Run tests

You should get this result with the original test:
image

Issue #730, opened 2021-03-25
Recreating an account after account deletion doesn't work properly


Loots-it:
- Jetstream Version:2.2.4
- Jetstream Stack: Livewire
- Uses Teams: no
- Laravel Version: 8.34
- PHP Version: 8.0
- Database Driver & Version: MariaDB 10.3
- Forge provisioned server

### Description:
When testing my application I found a small bug, nothing serious, mostly annoying when testing/debugging. If you create an account and delete it, there is a small bug when you try to recreate the same account (although I think it will also happen if you try to create another account). When you register, you won't be logged in immediately, but you will be redirected to the login page. I am pretty sure this is because the login cookie of the old account isn't deleted yet. If I delete all the cookies from the site before I try to recreate the account, everything works fine.

### Steps To Reproduce:
1. register a new account
1.1. as expected, you are now logged in
2. go to profile and delete this new account
3. try to recreate the account you just deleted
3.1. this time, you are not logged in directly, but are redirected to the login page -> bug

driesvints:
I can confirm this. I'll pick this up internally to see what the best solution would be.

driesvints:
We'd appreciate a PR for this, otherwise we'll pick this up eventually.

driesvints:
I'm going to mark this as an enhancement since nothing is really broken here.

Loots-it:
If I find some time, I will look into it! It shouldn't be too difficult I think/hope.

driesvints:
I think the latter and I think the 2.x branch is fine. The first one re-uses too much a bit there I believe.

Loots-it:
Inertia has the same bug. And ok, I will use the second approach, I personally think the first one is better though. We just want to destroy the authenticated session. With the second approach, this code will be duplicated across 3 places (logout, inertia delete account, livewire delete account). I will implement the fix today or tomorrow.

PR #752, opened 2021-04-08
replace $request->session() with session()

inmanturbo:
This PR (refactor) replaces The calls to methods on `$request->session()` with  `session()` as per @calebporzio 's recommendation in https://github.com/livewire/livewire/issues/936#issuecomment-652099995.

This also eliminates the need for the `if ($request->hasSession())` condition in my previous PR, thus bringing the code back one level of indentation. 

Tested with the following steps:

```
laravel new jetstream-demo
```
```
cd jetstream-demo
```
```
composer require laravel/jetstream
```

```
php artisan jetstream:install livewire
```
```
npm install && npm run dev
```
```
vi phpunit.xml
```
```diff
-        <!-- <server name="DB_CONNECTION" value="sqlite"/> -->
-       <!-- <server name="DB_DATABASE" value=":memory:"/> -->
+       <server name="DB_CONNECTION" value="sqlite"/>
+       <server name="DB_DATABASE" value=":memory:"/>
```
```
php artisan test
```

Also tested in chromium browser against https://github.com/laravel/jetstream/issues/730

----
@telkins @Loots-it 

https://github.com/laravel/jetstream/pull/742

https://github.com/laravel/jetstream/pull/750#issuecomment-815662156

### DIFF
--- a/src/Http/Livewire/DeleteUserForm.php
+++ b/src/Http/Livewire/DeleteUserForm.php
@@ -3,7 +3,6 @@
 namespace Laravel\Jetstream\Http\Livewire;
 
 use Illuminate\Contracts\Auth\StatefulGuard;
-use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Hash;
 use Illuminate\Validation\ValidationException;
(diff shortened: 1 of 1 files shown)


taylorotwell:
Is this actually fixing a problem? It sounds like it's just a preference thing?

inmanturbo:
> It sounds like it's just a preference thing?

Yeah I suppose it is more of a preference thing, since we're really not sure if there will be a problem. `Livewire` was having issues with `\Request` but in this case it may never be a problem for `Jetstream`, which is working fine as it is.

peterfox:
Replacing $request->session() with the Session contract fixes the DeleteUserForm test for me. Otherwise it fails due to the request not having a stateful session, most likely because the way Livewire component testing works.

PR #757, opened 2021-04-15
[2.x] Exclude hidden inputs from focusable elements

bjuppa:
Currently if there is any `<input type="hidden">` inside the modal, tabbing stops when that element is next.

This PR excludes such hidden inputs via the query-selector, so that focus progression is no longer stopped.
I couldn't find any existing tests for the modal or its tabbing in this repo, so I haven't updated or added any tests.

### DIFF
--- a/resources/views/components/modal.blade.php
+++ b/resources/views/components/modal.blade.php
@@ -17,7 +17,7 @@ $maxWidth = [
         show: @entangle($attributes->wire('model')).defer,
         focusables() {
             // All focusable element types...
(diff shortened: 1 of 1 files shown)


bjuppa:
When investigating this I've come across other situations where certain elements may break the tabbing, for example with the `hidden` attribute on an otherwise interactive element or any of its ancestors. But solving that probably requires a rewrite of the current tab-progression code, checking programatically that the desired element has actually received focus before calling it a day.

It's clearly outside the scope of this PR, let me know if you want me to open an issue and propose a solution?

PR #758, opened 2021-04-15
[2.x] Disable Teams tests when the feature is disabled

rogervila:
Automatically skip tests that affect Teams feature when it is disabled.

Thank you.


driesvints:
Just delete the tests when you don't need them. This is all dead code 🤷‍♂️

PR #766, opened 2021-04-21
Use quote for cancelling team member

mtawil:
Here I changed jetstream on my app to accept UUID as primary keys, so this must be escaped to accept both int and string primary keys.


taylorotwell:
Feel free to change this in your own application if necessary.

PR #770, opened 2021-04-22
Update tailwind.config.js

julieni:
should not watch './storage/framework/views/*.php' as these are cached files


taylorotwell:
Yes - we should. They can contain compiled files from vendors that include views which are not in the typical view directories. This is in fact a **great** directory to watch.

PR #775, opened 2021-04-28
Add register route link on login view

rogervila:
When a user lands on the Login view, there is no specific link or button that points to the registration route.

This PR adds a "Not registered yet?" link on the Login view for Inertia and Livewire presets.

Captura de pantalla 2021-04-28 a las 23 28 58


taylorotwell:
You can add this to your own applications if you wish.

PR #777, opened 2021-04-30
[2.x] Fix error messages margins around rounded inputs

rennokki:
It's just a common UX mistake to not add a small left margin to elements that stick nearby to rounded elements. And even if they had an `mt-2` to be breathable, an `mt-1` makes them stick better to the input. Can be a game-changer when you have multiple inputs that need validation.

Consider the following before & after:

Before:
before

After:
after

Even if it's not too much and it is a small change, it actually does offer a better overview of the elements due to the ratio compared with the input roundness.


driesvints:
I think the before was better tbh 😅

taylorotwell:
Not sure I agree.

PR #792, opened 2021-05-26
add banner and dangerBanner request macros

ArielMejiaDev:
## Description

it adds `banner` and `dangerBanner` request macros

## Benefit to end users

it could make easier to show jetstream banner and jetstream danger banner when a resource method returns a `view()` or an `inertia response`.

there is a scenario where a method does not return a redirect, an example could be:

```php
public function index(Request $request)
{
    if($this->paymentMethod->status === PaymentMethod::NOT_AVAILABLE) {
        $request->dangerBanner('your payment method is not available anymore');
    }
    
    $servers = Server::all();
    return Inertia::render('Servers/Index', compact('servers'));
}
```

Of course the logic could be change to another file/action, but for the sake to keep it simple to explain, it is enough, it would be useful on both stacks, it is just an example with Inertia.


driesvints:
Seems like some commits from your other repo snuck in. Can you send in again and also without the extra commits? Thanks

PR #793, opened 2021-05-26
Modify two factor auth test

ArielMejiaDev:
## Description

it adds a validation on `TwoFactorAuthenticationSettingsTest` on both stacks.

## How it benefit end users

it would prevent a break on tests if two factor authentication feature is not enabled or is commented/removed in `config/fortify.php`:

```php
'features' => [
        Features::registration(),
        Features::resetPasswords(),
        Features::emailVerification(),
        Features::updateProfileInformation(),
        Features::updatePasswords(),
//        Features::twoFactorAuthentication([
//            'confirmPassword' => true,
//        ]),
    ],
```


driesvints:
Remove the tests if you don't need them.

Issue #789, opened 2021-05-24
Reconsider the text type for the profile colmn with a more limited type


rabin-io:
https://github.com/laravel/jetstream/blob/7ed0cadd8d0c1cfe03dffa8553c8807f4c4fa92f/database/migrations/2014_10_12_000000_create_users_table.php#L24

After reading about this a bit, I see that this column type is overkill for this use case, on MySQL, MSSQL and PostgresSQL this type can hold up to 2GB/4GB of data

DB | Column Type | Length
--- | --------------- | -------
PostgreSQL | TEXT, VARCHAR | variable unlimited length. https://www.postgresql.org/docs/current/datatype-character.html
MSSQL | text  | Variable-length non-Unicode data in the code page of the server and with a maximum string length of 2^31-1 (2,147,483,647). When the server code page uses double-byte characters, the storage is still 2,147,483,647 bytes. Depending on the character string, the storage size may be less than 2,147,483,647 bytes. https://docs.microsoft.com/en-us/sql/t-sql/data-types/ntext-text-and-image-transact-sql?view=sql-server-ver15
MySQL | TEXT | 64K https://www.mysqltutorial.org/mysql-text/

driesvints:
Feel free to attempt a PR 👍

PR #794, opened 2021-05-26
[2.x] Changed the column type of profile_photo_path

rabin-io:
As the TEXT type on some DBs, creates an column with a limit of 2/4 GB.

This should increase compatibly with other DBs, with no impact on the migration it self. But we DO limit the column length, which is a good thing for this column use case. As a good practice URI should not be longer then 2K.

There is no mention of max URI length in the RFC3986, but there are limitation imposed by the software like the browsers them self.

|DB        |typeString                                                                           |typeTinyText |typeText     |typeMediumText|typeLongText |
|----------|-------------------------------------------------------------------------------------|-------------|-------------|--------------|-------------|
|MySQL     |varchar({$column->length})                                                           |tinytext     |text         |mediumtext    |longtext     |
|PostgreSQL|varchar({$column->length})                                                           |varchar(255) |text         |text          |text         |
|SQLite    |varchar                                                                              |text         |text         |text          |text         |
|MSSQL     |varchar({$column->length})                                                           |nvarchar(255)|nvarchar(max)|nvarchar(max) |nvarchar(max)|
|Informix  |varchar({$column->length}) : length < 255 <br/>lvarchar({$column->length}) : length >= 255| text         |text         |text          |text         |

ref: issue #789

### DIFF
--- a/database/migrations/2014_10_12_000000_create_users_table.php
+++ b/database/migrations/2014_10_12_000000_create_users_table.php
@@ -21,7 +21,7 @@ class CreateUsersTable extends Migration
             $table->string('password');
             $table->rememberToken();
             $table->foreignId('current_team_id')->nullable();
-            $table->text('profile_photo_path')->nullable();
+            $table->string('profile_photo_path', 2048)->nullable();
             $table->timestamps();
         });
     }


miclaus:
@rabin-io @taylorotwell 
wouldn't storing only the hash (and extension) of the profile photo suffice?

the way that it's setup at the moment, the profile photo path is saved for every entry, which is redundant. 

also if the path where profile pictures are stored were to change, entries would also have to be updated.

saving only the hash would only require at most `varchar(45)`, e.g. `{hashName}.jpeg` where `{hashName}` would have a length of 40 chars (because [hashName](https://github.com/laravel/framework/blob/b3329fad6809d69679075bee42b14d685e189217/src/Illuminate/Http/FileHelpers.php#L42)).

how about this:
```php
$table->string('profile_photo', 45)->nullable();
```
...plus all necessary changes resulting therefrom – i could do a PR if you'd want me to.

PR #798, opened 2021-06-01
[2.x] Allow Role's `$name` and `$description` to be translated

crynobone:
I'm not sure whether this should be considered as a breaking change. But this will avoid loading translation early in the request lifecycle and instead only require it to be translated when returning a response.

### DIFF
--- a/src/Role.php
+++ b/src/Role.php
@@ -71,8 +71,8 @@ class Role implements JsonSerializable
     {
         return [
             'key' => $this->key,
-            'name' => $this->name,
-            'description' => $this->description,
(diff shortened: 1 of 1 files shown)


driesvints:
Bit torn. I see what you're trying to do but it doesn't feels right to instantiate the app in a value object.

taylorotwell:
This could potentially lead to weird situations where the user is able to name roles but it happens to match a translated string in the language files, etc. I think best to just let the user override the method if they want.

taylorotwell:
Hmm, on second thought this change could be necesary.

PR #802, opened 2021-06-09
Improve inertia tests

reibengu:
Modest contribution to improve test cases so the end-user will easily catch mutations on published actions.


taylorotwell:

reibengu:
> Feel free to add to your own application.

I’m sorry but I disagree with you about this.
Why do you think that it doesn't make sense to assert `/reset-password` API is indeed changing the password?
Thought I was helping here :)

PR #804, opened 2021-06-13
Update InstallCommand.php

ninjaparade:
Similar to the issue I ran into with this https://github.com/laravel/breeze/pull/70/files

The current dependencies won't allow `spark-stripe` to be installed.


ninjaparade:
thanks Tays.

PR #809, opened 2021-06-21
Update Tailwindcss to latest version

fwartner:
Updated tailwindcss to the latest version.


driesvints:
The current version constrain already installs this version. Unless there's specific features we use from Tailwind v2.2 there's no reason to bump this.

PR #810, opened 2021-06-21
Tailwind JIT-Mode

sotten:
Tailwind offers a new just-in-time mode for Tailwind V2.1+. This feature is currently in preview, but I believe it is already productive in the current version. I suggest the use of Tailwind jit. It offers some very valuable advantages. See the further [documentation](https://tailwindcss.com/docs/just-in-time-mode)


driesvints:
Now *this* would require the version bump from https://github.com/laravel/jetstream/pull/809

taylorotwell:
See above and resubmit please.

PR #813, opened 2021-06-24
[2.x] Support both Alpine V2 and V3

calebporzio:
Problem scenario:
* Someone updates Jetstream to the latest version that supports Alpine V3
* They don't update their Alpine version to V3 (because package.json/app.js doesn't get automatically updated)
* Jetsream is now broken in the following two ways:
  * The top right menu doesn't work because it uses `@click.outside` which is new in V3
image
  * The little "Saved." indicator doesn't transition out because it uses `x-transition...` which is new in V3
image

Solution:
* Use `.away` instead of `.outside` - they are aliases and both exist in V2 and V3
* Include the legacy `x-show.transition...` as well as the new `x-transition...` helpers. Each will work on their respective versions without causing breakages.

After this PR, all Jetsream stubs will be upgrade safe and support any version of Alpine a project is using.

### DIFF
--- a/resources/views/components/action-message.blade.php
+++ b/resources/views/components/action-message.blade.php
@@ -2,7 +2,7 @@
 
 <div x-data="{ shown: false, timeout: null }"
     x-init="@this.on('{{ $on }}', () => { clearTimeout(timeout); shown = true; timeout = setTimeout(() => { shown = false }, 2000);  })"
-    x-show="shown"
+    x-show.transition.out.opacity.duration.1500ms="shown"
     x-transition:leave.opacity.duration.1500ms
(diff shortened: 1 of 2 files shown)


driesvints:
Thanks @calebporzio. We can probably move to v3 entirely in Jetstream v3 (master branch). Once this PR is merged and merged into master we can do the changes for v3 there.

calebporzio:
Ok, will PR - thanks @driesvints!

calebporzio:
Looking at master: looks like I don't have to.

driesvints:
@calebporzio I just merged 2.x into master 👍

PR #817, opened 2021-06-30
Update dropdown.blade.php

sotten:
Add `inline-flex` to the outermost div element. It ensures that the size of the element is adjusted to the child content.
Without `inline-flex`, I have the problem that the alpine `@click` function is also executed outside the `$trigger` area.

Bildschirmfoto von 2021-07-01 00-15-18

### DIFF
--- a/resources/views/components/dropdown.blade.php
+++ b/resources/views/components/dropdown.blade.php
@@ -25,7 +25,7 @@ switch ($width) {
 }
 @endphp
 
-<div class="relative" x-data="{ open: false }" @click.away="open = false" @close.stop="open = false">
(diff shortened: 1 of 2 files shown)


taylorotwell:
Do you plan to update the Inertia bootstrapping?

sotten:
I have no experience with Inertia. I have added a commit in the hope that it is correct.

timrspratt:
Adding this appears to result in this happening on the header
Screenshot 2021-07-01 at 21 36 45

driesvints:
@sotten ^

sotten:
I'm afraid I won't be able to review it until the week after next.

taylorotwell:
Reverted. If you want to fix something fix it in your own project. We are having too many breakages.

PR #823, opened 2021-07-08
add name to vue components that are useful to re-use

ArielMejiaDev:
## Description

it adds the name property to every vue component that is used in jetstream

## How it help the end user

it helps the IDE to autocomplete and import automatically the jetstream components to use it on any other form, page component.


taylorotwell:
Feel free to add this to your own application.

PR #825, opened 2021-07-12
Vue 3 Composition API

octoper:
This PR upgrades the Inertia version of Jetstream to use the Vue 3 composition API.

### DIFF
--- a/stubs/inertia/resources/js/Pages/Auth/ConfirmPassword.vue
+++ b/stubs/inertia/resources/js/Pages/Auth/ConfirmPassword.vue
@@ -26,6 +26,7 @@
 </template>
 
 <script>
+    import { useForm } from '@inertiajs/inertia-vue3'
     import JetAuthenticationCard from '@/Jetstream/AuthenticationCard'
     import JetAuthenticationCardLogo from '@/Jetstream/AuthenticationCardLogo'
     import JetButton from '@/Jetstream/Button'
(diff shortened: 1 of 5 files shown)


claudiodekker:
Hey @octoper!

Since Dries requested my opinion/review on this (partial implementation), I figured I'd leave this comment here as well so nobody's left out of the conversation: While I don't dislike the Composition API, I'm personally against these proposed changes for a couple of reasons:

1. Not everyone is familiar with or wants to use the Composition API
2. It doesn't really add value: The main value of the Composition API is sharing and/or re-using code, but there is barely any.
3. It increases the differences between both Vue versions of Jetstream, making maintaining them all the more difficult.

In my experience, some only upgrade to Vue 3 because it's (in their eyes) more "forwards compatible" or (in other words) not "soon to be legacy", while others upgrade to Vue 3 because some library/plugin they use is only Vue 3 compatible. There's also those whom create their app in Vue 3 because it was the obvious choice, and because everything they did so far more or less worked the same as in Vue 2 because of the Options API being a thing.

On the flipside, there's those who upgraded to Vue 3 specifically because of the Composition API and who would love to see things be as "modern" as possible, but even then I don't really see the added value: Is there really anything on these specific pages that you want to share and/or re-use within your codebase?

Another argument I can imagine, is that you'd simply want to enforce a 'common design pattern' all across your apps, where all pages are using the Composition API, and while I personally don't understand the value in doing so, I'd encourage you to convert those pages in that case: Most of the logic stays the same, so doing so really wouldn't take all that long, especially if you have some Composition API experience already anyway.

Finally, and while I intensely dislike this argument myself (although it is relevant): There is nothing that prevents you from still publishing a series of pages / package that you can auto-publish after Jetstream is done, that overwrites these pages with the Composition API variants.

With that said, if you do think I'm wrong or am missing some aspect, I'd love to have my opinion on this corrected, but as for right now this is why I don't think we should be accepting these changes.

Thank you for contributing!

octoper:
Hey @claudiodekker.
I understand everything that you said and I agree I just opened the PR to see if there is any value to move to the Composition API.

Thank you for your time!

driesvints:
@octoper gonna close this one then. Thanks @claudiodekker

PR #826, opened 2021-07-13
Fix Inertia Default Banner Component Does Not Show After Dismissing

jamesbhatta:
Right now dismissing the banner, it is not shown when newer message is flashed because the show property is set to false while dismissing.

This PR removes the show property and instead only depend upon the message property to toggle the banner. While dismissing just set ``this.$page.props.jetstream.flash.banner == null``

### DIFF
--- a/stubs/inertia/resources/js/Jetstream/Banner.vue
+++ b/stubs/inertia/resources/js/Jetstream/Banner.vue
@@ -1,6 +1,6 @@
 <template>
     <div>
-        <div :class="{ 'bg-indigo-500': style == 'success', 'bg-red-700': style == 'danger' }" v-if="show && message">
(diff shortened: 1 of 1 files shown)


claudiodekker (changes requested):
Could we instead use a watcher (or some other approach) that updates the `show` property? 

Right now, this PR is modifying props (albeit deeply nested), which isn't recommended in Vue.

taylorotwell:
Feel free to adjust the behavior of the banner in your own applications.

jamesbhatta:
For most of us, I think we expect to dismiss only the current banner notification not to disable all of the upcoming ones.

eskiesirius:
any fix on this issue?

jamesbhatta:
You can check the changes in commit and customize as you need.

PR #828, opened 2021-07-17
redirect by route name

rahulkhimsuriya:
- redirect to `dashboard` by route name.


driesvints:
This is a Vue component so this wouldn't work.

Issue #831, opened 2021-07-19
Use custom ConfirmPassword hook for Inertia CurrentUserController@destroy password validation 


Gimcrack:
Background:

- Jetstream (Inertia)
- User Provider: Active Directory via LdapRecord package

Problem:

I'm using a non-laravel user provider (Active Directory via LdapRecord package) and therefore need a custom hook for confirming passwords:

```
// AuthServiceProvider
public function boot() 
{
    $this->registerPolicies();

    Fortify::confirmPasswordsUsing(function($user, $password) {
        return Auth::validate([
            'mail' => $user->email,
            'password' => $password
        ]);
    });
}
```

This works great for confirming passwords, but when I try to delete an account the `@destroy` method on Interia's `CurrentUserController` does not use the custom hook to verify the password and the validation fails.

```
// Laravel\Jetstream\Http\Controllers\Inertia\CurrentUserController
public function destroy(Request $request, StatefulGuard $auth)
{
    // this does not use the custom hook confirmPasswordsUsing and the validation fails.
    $request->validate([
        'password' => 'required|string|password',
    ]);

   ...
}
```

I can get it to work if I modify the `@destroy` method. Code comments added for clarity.

```
// Laravel\Jetstream\Http\Controllers\Inertia\CurrentUserController
public function destroy(Request $request, StatefulGuard $auth)
{
    // check for presence of password in request, but don't check it yet
    $request->validate([
        'password' => 'required|string', 
    ]);

    // confirm password using ConfirmPassword action
    $confirmed = app(Laravel\Fortify\Actions\ConfirmPassword::class)(
        $auth, $request->user(), $request->input('password')
    );

   // throw validation exception on bad password
    if (! $confirmed)
        throw new ValidationException("The password is incorrect.");

    // continue to delete account ...
   ...
}
```

It seems like there should be some consistency to how passwords are confirmed, but I'm not sure if this is the best approach. Is there a way to hook into the request's password validation to make it use a different method? I'm happy to try my hand at a PR, but would probably need some assistance to make sure it follows the right conventions.

Thanks,
-Jeremy

Gimcrack:
I found another way to solve the issue that does not require editing framework code, but it still seems a little gross. 

I can substitute my own validator by passing it in via the `Validator@resolve` method.

```
// AppServiceProvider
    Validator::resolver( function($translator, $data, $rules, $messages, $customAttributes) {
[...]

PR #835, opened 2021-07-22
[2.x] Defer to Fortify for password confirmation

Gimcrack:
**Problem**: Password confirmation within Jetstream (Inertia) is not consistent. This breaks certain key features when using custom handling for confirming passwords e.g. when using a non-standard guard or user provider. 

- Password confirmation for enabling 2FA uses Fortify's `ConfirmPassword` action and can be customized by hooking into `confirmPasswordUsing`. This feature is working.
```
$confirmed = app(ConfirmPassword::class)(
    $this->guard, $request->user(), $request->input('password')
);
```
- **Password confirmation for account deletion** and **signing out other browser sessions** use Laravel's password validation rule and can't be customized (easily). Password validation fails and these features don't work.
```
$request->validate([
    'password' => 'required|string|password',
]);
```

**Solution**: Defer to Fortify for password confirmation during account deletion and signing out other browser sessions.

@driesvints I did see your comment on #831 and I'm not trying to be spammy with this. I did try asking around in the community about this and got zero feedback. Nevertheless, I do believe this is an actual bug (or incomplete feature) and more than a question about implementation. Accounts can't be deleted without hacking Laravel's password validation rule.

### DIFF
--- a/src/Http/Controllers/Inertia/CurrentUserController.php
+++ b/src/Http/Controllers/Inertia/CurrentUserController.php
@@ -9,6 +9,7 @@ use Illuminate\Support\Facades\Hash;
 use Illuminate\Validation\ValidationException;
 use Inertia\Inertia;
 use Laravel\Jetstream\Contracts\DeletesUsers;
+use Laravel\Fortify\Actions\ConfirmPassword;
 
 class CurrentUserController extends Controller
 {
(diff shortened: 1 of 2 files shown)


taylorotwell:
Seems fairly reasonable.

driesvints:
Thanks @Gimcrack

Gimcrack:
Happy to help. Thanks.

Issue #833, opened 2021-07-20
Jetstream::findRole() expecting a string even if role column is nullable


rinodrummer:
- Jetstream Version: 2.3.11
- Jetstream Stack: Livewire
- Uses Teams: yes
- Laravel Version: 8.50.0
- PHP Version: 7.4.11
- Database Driver & Version: MySQL 5.6

### Description:

In Jetstream, the `role` column of the pivot table for the relation between users and teams (`team_user`) can be nullable, but the method `Jetstream::findRole()` method (used in `HasTeams::teamRole() and in `HasTeams::hasTeamRole()`) requires a non-nullable string argument and returns this error:

> Argument 1 passed to Laravel\Jetstream\Jetstream::findRole() must be of the type string, null given, called in ..\vendor\laravel\jetstream\src\HasTeams.php on line 145

The `Jetstream::findRole()` method should default to an empty string instead or `HasTeams::teamRole()`/`HasTeams::hasTeamRole()` should have a fallback value if the membership role is not set.

What I'm trying to do go is to get the user role in a team using `HasTeam::teamRole()` and I'm iterating all the teams of a user (retrieved using `HasTeams::allTeams()`) and already excluded the owned teams.

### Steps To Reproduce:

1. Create a user;
2. Create a team **not owned** by the user;
3. Attach the team to the user without setting a role;
4. Call `$user->teamRole($team)` or `$user->hasTeamRole($team, null)`;
5. Error occurs.

**N.B.:** I'm sorry if my issue may be not well presented or not so clear, but I'm not that expert in issuing bugs or submitting PR. I thought to open an issue before because the approaches (or the whole thing) I suggested may have to be discussed.

rinodrummer:
i solved using a the `rescue()` helper to prevent the exception to raise.

```php
$role = rescue(() => $user->teamRole($team));
```

PR #836, opened 2021-07-22
Missing localization for owner role name

rinodrummer:
Working on with my issue (#833), I've seen that `OwnerRole` was missing of the localization.

### DIFF
--- a/src/OwnerRole.php
+++ b/src/OwnerRole.php
@@ -11,6 +11,6 @@ class OwnerRole extends Role
      */
     public function __construct()
     {
-        parent::__construct('owner', 'Owner', ['*']);
+        parent::__construct('owner', __('Owner'), ['*']);
     }
 }


crynobone:
Initializing translation during construct will load translation way before middleware (where we normally set locale based on user session etc), this can cause unexpected results in some scenarios.

PR #798 already cover this requirement and only use translation when we want to use it on the views.

rinodrummer:
So , thanks to the definition in `JeststreamServiceProvider`, other roles have a translated name by default, like this:

```php
Jetstream::role('admin', __('Administrator'), [
    'create',
    'read',
    'update',
    'delete',
])->description(__('Administrator users can perform any action.'));
```

For example, if execute this code with Italian language set as default:

```php
{{ $role->name }} 
```

It prints "Amministratore" instead of "Administrator".

How can it be achieved with `OwnerRole`?

Issue #838, opened 2021-07-24
Install fails with fresh, bone-stock Laravel instance.


unrivaledcreations:
- Jetstream Version: `^1.3|^2.0` (via `composer.json`)
- Jetstream Stack: Inertia / (N/A)
- Uses Teams: yes / no (N/A)
- Laravel Version: 8.51.0
- PHP Version: 8.0.8
- Database Driver & Version: (N/A)

### Description:

Getting a Composer error on install, using the command `composer require laravel/jetstream`:

```
Problem 1
    - laravel/jetstream[v2.3.0, ..., 2.x-dev] require league/commonmark ^1.3 -> found league/commonmark[1.3.0, ..., 1.6.x-dev] but the package is fixed to 2.0.0 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
    - Root composer.json requires laravel/jetstream ^2.3 -> satisfiable by laravel/jetstream[v2.3.0, ..., 2.x-dev].
```

The bone-stock Laravel installation uses Jetstream v2 because `vendor/laravel/framework/src/Illuminate/Mail/composer.json` contains `"league/commonmark": "^1.3|^2.0"`. However, Jetstream's `composer.json` requires `"league/commonmark"` `"^1.3"`. This breaks the installation process because the core Laravel framework already installed v2.x when Jetstream seems to insist on v1.3 (of the `league/commonmark` dependency).

### Steps To Reproduce:

```sh
laravel new test
cd test
composer require laravel/jetstream
```

The single command, `laravel new --jet --stack=inertia test`, also fails for the same reason.

### Workaround:

Manually adding `"league/commonmark": "^1.3",` to the root `composer.json` project file before `composer require laravel/jetstream` seems to work around the problem.

jonorozcoc:
I confirm the issue with

- PHP: 8.0.6
- Stack: Livewire

It happens on fresh install and the workaround mentioned by @unrivaledcreations works

zamarproject:
I had the same issue.
I tried your suggestion but didn´t work.
I placed "league/commonmark": "^1.3" in the root composer.json project file:

    "require": {
        "php": "^7.3|^8.0",
        "fideloper/proxy": "^4.4",
        "fruitcake/laravel-cors": "^2.0",
        "guzzlehttp/guzzle": "^7.0.1",
        "league/commonmark": "^1.3",   <-- right there...        
        "laravel/framework": "^8.40",
        "laravel/tinker": "^2.5"
[...]

PR #840, opened 2021-07-25
[2.x] Fix commonmark install issue

driesvints:
Fixes https://github.com/laravel/jetstream/issues/838


RomanSarvarov:
Are you sure that it fixes it?
Jetstream 2.3.13 calling a class that commonmark 2.0 does not have.

Issue #844, opened 2021-07-28
Class "League\CommonMark\Environment" not found


caiokawasaki:
- Jetstream Version: 2.3.13
- Jetstream Stack: Livewire
- Uses Teams: no
- Laravel Version: 8.52.0
- PHP Version: 8.0.8 (Using Sail)
- Database Driver & Version: None

### Description:

Yesterday I decided to run away from the Laravel/Vue stack and test Livewire, I just created a simple CRUD, and after making some changes, I noticed that on the registration screen the `Terms of Service` and `Privacy Policy` pages weren't rendering. I even tried it with a clean install...

On both pages the following error is displayed: [https://flareapp.io/share/87ng2Z05#F48](https://flareapp.io/share/87ng2Z05#F48)

### Steps To Reproduce:

```sh
curl -s https://laravel.build/example-app | bash
```

```sh
sail composer require laravel/jetstream
sail artisan jetstream:install livewire
sail npm install
sail npm run dev
sail artisan migrate
```

Then enable `Features::termsAndPrivacyPolicy()` inside `config/jetstream.php`.

Both pages displays the error:

- [http://localhost/terms-of-service](http://localhost/terms-of-service)
- [http://localhost/privacy-policy](http://localhost/privacy-policy)

### Edit

Looks like a `PR` related to this issue is open: https://github.com/laravel/jetstream/pull/843

driesvints:
Tagged v2.3.14 with a fix.

PR #845, opened 2021-07-28
[2.x] Support both CommonMark v1 & v2

driesvints:
Use `Str::markdown` to support both CommonMark v1 and v2.

Fixes https://github.com/laravel/jetstream/issues/844

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "ext-json": "*",
         "illuminate/support": "^8.0",
         "jenssegers/agent": "^2.6",
-        "league/commonmark": "^2.0",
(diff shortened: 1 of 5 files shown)


RomanSarvarov:
`Str::markdown` is good solution!
So now CommonMark package is unused in JetStream, right?

driesvints:
@RomanSarvarov CommonMark is used behind the scenes.

RomanSarvarov:
> @RomanSarvarov CommonMark is used behind the scenes.

All I find is that 4 inertia/livewire controllers.
Now Common mark is used by laravel framework only, I guess.

driesvints:
Yeah we could remove it now.

PR #847, opened 2021-07-29
added check for if developer has disabled TwoFactor and update Profil…

msonowal:
This will make the Two-factor stubs tests generated for both of stacks respect the configurations as per developer toggles


taylorotwell:
I still think if you're removing that feature just remove the tests... this adds a lot of noise.

PR #855, opened 2021-08-14
Fix typo

wassim:


driesvints:
Current test names are already correct?

PR #858, opened 2021-08-16
Add link button to banner component

bilfeldt:
This PR makes it possible to easily add a link button to the banner component. This is useful for example to redirecting unsubscribed users to the billing page.

## Example

See "Signup" button that has been added to the top right corner.

Screenshot 2021-08-16 at 10 07 22


driesvints:
The button atm seems different from the "save" button in the example. I'd make the padding between the button and the edge of the bar the same if possible.

bilfeldt:
> The button atm seems different from the "save" button in the example. I'd make the padding between the button and the edge of the bar the same if possible.

Now using same styling as the `secondary-button` component with the following result:
Screenshot 2021-08-16 at 12 21 19

taylorotwell:

PR #864, opened 2021-08-22
fix: teams tests stubs

faissaloux:
This PR checks if teams feature is activated or not before running teams tests.
if teams feature is not activated tests will be skipped

Same for teams invitations


driesvints:
Delete the tests you don't need.

Issue #867, opened 2021-08-30
Making Vue components more Typescript friendly


studnitz:
Hi, I'm not sure if here is the right place to open this, but this repo doesn't have Github discussions enabled, where I would post this instead.

#829 made the development with Vite out of the box better.

As I'm finding myself using Typescript more and more as Vue 3 came out I had now several projects where I converted the Jetstream components to a Typescript version.

The most basic conversion is done by [adding a no-op wrapper around `export default { ... }` called `defineComponent`](https://v3.vuejs.org/guide/typescript-support.html#defining-vue-components), which just is there to tell Typescript that the enclosed object indeed is a component.

So the change is mainly:

```diff
+ import { defineComponent } from 'vue';

- export default {
+ export default defineComponent({
  ...
  // component options
  ...
- }
+ })
```

So before opening a PR with these changes made I wanted to ask, whether this would be considered to be merged. As I said, this change would not affect users who do not use Typescript, it would just make life easier for users, who want to add Typescript to their Jetstream project and get proper typings.

If that is not planned, I save some time creating the PR. :)

driesvints:
I'll check with @taylorotwell 👍

taylorotwell:
If that is the only change I don't have a problem with it. If there are other, larger changes I would need a list of them.

studnitz:
This would be mainly it. Everything else is mainly additive. For example, adding a `tsconfig.json` makes only sense, when you are really adding Typescript as a dependency to the project.

Another additional thing, that could be added with a later PR would be adding a `jetstream.d.ts` file, which could add proper typings for the additional props in `$page.props.jetstream` defined in the [ShareInertiaData middleware](https://github.com/laravel/jetstream/blob/b58d6fb4702763860809e844884073f7526f0ff2/src/Http/Middleware/ShareInertiaData.php). But that is not a must-have.

driesvints:
Thanks @studnitz. I think the next step would be a PR for the minimal changes you mentioned. Feel free to send that in 👍

jrrmcalcio:
If this is helpful to someone, I added a types.d.ts file with the following content for jetstream:

```
declare module "@inertiajs/inertia" {
    export interface PageProps {
        jetstream: {
            canCreateTeams: boolean;
            canManageTwoFactorAuthentication: boolean;
            canUpdatePassword: boolean;
            canUpdateProfileInformation: boolean;
            flash: { [key: string]: unknown };
            hasAccountDeletionFeatures: boolean;
[...]

PR #869, opened 2021-08-31
[2.x] Use defineComponent in vue stubs for easier typescript setup

studnitz:
As discussed in #867, this PR adds a [`defineComponent()`](https://v3.vuejs.org/api/global-api.html#definecomponent) around the `export default { ... }` component options of all Vue components. 

`defineComponent` is just a function that returns the object passed into it. But it makes life easier for people, who want to use Typescript in their Jetstream project.


driesvints:
Thanks @studnitz

PR #883, opened 2021-09-16
Revert "fix cors policy errors with inertia stack (#797)"

### LINKED ISSUE #850: BrowserSync Error on fresh installation
Hey everyone,
this is the first issue I'm submitting, so if I'm doing anything wrong, please let me know - I am more than happy to learn and adapt.

- Jetstream Version: 2.3.14
- Jetstream Stack: Inertia
- Uses Teams: no
- Laravel Version: 8.53.0
- PHP Version: 7.4.6
- Database Driver & Version: MySQL

### Description:
Using the regular [laravel installer](https://laravel.com/docs/8.x/installation#the-laravel-installer) to setup a fresh version of laravel or Sail an error will be thrown on the client when opening up the app (not tested with Valet):

```
GET http://localhost:3000/browser-sync/browser-sync-client.js net::ERR_CONNECTION_REFUSED
```

I believe this was introduced by PR: https://github.com/laravel/jetstream/pull/797 , where the import of the script `http://localhost:3000/browser-sync/browser-sync-client.js` was hardcoded for whenever the app is running locally into app.blade.php, meaning that the client will try to load this script in local env even if there is no sync server - and having no sync server running seems to be the default even if you go with `npm run watch`.

Another thing that could be problematic with hardcoding the URL of the browser-sync-client.js : It will also fail if there already is an application that is running on this port, which shouldn't be that unusual considering the port number.

I don't know if this actually has great impact to an end user as the issue can be fixed somewhat quickly. But I can imagine that it could lead to some negative emotional reaction, stemming from the situation that you're doing a fresh install and the first thing you're seeing is an error popping up in your console. For newcomers this might be frustrating.

#### Possible Solutions I can think of at the moment
I am not really sure about the behavior that is actually desired. That's why I didn't open up a PR, but I am happy to help, working on a solution.

##### Add browserSync to webpack.mix.js
One could add a few lines to webpack.mix.js, making sure there is a browserSync in non production environments:

```javascript
if (!mix.inProduction()) {
    mix.browserSync(process.env.APP_URL);
}
```

This works when running `npm run watch` but not with `npm run dev`.

##### Revert https://github.com/laravel/jetstream/pull/797 and add it to the docs
One could also remove the lines added by https://github.com/laravel/jetstream/pull/797 and add a section to the docs, explaining how to resolve the issue for Valet.

##### Keep https://github.com/laravel/jetstream/pull/797 and add step to installation docs
One could also keep the lines and add a section to the installation of Jetstream Inertia, that explain what is happening with browserSync, how to get it to work and why it won't work on `npm run dev`. 

### Steps To Reproduce:
- Simply create a fresh laravel app with Jetstream Inertia using either `laravel new example-app -jet` or create a fresh laravel app using sail and install Jetstream Inertia.
- Then open up your browser and visit the site. The console should throw an error telling that connection was refused to load browser-sync.

DSpeichert:
Attempting to load browser-sync-client.js without it being used in
Jetstream by default results in blocked JS execution until timeout.

Fixes #850

### DIFF
--- a/stubs/inertia/resources/views/app.blade.php
+++ b/stubs/inertia/resources/views/app.blade.php
@@ -18,9 +18,5 @@
     </head>
     <body class="font-sans antialiased">
         @inertia
-
-        @env ('local')
-            <script src="http://localhost:3000/browser-sync/browser-sync-client.js"></script>
-        @endenv
     </body>
 </html>


taylorotwell:
It is to support Mix's `npm run hot` out of the box for browser reloading.

DSpeichert:
@taylorotwell But `npm run hot` does not actually start the browsersync server component (nothing is listening on port 3000).
`browser-sync` is not in package.json or package-lock.json.

PR #888, opened 2021-09-23
Allow team invitation table to use connection out of the box...

telkins:
A project I work on uses Spatie's multi-tenancy package with a landlord database and multiple tenant databases. In order to guarantee that the correct connection is used, I need to change it in this file (`Rule::unique('my_connection.team_invitations')`).

This change should allow the correct connection/table in all (?) cases without having to change anything here.

I'm not convinced this is the best way to resolve the problem.  I'd love feedback or, if it's not clear, any questions about why this might be necessary.

Thanks.  🤓

### DIFF
--- a/stubs/app/Actions/Jetstream/InviteTeamMember.php
+++ b/stubs/app/Actions/Jetstream/InviteTeamMember.php
@@ -67,8 +67,11 @@ class InviteTeamMember implements InvitesTeamMembers
      */
     protected function rules($team)
     {
+        $teamInvitation = new (Jetstream::teamInvitationModel());
+        $connectionTable = $teamInvitation->getConnectionName() . '.' . $teamInvitation->getTable();
+
         return array_filter([
-            'email' => ['required', 'email', Rule::unique('team_invitations')->where(function ($query) use ($team) {
+            'email' => ['required', 'email', Rule::unique($connectionTable)->where(function ($query) use ($team) {
                 $query->where('team_id', $team->id);
             })],
             'role' => Jetstream::hasRoles()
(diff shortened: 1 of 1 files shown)


stubs/app/Actions/Jetstream/InviteTeamMember.php
```diff
@@ -67,8 +67,12 @@ protected function validate($team, string $email, ?string $role)
      */
     protected function rules($team)
     {
+        $class = Jetstream::teamInvitationModel();
+        $teamInvitation = $class::make();
```
taylorotwell, line 71: Where is this `make` method defined?

telkins: I could be wrong, but I thought that it was on `Illuminate\Database\Eloquent\Model`...although I'm looking now and I can't find it. 😳 Makes me wonder how it's working.

I had originally simply instantiated the model via `new`, but got an error that I couldn't reproduce locally, so I changed it to `Model::make()`....which doesn't appear to be where I thought it was, and yet somehow works.

I'll change it to `new $class`.... 🤓

telkins: It's odd....locally, with `laravel/framework v8.61.0`, I run the `InviteTeamMemberTest` test, dumping `$teamInvitation` after using `make()` and get the following:
```bash
App\Models\TeamInvitation^ {#3791
  #fillable: array:2 [
    0 => "email"
    1 => "role"
  ]
...
```

Not sure how/why it's working.  After a quick search, I can't find any `make()` method in the API docs or in my own copy of the framework source.  🤔

Anyway...it's changed and still works locally.  Let me know if you have any other questions/concerns. 🤓

taylorotwell:
Since this is a stub you can just change it to whatever you want in your own application.

telkins:
> Since this is a stub you can just change it to whatever you want in your own application.

Fair enough...but just to make it clear, it did take me a while to track down where the issue was. Changing this in the stub "at the source" would eliminate the problem for everyone.

Anyway, thanks for your time.  Can't win 'em all....  🤓

PR #892, opened 2021-09-28
Use custom TeamInvitation model in TeamInvitationController

### LINKED ISSUE #890: TeamInvitationController not using custom TeamInvitation model
- Jetstream Version: v2.4.0
- Jetstream Stack: Inertia
- Uses Teams: yes
- Laravel Version: v8.61.0
- PHP Version: v7.3.28
- Database Driver & Version: mysql

### Description:
The [accept](https://github.com/laravel/jetstream/blob/2.x/src/Http/Controllers/TeamInvitationController.php#L21) and [destroy](https://github.com/laravel/jetstream/blob/2.x/src/Http/Controllers/TeamInvitationController.php#L44) methods in the TeamInvitationController use route model binding to inject the requested team invitation model. However, those instances are always of class `\Laravel\Jetstream\TeamInvitation`. The custom model class, e.g. `\App\Models\TeamInvitation` is not used.

Our application creates audit logs when models are being changed using an [external package](https://github.com/owen-it/laravel-auditing). Team invitations are audited as well so `\App\Models\TeamInvitation` implements interface and trait provided by the external package. As the `TeamInvitationController` retrieves the wrong classes the "magic" will not happen when the model gets deleted.

I am more than happy to send a pull request. But before I do so what are the chances of that PR to be merged? I provided a description for two of many possible solutions below. Do you think that this is a breaking change?

### Steps To Reproduce:

1. Open the team settings
2. Invite a user to your team via email
3. Add `dd(get_class($invitation));` [here](https://github.com/laravel/jetstream/blob/2.x/src/Http/Controllers/TeamInvitationController.php#L49)
4. The output will be `\Laravel\Jetstream\TeamInvitation`

### Solution 1
Override `resolveRouteBinding` to use `Jetstream::teamInvitationModel()` in the base [TeamInvitation](https://github.com/laravel/jetstream/blob/2.x/src/TeamInvitation.php) model. 

**Downside**: might not be fully backward compatible when projects created a custom team invitation model not inheriting from the base Jetstream [TeamInvitation](https://github.com/laravel/jetstream/blob/2.x/src/TeamInvitation.php) model.

### Solution 2
Remove the the type-hint from the controller methods and query manually, see #651.

**Downside**: again not fully sure about backward compatibility when projects inherit from Jetstream base controller and override the methods mentioned before. All of sudden the base class would not have a type-hint anymore.

ManuelLeiner:
### Description
Use the custom `TeamInvitation` class in the `accept` and `destroy` methods of the `TeamInvitationController` to retrieve the invitation model.

### Motivation
Fixes #890. 

### Solution
Removes the type-hint from the controller methods. Therefore the model will not be resolved by the framework and the invitation id will be passed to the controller. Uses the invitation id and the Jetstream helper to query and retrieve an instance of the custom `TeamInvitation`. Uses `firstOrFail` as the route model binding responds with a `ModelNotFoundException` as well when an invalid id is passed.

### DIFF
--- a/src/Http/Controllers/TeamInvitationController.php
+++ b/src/Http/Controllers/TeamInvitationController.php
@@ -7,7 +7,7 @@ use Illuminate\Http\Request;
 use Illuminate\Routing\Controller;
 use Illuminate\Support\Facades\Gate;
 use Laravel\Jetstream\Contracts\AddsTeamMembers;
-use Laravel\Jetstream\TeamInvitation;
+use Laravel\Jetstream\Jetstream;
 
 class TeamInvitationController extends Controller
 {
@@ -15,11 +15,15 @@ class TeamInvitationController extends Controller
      * Accept a team invitation.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @param  \Laravel\Jetstream\TeamInvitation  $invitation
+     * @param  int  $invitationId
      * @return \Illuminate\Http\RedirectResponse
      */
-    public function accept(Request $request, TeamInvitation $invitation)
+    public function accept(Request $request, $invitationId)
     {
+        $model = Jetstream::teamInvitationModel();
+
+        $invitation = $model::whereKey($invitationId)->firstOrFail();
+
         app(AddsTeamMembers::class)->add(
             $invitation->team->owner,
             $invitation->team,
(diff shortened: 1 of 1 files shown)


driesvints (approved):
I also think this is probably the best solution.

ManuelLeiner:
@driesvints I actually think that this PR might have been merged to early. 

As I mentioned in #890 there might be a potential breaking change as soon as someone extends from the Jetstream TeamInvitationController and overrides one of the two methods. 

image

I quickly added an [example controller](https://github.com/ManuelLeiner/jetstream/blob/patch-team-invitation-model/src/Http/Controllers/CustomTeamInvitationController.php).

What do you think? Might solution 1 (route model resolving) be safer? Should we roll back and add the change to the next major release? Or am I overthinking?

driesvints:
@ManuelLeiner let's tackle that when someone actually reports an issue there.

freekmurze:
@freekmurze I got bitten by this. It's a breaking change indeed. Luckily I did have test to notify me of this change.

driesvints:
@freekmurze sorry about that. We reverted this PR.

@ManuelLeiner it's probably best that we retarget master for this.

ManuelLeiner:
> @driesvints I got bitten by this. It's a breaking change indeed. Luckily I did have tests to notify me of this change.
> 
> For me this is fine, but I guess for people that didn't write tests for their custom endpoint, this will silently break in their production environments.

> @freekmurze sorry about that. We reverted this PR.
> 
> @ManuelLeiner it's probably best that we retarget master for this.

Hey @driesvints, @freekmurze,

sorry for the inconveniences. 

Of course the next major release is perfectly fine for me/us. Anything I can do? New PR based on master?

driesvints:
> New PR based on master?

Yeah probably best

PR #893, opened 2021-09-29
Add "required" attribute to terms and privacy policy checkbox

LucaRed:
You must agree to the terms of service and the privacy policy during registration, if the feature is active. This is already validated, but only server-side.

This commit adds the `required` attribute to the checkbox, in order to take advantage of the native browser validation (just like with other required fields in the same form).


taylorotwell:
It's fine.

PR #906, opened 2021-11-01
simplify updateProfilePhoto function

miclaus:
there doesn't seem to be a reason to use tap for the `updateProfilePhoto` function, is there?

`@return void` in the updateProfilePhoto function description indicates that there isn't a return value. otherwise the function description should be updated.

in the only [usage of the updateProfilePhoto function](https://github.com/laravel/jetstream/blob/9663d7eb4e58ec86f105efb07f712fb81e6c92f2/stubs/app/Actions/Fortify/UpdateUserProfileInformation.php#L28) the return value of the function isn't used.

if devs currently rely on the function to return a value despite the function description, this PR might break their code.

### DIFF
--- a/src/HasProfilePhoto.php
+++ b/src/HasProfilePhoto.php
@@ -16,17 +16,15 @@ trait HasProfilePhoto
      */
     public function updateProfilePhoto(UploadedFile $photo)
     {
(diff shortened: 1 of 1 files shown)


miclaus:
ah okay, i see what you did there with `$previous`

PR #910, opened 2021-11-26
Allow personal teams to be optional

skuIIs:
The PR makes personal teams still default on a new install but has the option of disabling. This will allow a broader use of this package where registering a user doesn't create a personal team. This behavior is controlled by the `config/jetstream.php` file:

```PHP
Features::teams(['personal' => true, 'invitations' => true]),
```

The default behavior on a new install is unchanged. When enabled, personal teams still cannot be deleted. When disabled, no personal team is created at the time of user registration and allows the user to not be a member of any team. Team creation is still controlled by the `create` team gate in both cases. Both inertia and livewire templates have been updated and all tests pass.

The installer should decide during activation of the teams feature if this option should be enabled or disabled.

### DIFF
--- a/config/jetstream.php
+++ b/config/jetstream.php
@@ -45,7 +45,7 @@ return [
         // Features::termsAndPrivacyPolicy(),
         // Features::profilePhotos(),
         // Features::api(),
-        // Features::teams(['invitations' => true]),
+        // Features::teams(['personal' => true, 'invitations' => true]),
         Features::accountDeletion(),
     ],
 
--- a/database/factories/TeamFactory.php
+++ b/database/factories/TeamFactory.php
@@ -5,6 +5,7 @@ namespace Database\Factories;
 use App\Models\Team;
 use App\Models\User;
 use Illuminate\Database\Eloquent\Factories\Factory;
+use Laravel\Jetstream\Features;
 
 class TeamFactory extends Factory
 {
@@ -25,7 +26,7 @@ class TeamFactory extends Factory
         return [
             'name' => $this->faker->unique()->company(),
             'user_id' => User::factory(),
-            'personal_team' => true,
+            'personal_team' => Features::createsPersonalTeam(),
         ];
     }
 }
--- a/database/factories/UserFactory.php
+++ b/database/factories/UserFactory.php
@@ -61,7 +61,7 @@ class UserFactory extends Factory
         return $this->has(
             Team::factory()
                 ->state(function (array $attributes, User $user) {
-                    return ['name' => $user->name.'\'s Team', 'user_id' => $user->id, 'personal_team' => true];
+                    return ['name' => $user->name.'\'s Team', 'user_id' => $user->id, 'personal_team' => Features::createsPersonalTeam()];
                 }),
             'ownedTeams'
         );
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -447,7 +447,7 @@ EOF;
         $this->callSilent('vendor:publish', ['--tag' => 'jetstream-team-migrations', '--force' => true]);
 
         // Configuration...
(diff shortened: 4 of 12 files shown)


taylorotwell:

lostdesign:
@taylorotwell the PR to me seemed high valuable, bringing in a config options just like you'd expect in fortify. 

Despite that the test could be extended rather then skipping if there is no personal team desired, it should still be a first party solution and not a community package. 

Disabling a personal team isn't much of a change code wise, most that was done was replacing wether the team is desired over a hardcoded true.

I'd love if you could reconsider checking this PR. If necessary I am more than willing to help OP to write the tests for the non-personal-team option.

kyranb:
Just going to chime in and say too, across a variety of SaaS apps, this is the most common use case. Where someone is a team owner and wants to invite a 'teammate', without creating their own team/tenant at the same time. 

If this decision can be re-considered it would make Jetstream much more versatile :)

alesf:
It's kind of restricting if you can't disable personal teams.
Not every app fits in this flow of managing teams.
I really miss this option for when I want only admins to have control over team management and creating user roles.

RhysLees:
I think this needs to be reconsidered. I have been working on 5 or 6 websites now that use jetstream and only one of them have required a personal team, while the rest of them only needed to have teams optionally. in the end, for them, we ended up just disabling teams and adding in our own integration but it would be 1000x better.

MCKLtech:
Chiming in as others have said, personal teams make no sense and I'd welcome an explanation as to why they are mandatory. It is a poor design decision that has now got out hand? 

I've never come across a system that works in this manner. If I create a team and ask people to join, they join my team directly.

skuIIs:
@taylorotwell if this PR isn't going to be merged then can we get this feature in the near future?

RhysLees:
> @taylorotwell if this PR isn't going to be merged then can we get this feature in the near future?

Probably not since his response is a cookie-cutter template for the laravel/laravel repo. Just a straight-up close with some obscure reason 'To preserve our ability to adequately maintain the framework, we need to be very careful regarding the amount of code we include.'

The PR isn't even that much code added.

lnfel:
I just implemented similar case by changing Fortify actions provided by Jetstream, in my case users can only be created through a valid invitation.
Helpful article: https://devlan.io/disabling-personal-teams-in-laravel-8-and-jetstream-1fd083593e08/

I highly think that having Personal team to be optional is needed as explained by top comments.

```php
// in App\Actions\Fortify\CreateNewUser

use App\Models\TeamInvitation;
use Laravel\Jetstream\Contracts\AddsTeamMembers;

class CreateNewUser implements CreatesNewUsers
{
    public function create(array $input)
    {
        Validator::make($input, [
            'name' => ['required', 'string', 'max:255'],
            'email' => ['required', 'string', 'email', 'max:255', 'unique:users', 'exists:team_invitations'],
            'password' => $this->passwordRules(),
            'terms' => Jetstream::hasTermsAndPrivacyPolicyFeature() ? ['required', 'accepted'] : '',
        ],
        [
            'email.exists' => 'Can only register using invited email.',
        ])->validate();
    
        return DB::transaction(function () use ($input) {
            return tap(User::create([
                'name' => $input['name'],
                'email' => $input['email'],
                'password' => Hash::make($input['password']),
            ]), function (User $user) {
                // Disable creating personal team for each user
                //$this->createTeam($user);
    
                $invitation = TeamInvitation::where('email', $user->email)->first();
                $this->assignTeam($user, $invitation);
            });
        });
    }
    
    protected function assignTeam(User $user, TeamInvitation $invitation)
    {
        app(AddsTeamMembers::class)->add(
            $invitation->team->owner,
            $invitation->team,
            $invitation->email,
            $invitation->role
        );
    
        $user->current_team_id = $invitation->team_id;
        $user->save();
    }
}
```

a2sc:
The link to the article mentioned above has moved to: https://npratley.net/disabling-personal-teams-in-laravel-8-and-jetstream-1fd083593e08/

PR #911, opened 2021-11-30
Use KeyboardEvent `key` attribute for Escape

LeoColomb:
Use KeyboardEvent `key` attribute for Escape instead of deprecated `keyCode`.

See https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode


emargareten:
@LeoColomb PR in breeze as well?

LeoColomb:
@emargareten No problem 😃 https://github.com/laravel/breeze/pull/108

PR #912, opened 2021-11-30
Replaced hardcoded password validation message on LogoutOtherBrowserS…

evandrolacerda:
Since threre is no way to customize the LogoutOtherBrowserSessionsForm it would be great if the validations messages use the lang files of the application.


taylorotwell:
It is already using the JSON file localization feature - so it should already be customizable? Note the call to `__`.

https://laravel.com/docs/8.x/localization#using-translation-strings-as-keys

evandrolacerda:
Thanks for your reply.
Even thought my project has the json translation file, this string is not being translated, I should have missed sometihing, gonna take a good look into the documentation again.
Thanks a lot.

PR #915, opened 2021-12-03
Change stub from http://localhost:3000 to the config APP_URL

victoryoalli:
When working on a custom domain name like when using Valet is a better option to have only one place to change the application URL.

### DIFF
--- a/stubs/inertia/resources/views/app.blade.php
+++ b/stubs/inertia/resources/views/app.blade.php
@@ -20,7 +20,7 @@
         @inertia
 
         @env ('local')
-            <script src="http://localhost:3000/browser-sync/browser-sync-client.js"></script>
+            <script src="{{ config('app.url', 'http://localhost') }}:3000/browser-sync/browser-sync-client.js"></script>
         @endenv
     </body>
 </html>


taylorotwell:
APP_URL isn't relevant to browser sync?

victoryoalli:
Sorry, I missed the most important part, explaining why.

When developing there is two error that I get often when I start a new project with Inertia.

1. `http://localhost:3000/browser-sync/browser-sync-client.js net::ERR_CONNECTION_REFUSED`

2. `https://superapp.test:3000/browser-sync/socket.io/?EIO=3&transport=polling&t=NsBKpSo net::ERR_SSL_PROTOCOL_ERROR`

I wrote a solution for the problem in the following post 
[Laravel Inertia using Valete secure and Browser sync error](https://victoryoalli.me/browser-sync-using-https-and-valet-domain)

But in summary, the first error is solved by installing browser-sync and adding some browser-sync statements to webpack.mix.js configuration. But the second error depends on the domain name and https scheme. This is when APP_URL is relevant for hot reload.

What do you think? or can you guide in a better solution?

PR #920, opened 2021-12-09
Updates Tailwind on Installation to v3.0.0

RhysLees:
Updates:

@tailwindcss/forms to ^0.4.0
@tailwindcss/typography to ^0.5.0
tailwindcss to ^3.0.0
Replaced flex-grow-* and flex-shrink-* with grow-* and shrink-*

PR #926, opened 2021-12-14
Remove JIT mode from tailwind config

geisi:
Since Jetstream is using TailwindCSS 3.x (#920) there aren't other modes than Just in Time supported anymore. 
The JIT mode is used by default.
So the mode:'jit' config entry can completely be  removed.
[Upgrade Guide for reference](https://tailwindcss.com/docs/upgrade-guide#migrating-to-the-jit-engine)

The config files should be clean and should not contain unnecessary entries.


RhysLees:
Completly missed this, my bad

PR #930, opened 2021-12-23
Fixed a button spacing issue in dialog modal footer.

ravibpatel:
Fixed a button spacing issue in dialog modal footer.

### Before

Before fix

### After

After fix

### DIFF
--- a/resources/views/components/dialog-modal.blade.php
+++ b/resources/views/components/dialog-modal.blade.php
@@ -11,7 +11,7 @@
         </div>
     </div>
 
-    <div class="px-6 py-4 bg-gray-100 text-right">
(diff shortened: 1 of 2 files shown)


taylorotwell:
Can you share before / after on desktop as well?

ravibpatel:
Sure, You can see it in below screenshots.

### Before 

Before fix

### After

After fix

ravibpatel:
@taylorotwell Do you want this change in confirmation-modal too? This issue can happen on that modal too if developer using it use big button name like this. Also, I notice there was a little bit of change in spacing between buttons so It should be good idea to increase the margin from ml-2 to ml-3 to match the spacing before this modification. Let me know if you want this change so I can open a pull request.

PR #931, opened 2021-12-24
Support RTL layout [Livewire Stack]

mohamedsabil83:
By adding `dir="rtl"` to the `html`, the layout will shown correctly out-of-the-box.

### DIFF
--- a/resources/views/components/banner.blade.php
+++ b/resources/views/components/banner.blade.php
@@ -26,10 +26,10 @@
                     </svg>
                 </span>
 
-                <p class="ml-3 font-medium text-sm text-white truncate" x-text="message"></p>
+                <p class="ml-3 rtl:ml-0 rtl:mr-3 font-medium text-sm text-white truncate" x-text="message"></p>
             </div>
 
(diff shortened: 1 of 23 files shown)


taylorotwell:
Hey there! Are you able to fork Jetstream and maintain these changes? I'm not sure I can maintain this styling going forward as I'm not familiar with this kind of web development.

Thanks.

mohamedsabil83:
I always made these changes in mine. I thought it will be a good idea to be their out-of-the-box for those who use RTL direction with Tailwindcss.

mohamedsabil83:
Hi @taylorotwell,
I can be responsible for any change required to support RTL for the Tailwind version. Hope to re-think about accepting this as it's very important for us.

driesvints:
@mohamedsabil83 I'm sorry but we can't rely on third parties to maintain our own stuff. We've noticed enough times in the past that promises were made which couldn't be held in the end. It's probably best that you add this to your own code.

mohamedsabil83:
It's ok

PR #936, opened 2022-01-03
Gravatar for default profile photo url with UI-Avatars as fallback

onlime:
Make the default profile photo more configurable.

By default, using Gravatar with UI-Avatars as fallback. Like this, we get all avatars delivered by Gravatar with the following bonus:

- Prevent rate-limiting (ui-avatars.com uses Cloudflare that blocks you fast, while Gravatar has virtually no rate-limiting)
- Nice UI-Avatars with initials (which is currently not offered by Gravatar) as it was before
- Let the user configure other Gravatar default images like `mp|identicon|monsterid|wavatar|retro|robohash|blank`

If you want to use this before it gets merged, simply "extend" the Jetstream trait and use your own:

```php
<?php

namespace App\Traits;

trait HasProfilePhoto
{
    use \Laravel\Jetstream\HasProfilePhoto;

    {
        $hash         = md5(strtolower(trim($this->email)));
        $defaultImage = config('jetstream.profile_photo_fallback.default_image', 'initials');

        if ($defaultImage === 'initials') {
            $name     = urlencode($this->name); // name needs to be double-urlencoded for Gravatar fallback URL
            $size     = config('jetstream.profile_photo_fallback.initials.size', 64);
            $bg       = config('jetstream.profile_photo_fallback.initials.bg', 'ebf4ff');
            $color    = config('jetstream.profile_photo_fallback.initials.color', '7f9cf5');
            $fallback = urlencode("https://ui-avatars.com/api/{$name}/{$size}/{$bg}/{$color}");
        } else {
            $fallback = $defaultImage;
        }

        return "https://www.gravatar.com/avatar/{$hash}?d={$fallback}";
    }
}
```

### DIFF
--- a/config/jetstream.php
+++ b/config/jetstream.php
@@ -55,11 +55,37 @@ return [
     |--------------------------------------------------------------------------
     |
     | This configuration value determines the default disk that will be used
-    | when storing profile photos for your application's users. Typically
-    | this will be the "public" disk but you may adjust this if needed.
+    | when storing profile photos for your application's users. Typically,
+    | this will be the "public" disk, but you may adjust this if needed.
     |
     */
 
     'profile_photo_disk' => 'public',
+
+    /*
+    |--------------------------------------------------------------------------
+    | Profile Photo Fallback Image
+    |--------------------------------------------------------------------------
+    |
+    | This configuration value determines the default image that is used as a
+    | fallback if the user has not stored a profile photo.
+    | By default, we use Gravatar images with UI-Avatars as a fallback.
+    |
+    | Possible values for profile_photo_fallback.default_image:
+    | mp|identicon|monsterid|wavatar|retro|robohash|blank|initials
+    | (see http://gravatar.com/site/implement/images/)
+    | 'initials' (default) is a special extension where we use ui-avatars.com as
+    | a Gravatar fallback service to generate avatars with initials from names.
+    |
+    */
+
+    'profile_photo_fallback' => [
+        'default_image' => 'initials',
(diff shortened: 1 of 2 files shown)


kai0310:
@onlime 
Yes, you are right. However, not all users will want to use ui-avatars. This is just one example of using ui-avatars, and the current code is minimal.

onlime:
> @kai0310 Yes, you are right. However, not all users will want to use ui-avatars. This is just one example of using ui-avatars, and the current code is minimal.

Yes, but in the existing minimal implementation all requests go through ui-avatars, which is not optimal and not configurable. I understand that you want to keep Jetstream implementations minimal. But in this case, I would find it very welcome for a beginner to have it a bit less hard-coded and give the user a more functional default photo url. By setting `profile_photo_fallback.default_image` to e.g. `identicon` would turn off ui-avatars, and using Gravatar as default would maybe make some eyes shine.

I think a lot of users who bootstrap their project with Jetstream will keep a larger part of the components as is, so if we can accomplish more functionality with some single added lines, let's go for it! You could skim down my PR by not making it configurable. What do you think about this?:

```php
    {
        $hash = md5(strtolower(trim($this->email)));
        $name = urlencode($this->name);
        $fallback = urlencode("https://ui-avatars.com/api/{$name}/64/ebf4ff/7f9cf5");
        return "https://www.gravatar.com/avatar/{$hash}?d={$fallback}";
    }
```

that would maybe better fit the Jetstream philosophy.

taylorotwell:

PR #940, opened 2022-01-07
Anonymize default profile photo url calls

geisi:
Hello and Happy New Year!

This pr only affects instances with ProfilePhoto feature enabled .

At a security and privacy review of one of our Jetstream applications it was criticised that the applications sends the whole user name to an external service (https://ui-avatars.com/) on every request. So every single name of your application users potentially gets exposed to ui-avatars.com.

I think if someone randomly asks you to hand over all of your applications user names you won't approve that.

**I want to clarify that this is no criticism on services provided by ui-avatars.com they are awesome and we can be glad that these services exist.**

Since ui-avatars.com is only using the first letters of a users given name to generate its avatars this is completely unnecessary. So this PR corrects the application behaviour so that only the first letters of the name are added to the users default profile photo url.

I know that the rest of the world outside of Europe does not care that much about privacy concerns. But for all european and UK users this is a well being change.

DefaultUserPhotoUrl Before this PR:
https://ui-avatars.com/api/?name=Taylor+Otwell&color=7F9CF5&background=EBF4FF

And After:
https://ui-avatars.com/api/?name=T+O&color=7F9CF5&background=EBF4FF

Thank you for your amazing work and have a nice weekend!


taylorotwell:
> I know that the rest of the world outside of Europe does not care that much about privacy concerns. But for all european and UK users this is a well being change.

I do not think Europe is the only entity that cares about privacy.

geisi:
> > I know that the rest of the world outside of Europe does not care that much about privacy concerns. But for all european and UK users this is a well being change.
> 
> I do not think Europe is the only entity that cares about privacy.

Don't get me wrong I don't want to criticise or judge about other countries regularities. And I did not want to give this PR a political touch. :D

src/HasProfilePhoto.php
```diff
@@ -66,7 +67,11 @@ public function getProfilePhotoUrlAttribute()
      */
     protected function defaultProfilePhotoUrl()
     {
-        return 'https://ui-avatars.com/api/?name='.urlencode($this->name).'&color=7F9CF5&background=EBF4FF';
+        $name = trim(collect(explode(' ', $this->name))->map(function ($segment) {
+            return $segment[0] ?? '';
```
nvi9, line 71: @taylorotwell, @geisi as mentioned in a comment at https://github.com/laravel/framework/pull/40381, this implementation doesn't work as expected with multi-byte letters (eg. `Á`).

It should be either changed to `return Str::substr($segment, 0, 1);`, or the linked PR should be included in the framework package and its new function should be used here like this:

```
return 'https://ui-avatars.com/api/?name='.urlencode(Str::initials($this->name)).'&color=7F9CF5&background=EBF4FF';
```

PR #941, opened 2022-01-08
Changed profile preview generation from raw JS to Alpine

titonova:
Looks  much more cleaner and tidier now, plus reducing the DOM content a teeny bit :-)


taylorotwell:
I'm making anymore changes to profile photos for a bit. It's generating a lot of PRs.

PR #943, opened 2022-01-10
Generate default profile avatar locally if it is supported.

ravibpatel:
Rather than using a third party service to generate the avatar, it is better if we generate the avatar locally. That way it can load faster, and we can still fallback to third party service if it is not possible to generate it due to lack of gd or imagick PHP extension.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,8 @@
         "ext-json": "*",
         "illuminate/support": "^8.0",
         "jenssegers/agent": "^2.6",
-        "laravel/fortify": "^1.6.1"
+        "laravel/fortify": "^1.6.1",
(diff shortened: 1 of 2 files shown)


driesvints:
Can you show some comparison screenshots?

ravibpatel:
Sure, here it is. Difference in initials comes due to fix in #940. As we are generating avatar locally, we can use full name instead of initials. It is also possible to publish the config file of Intervention and change driver to imagick if gd is not available. By default, it uses gd to generate it. If any of the dependency is unavailable, it will fall back to the original method of getting the avatar.

### Before

Before

### After

After

taylorotwell:
Feel free to change this in your own code if you wish.

PR #946, opened 2022-01-11
Remove duplicated code usage

GeoSot:
Just a minor refactoring as `\Laravel\Jetstream\HasTeams::teamRole` is almost same with `\Laravel\Jetstream\HasTeams::hasTeamRole`


taylorotwell:
If the code is not broken I do not plan to change it.

PR #949, opened 2022-01-11
Add route prefix and subdomain to config file

romeroqe:
It allows to give more visibility to this functionality, otherwise, it is necessary to check the `JetstreamServiceProvider.php` file to know it.


taylorotwell:
I'll try to get around to adding this myself. I'll have to adjust the comment.

PR #962, opened 2022-01-22
[2.x] switch to Laravel 8.80 >= route group controllers

geisi:
This PR refactors the route files to the new route group controller syntax by @lukeraymonddowning added in this [pr](https://github.com/laravel/framework/pull/40276).

Thank you for your time and have a nice weekend!

### DIFF
--- a/routes/inertia.php
+++ b/routes/inertia.php
@@ -42,23 +42,31 @@ Route::group(['middleware' => config('jetstream.middleware', ['web'])], function
 
         // API...
         if (Jetstream::hasApiFeatures()) {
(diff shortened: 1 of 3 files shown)


geisi:
I'm not quite sure why this does not work... I have to dive deeper into this problem. So this PR is marked as draft.

geisi:
It seems that the Laravel Router has a problem with the new route group feature in combination with group namespaces. So i removed the namespace configuration in the JetstreamServiceProvider.

I think this should be a Bug in the Router? The group namespace only should not be applied when using full class path references.

taylorotwell:

PR #964, opened 2022-01-25
[2.x] Added "dir='ltr'"  attribute to the html tag, according to tw 3's specs

titonova:
Tailwind CSS docs specifically states to  _"always set the direction, even if left-to-right is your default"_

https://tailwindcss.com/docs/hover-focus-and-other-states#rtl-support

### DIFF
--- a/stubs/inertia/resources/views/app.blade.php
+++ b/stubs/inertia/resources/views/app.blade.php
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="{{ str_replace('_', '-', app()->getLocale()) }}">
+<html lang="{{ str_replace('_', '-', app()->getLocale()) }}" dir="ltr">
     <head>
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1">


driesvints:
Can you also update inertia? https://github.com/laravel/jetstream/blob/2.x/stubs/inertia/resources/views/app.blade.php#L2

driesvints:
I also think you mistook this because right below that section it says:

> Remember, these modifiers are only useful if you are building a site that needs to support both left-to-right and right-to-left layouts. If you’re building a site that only needs to support a single direction, you don’t need these modifiers 

And I don't think many apps are supporting both.

titonova:
> I also think you mistook this because right below that section it says:
> 
> > Remember, these modifiers are only useful if you are building a site that needs to support both left-to-right and right-to-left layouts. If you’re building a site that only needs to support a single direction, you don’t need these modifiers
> 
> And I don't think many apps are supporting both.

Fair. But noticing that attribute may help sites supporting it to know that it is meant to be there, compulsory

titonova:
> Can you also update inertia? https://github.com/laravel/jetstream/blob/2.x/stubs/inertia/resources/views/app.blade.php#L2

Should I still do this? Or are you closing the issue :-(

titonova:
> > Can you also update inertia? https://github.com/laravel/jetstream/blob/2.x/stubs/inertia/resources/views/app.blade.php#L2
> 
> Should I still do this? Or are you closing the issue :-(

Ah well, Imma do it anyways

PR #966, opened 2022-01-25
Always return array to ui from allTeams()

scrummitch:
Hi there, this is my first laravel PR

The `$user->allTeams()` method returns a merged collection in the HasTeams trait with `$this->ownedTeams` and `$this->teams`.

Unfortunately eloquent `sortBy` method will re-index the keys out of order, so json encoding the output value of this method will convert it to a map in javascript rather than an array.


driesvints:
Thanks @scrummitch!

Issue #944, opened 2022-01-10
Multiple roles in team


gkakosouris:
Currently, Jetstream only allows for a user to have one role in a team.

Do you have any plans to allow users to have multiple roles in a team? 
Would you be open to a pull-request from the community that would add support for this?

driesvints:
At the moment we have no plans for this, sorry.

GeoSot:
@driesvints we can imagine your difficult position there, and that you may doesn't have any plans on this feature.

As this feature is very important for many developers (most business logics demand it), and jetstream tries to play an important role on authorization functionality, could you please give it a chance, reconsidering & answering the second question too?

It's too pity for many people to try to adapt on this official package and have to do workarounds to achieve this feature

PR #970, opened 2022-01-27
Support multiple Roles per Team

GeoSot:
Given that jetstream is a popular official laravel package that gives a head start to many projects, and provides basic acl functionality, I would like to help on this, providing a way to support **multiple** user roles per Team. 

During research & planning for different projects, I have been in the need of an acl package to cover business needs. 
The only thing that filters out this package from the eligible choices is the lack of multiple roles support.

So taking advantage of #944, I opened this PR as a proof of concept.

It **includes** the proper **PHP scoped changes** & **testing,** & **needs** tweaks on the **front templates**

<br><br>
_PS: Probably I will be suggested to begin my personal fork and continue on it, but it would be better for most devs, to support & trust one package, instead of random subversions around github._

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.x-dev"
+            "dev-master": "3.x-dev"
         },
         "laravel": {
             "providers": [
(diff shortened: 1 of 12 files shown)


GeoSot:
@driesvints I 've rebased my origin branch, but doesn't appear here, probably because is closed.  Do you want me to open a new PR, or you will re-open this?

driesvints:
New PR, thanks

PR #976, opened 2022-02-10
Install command find the path to the php binary used

Yi-pixel:
When I have multiple php versions, when trying to differentiate them using different aliases like `php`(7.1), `php74`, `php81`, now when I execute `php81 artisan jetstream:install livewire` in CLI , here will be unexpectedly fallback to `php`(7.1) execution, which will fail.

Now he will correctly get the `php81` I am using

I'm not sure how to write a test case for this modification, but it works for me, and I installed it by modifying the project's composer.json.

```json
    "repositories": [
        {
            "type": "path",
            "url": "C:\\web\\server\\laravel\\jetstream"
        }
    ],
```

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -6,6 +6,7 @@ use Exception;
 use Illuminate\Console\Command;
 use Illuminate\Filesystem\Filesystem;
 use Illuminate\Support\Str;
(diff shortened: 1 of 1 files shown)


Yi-pixel:
The methods on the `Application` object can also be used here: `$this->getApplication()->phpBinary()`

PR #981, opened 2022-02-15
Use typesetter’s apostrophe instead of typewriter’s one when creating user’s own team

maximal:
It looks better and more semantically correct in UIs.

https://en.wikipedia.org/wiki/Apostrophe


driesvints:
Feel free to do this in your own app, thanks.

PR #983, opened 2022-02-20
Fix: don't show error when user has no teams

M074554N:
This is a quick fix to avoid throwing `ErrorException` when Jetstream is installed but the user has no teams yet.

Description: when Jetstream is installed and the teams feature is enabled but the user doesn't have any teams, it throws an `ErrorException` with the message `Attempt to read property "name" on null`.

Error screenshot attached.
Screenshot 2022-02-20 at 11 23 46 PM


driesvints:
A user *always* has a team when the teams feature is enabled. You need to decide upfront upon installation wether your app will use teams or not.

PR #989, opened 2022-03-01
Use Alpine's @click.prevent for log out

jrd-lewis:
This PR will switch the ```Log Out``` links in the Livewire navigation menu component from vanilla JS to Alpine's ```@click.prevent``` directive for a cleaner approach with the same result. Also, it should eliminate squawking from IDE's about using the ```event``` context when preventing the default action on clicking the links.

### DIFF
--- a/stubs/livewire/resources/views/navigation-menu.blade.php
+++ b/stubs/livewire/resources/views/navigation-menu.blade.php
@@ -109,12 +109,11 @@
                             <div class="border-t border-gray-100"></div>
 
                             <!-- Authentication -->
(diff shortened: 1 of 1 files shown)


taylorotwell:
Clean - thanks.

mikemand:
This would also be great for the [`switchable-team` component](https://github.com/laravel/jetstream/blob/51be26911598ee47710124cf8df5004a0f7d6a59/resources/views/components/switchable-team.blade.php#L10). @jrd-lewis, do you want to PR it or shall I?

PR #991, opened 2022-03-03
Solving profile photo issue when using Socialite

mnaderian:
When using Laravel Jetstream along with Socialite, the avatar URL which is retrieved from the provider gets broken!  

For example if the URL is like this:  
``
https://avatars.githubusercontent.com/u/30769988
``

Jetstream turns it to something like this:
``
/storage/profile-photos/https://avatars.githubusercontent.com/u/30769988
``

This PR solves that problem.

### DIFF
--- a/src/HasProfilePhoto.php
+++ b/src/HasProfilePhoto.php
@@ -59,9 +59,15 @@ trait HasProfilePhoto
      */
     public function getProfilePhotoUrlAttribute()
     {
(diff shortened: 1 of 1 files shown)


ArnaudLier:
I don't think this will get merged as there is no problem in the base logic, developers that integrate Laravel Socialite in their apps will make this change themselves.

taylorotwell:

PR #998, opened 2022-03-15
Adding Storage Path Support to Has Profile Photo Trait

baminc:
In order to use the hasProfilePhoto Trait on other Models i added a second parameter to set the storage path for the model. With this you can store profile photos for teams in a different folder as user profile photos.


driesvints:
Adding a new argument to a method is a breaking change for anyone extending the a class that implements this and overriding the method. This needs to go to master.

PR #1007, opened 2022-03-18
Add `lang` support to Inertia stubs

xiCO2k:
This PR is a possible addition to add to the Inertia stubs, looking on the blade views, it looks like it's using the translation helper.

My proposal is to do the same for Inertia, but since there is no built in way to do it, I built a package called [`laravel-vue-i18n`](https://github.com/xiCO2k/laravel-vue-i18n) couple of months ago for that exact purpose, it even supports the same syntax for replacements.

If you guys think this is a good addition to have, I can finish the implementation.

Let me know if there is any questions.


taylorotwell:
Hey there! We have thought about adding this in the past but it just really adds a lot of noise to the templates.

xiCO2k:
Got it.

PR #1008, opened 2022-03-18
2FA Setup Key

ps-sean:
* Allow user to enter setup key in authenticator apps if they can't scan the QR code

The inertia side relies on the matching fortify pull request: https://github.com/laravel/fortify/pull/371


ps-sean:
@taylorotwell Wouldn't this now fail in inertia due to fortify still having the route as '/user/two-factor-setup-key'?

driesvints:
@ps-sean this was later updated: https://github.com/laravel/jetstream/commit/59f0972274afba7f271552139ac4ccd46ede64ed

PR #1009, opened 2022-03-19
Use `.alias` method from `Mix`

xiCO2k:
This PR removes the `webpack.config.js` file in order to use the `alias` method from `Mix`.


taylorotwell:
@xiCO2k should we do the same in Breeze?

xiCO2k:
will check.

LucaRed:
I think this breaks autocomplete and code navigation in IDEs that search for aliases in a `webpack.config.js` file.

PR #1012, opened 2022-03-20
[2.x] Add Inertia SSR Support

xiCO2k:
This PR adds the SSR support for the Inertia stack.

To not be a default option, I added a `--ssr` option, this way we will only include the SSR part if the user wants.

to have the SSR running make sure to run:

```sh
npm run prod
```

then to start the ssr server run:

```sh
node public/js/ssr.js
```

More information on the Inertia website: https://inertiajs.com/server-side-rendering

The directive `@inertiaHead` added to the `app.blade.php` will not do anything if the SSR is not enabled.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -289,16 +289,13 @@ EOF;
                 '@inertiajs/inertia' => '^0.11.0',
                 '@inertiajs/inertia-vue3' => '^0.6.0',
                 '@inertiajs/progress' => '^0.2.7',
-                '@inertiajs/server' => '^0.1.0',
                 '@tailwindcss/forms' => '^0.5.0',
                 '@tailwindcss/typography' => '^0.5.2',
                 'postcss-import' => '^14.0.2',
                 'tailwindcss' => '^3.0.0',
                 'vue' => '^3.2.31',
                 '@vue/compiler-sfc' => '^3.2.31',
-                '@vue/server-renderer' => '^3.2.31',
                 'vue-loader' => '^17.0.0',
-                'webpack-node-externals' => '^3.0.0'
             ] + $packages;
         });
 
@@ -502,7 +499,7 @@ EOF;
             return [
                 '@inertiajs/server' => '^0.1.0',
                 '@vue/server-renderer' => '^3.2.31',
-                'webpack-node-externals' => '^3.0.0'
+                'webpack-node-externals' => '^3.0.0',
             ] + $packages;
         });
 
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -513,7 +513,7 @@ EOF;
             });
 
         $this->replaceInFile("'enabled' => false", "'enabled' => true", config_path('inertia.php'));
-        $this->replaceInFile("mix --production", "mix --production --mix-config=webpack.ssr.mix.js && mix --production", base_path('package.json'));
+        $this->replaceInFile('mix --production', 'mix --production --mix-config=webpack.ssr.mix.js && mix --production', base_path('package.json'));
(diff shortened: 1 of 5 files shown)


taylorotwell:
@xiCO2k I guess we could also port the same option to Breeze now. 😅

xiCO2k:
Count me in :)

bakerkretzmar:
Hi! Maintainer of Ziggy here. Running `new Ziggy` on every single request is _very_ strongly discouraged because it gets slow if you have tons of routes.

Not sure who the best person is to answer this question, but could we cache it? Like can we use `Cache` in jetstream / a starter kit?

@xiCO2k is there any way to tell on the server if Inertia is running in SSR 'mode' so we don't generate Ziggy's routes unless it is? I assume in SSR mode it doesn't matter if it's a bit slow because all the pages are 'visited' and rendered in advance, right?

xiCO2k:
> Hi! Maintainer of Ziggy here. Running `new Ziggy` on every single request is _very_ strongly discouraged because it gets slow if you have tons of routes.
> 
> 
> 
> Not sure who the best person is to answer this question, but could we cache it? Like can we use `Cache` in jetstream / a starter kit?
> 
> 
> 
> @xiCO2k is there any way to tell on the server if Inertia is running in SSR 'mode' so we don't generate Ziggy's routes unless it is? I assume in SSR mode it doesn't matter if it's a bit slow because all the pages are 'visited' and rendered in advance, right?

Hey, normally if `typeof window === 'undefined'` you are on SSR Mode.

Cheers

timacdonald:
@bakerkretzmar I've made a note of this and will take a look at what we can do. Happy for someone else to jump on it, but i'll circle back shortly and see what we can do.

timacdonald:
Hey folks,

I spent some time looking at this today.

My initial feeling is that if we introduce caching it will have flow on effects around the development and deployment story, which I don't love the idea of. We would have to only cache in prod to not impact local development, and then we'd also have to bust the cache on deployment.

We also push everything out to user-land, so if this does become a performance issue (which won't be the case with a new Jetstream / Breeze project as they are only for new projects),  the individual project can introduce caching in their own applications.

I believe we still need to generate the routes in SSR and non-SSR as the SSR generated pages can still utilise the `route()` helper.

I wonder if the performance implications of Ziggy with a large routes definition could be written up on the Ziggy docs for users that do end up hitting the issue and some potential solutions?

I'm not calling any shots here btw, just my thinking after taking a quick look - so very open to discussion and push back.

bakerkretzmar:
That sounds good to me, I can work on adding some notes on this to the Ziggy docs. It's come up a few times in discussions/issues so it's definitely worth writing up properly.

I'll think about better ways to only run this once, too. In most apps I'd just suggest using the `@routes` Blade directive since it's uncommon for routes to change in production, so whatever loads initially should be fine to keep around globally and re-use. I've never worked with SSR and don't really know how it works so I'm not sure if that would still make sense there.

timacdonald:
In a non-SSR request, we would have a single request come through and do...

```blade
<head>
    @routes
</head>
```

Nothing special here. This means that `new Ziggy` is called once on every request.

---

In SSR mode, a similar thing will happen still. 

1. The request will come into Laravel.
2. The "before" Inertia middleware is called, which in turn calls the "share" function where we can see `new Ziggy`.
3. Then, in the same request, Laravel will generate the view - which still calls the `@routes` directive...


My understanding is that as Ziggy statically caches its calculated routes in the constructor, this should have all the same performance issues / tradeoffs as the non-SSR version.

That, again, is my understanding of how all this interacts.

santutu:
npm run build
node bootstrap/ssr/ssr.js

PR #1021, opened 2022-03-27
prop to prevent closing dropdown on item click

devsrv:
`closeOnClick` prop to prevent the closing of dropdown when any item is clicked.

useful when the dropdown items are non link item and require the dropdown to stay opened even after click, for example a delete action button that shows a pair of confirm / cancel buttons.

Usage:
```html
<x-jet-dropdown :close-on-click="false" ...>
  ...
</x-jet-dropdown>
```

Exaple:
ezgif com-gif-maker


taylorotwell:
Feel free to add this to your own application.

PR #1024, opened 2022-03-30
[2.x] Not require `Ziggy` on `ShareInertiaData` middleware.

### LINKED ISSUE #1023: Ziggy is no longer optional
- Jetstream Version: 2.7.0
- Jetstream Stack: Inertia
- Uses Teams:  no
- Laravel Version: 9.6.0
- PHP Version: 8.1.4
- Database Driver & Version: Postgres 14.2

### Description:
Well with the newly added Intertia SSR Support https://github.com/laravel/jetstream/pull/1012 Ziggy is now imposed to us

https://github.com/laravel/jetstream/blob/090fdfbdf7f622836d2e6336f27f9749028247c2/src/Http/Middleware/ShareInertiaData.php#L10

https://github.com/laravel/jetstream/blob/090fdfbdf7f622836d2e6336f27f9749028247c2/src/Http/Middleware/ShareInertiaData.php#L60-L63

So i believe theres no more reason to install ziggy as an optional dependecy

https://github.com/laravel/jetstream/blob/090fdfbdf7f622836d2e6336f27f9749028247c2/src/Console/InstallCommand.php#L284

and it should now be part of the package, also i think this is a breaking chance (since it is) because y if people decide not to use ziggy because as the creator of Inertia Jonathan Reinink says :https://twitter.com/reinink/status/1298577581650980871 some of use get rid of Ziggy because it causes more issues than it solves. But now we are forced to. wich is ok, it's your package guys, thanks for all the work you. just, like I mentioned above, now require the package withing the package and don't make it optional if we are not able to remove it, or maybe make it so we can publish de middleware and we might if we choose get rid of it completley.

To add on that, i think it's even bad practice to be using ziggy the way the Pull Request uses it:
https://github.com/laravel/jetstream/blob/090fdfbdf7f622836d2e6336f27f9749028247c2/stubs/inertia/resources/js/ssr.js#L18-L27
 since, in Vue 3 mixins are no longer recommended https://vuejs.org/api/options-composition.html#mixins

And since in the same relase we just got the composition API https://github.com/laravel/jetstream/pull/1001 as default it does make sense not to force ziggy on us

Sorry for the probably bad english btw, not my first tongue

### Steps To Reproduce:

Fresh install of Jetstream with the removal of Ziggy as dependency

xiCO2k:
This PR fixes #1023

Removes the addition of `Ziggy` on the `ShareInertiaData` middleware, and only add `Ziggy` to the `HandleInertiaData` stub if `--ssr` is passed, so that way if anyone wants to remove, they have the full control.

Thanks,
Francisco

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -511,6 +511,8 @@ EOF;
                 $this->output->write($output);
             });
 
+        copy(__DIR__.'/../../stubs/inertia/app/Http/Middleware/HandleInertiaRequests.php', app_path('Http/Middleware/HandleInertiaRequests.php'));
+
         $this->replaceInFile("'enabled' => false", "'enabled' => true", config_path('inertia.php'));
         $this->replaceInFile('mix --production', 'mix --production --mix-config=webpack.ssr.mix.js && mix --production', base_path('package.json'));
     }
--- a/src/Http/Middleware/ShareInertiaData.php
+++ b/src/Http/Middleware/ShareInertiaData.php
@@ -7,7 +7,6 @@ use Illuminate\Support\Facades\Session;
 use Inertia\Inertia;
 use Laravel\Fortify\Features;
 use Laravel\Jetstream\Jetstream;
-use Tightenco\Ziggy\Ziggy;
 
 class ShareInertiaData
 {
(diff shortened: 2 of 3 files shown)


taylorotwell:
@xiCO2k how will this middleware actually be invoked? where is it registered with the routes?

xiCO2k:
Yes, the middleware is already added here:

https://github.com/laravel/jetstream/blob/05b6781c7b69123935b4410dfbfe223bced9187b/src/Console/InstallCommand.php#L345

PR #1027, opened 2022-04-03
Use Alpine's @click.prevent for switching teams

jrd-lewis:
This PR will switch the `switchable-team` component in the Livewire stack from vanilla JS to Alpine's `@click.prevent` directive, like my previous PR #989 for the log out link.

### DIFF
--- a/resources/views/components/switchable-team.blade.php
+++ b/resources/views/components/switchable-team.blade.php
@@ -1,13 +1,13 @@
 @props(['team', 'component' => 'jet-dropdown-link'])
 
-<form method="POST" action="{{ route('current-team.update') }}">
+<form method="POST" action="{{ route('current-team.update') }}" x-data>
     @method('PUT')
     @csrf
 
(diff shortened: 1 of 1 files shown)


seabasss:
This commit is breaking the jetstream component `x-jet-switchable-team` that is used on `navigation-menu.blade.php`. I now get the following error:

```syntax error, unexpected token "@", expecting variable or "{" or "$"```

Rolling back to 7.2.1 still works.

brownkevg:
I also ran into the same issue @seabasss had with version 7.2.2

`yntax error, unexpected token "@", expecting variable or "{" or "$"`

This fired after creating a new user for me after following the instructions [here](https://jetstream.laravel.com/2.x/installation.html#livewire) to create a new app
I was able to downgrade to 7.2.1 and it worked as well for me.

driesvints:
Does it work for you both if you use `@@click.prevent="$root.submit();"`? Note the double `@@`.

PR #1030, opened 2022-04-05
[2.x] Fix error handling during 2FA confirmation

### LINKED ISSUE #1028: Confirming 2FA lacks error message and refreshes page/state incorrectly (Inertia stack)
- Jetstream Version: 2.7.1
- Jetstream Stack: Inertia
- Uses Teams: no
- Laravel Version: 9.6.0
- PHP Version: 8.1.2
- Database Driver & Version: mysql

### Description:
This is an **Inertia stack only** problem. The Livewire stack works without any problems. 

When enabling 2FA and providing a wrong or empty confirmation code, the user will not see any error message. The UI partially refreshes, e.g. buttons change but texts stay the same. I am happy to provide a PR as soon as I find the time. I do not know how trivial the solution actually is. I provided a solution for one part and some sources for the other part of the problem, see below.

https://user-images.githubusercontent.com/8204574/161546265-028acf02-fa74-4e5d-bc68-3c77a62c6fc4.mov

### Steps To Reproduce:

- install a new application ([docs](https://laravel.com/docs/9.x/installation#getting-started-on-macos))
- install Laravel Jetstream ([docs](https://jetstream.laravel.com/2.x/installation.html#installing-jetstream))
- install Inertia stack ([docs](https://jetstream.laravel.com/2.x/installation.html#or-install-jetstream-with-inertia))
- finalize installation ([docs](https://jetstream.laravel.com/2.x/installation.html#finalizing-the-installation))
- make sure the fortify 2FA-feature is enabled with `'confirm' => true,` (default)
- register a new user
- open profile settings
- click button to enable 2FA, `$user->two_factor_secret` is set in database
- enter wrong or empty confirmation code
- page refreshes, no error message is displayed, user in database gets reset, e.g. `two_factor_secret` is null again

### Missing error message:
When provided a wrong/empty code, Fortify throws the validation error message into an [error bag](https://github.com/laravel/fortify/blob/1.x/src/Actions/ConfirmTwoFactorAuthentication.php#L43). Adding an additional option and provide the error bag name in the [Vue component](https://github.com/laravel/jetstream/blob/2.x/stubs/inertia/resources/js/Pages/Profile/Partials/TwoFactorAuthenticationForm.vue#L70) fixes that and the error can be retrieved as intended.

```
confirmationForm.post("/user/confirmed-two-factor-authentication", {
    errorBag: "confirmTwoFactorAuthentication",
    preserveScroll: true,
    preserveState: true,
    onSuccess: () => {
      confirming.value = false;
      qrCode.value = null;
      setupKey.value = null;
    },
  });


...


<JetInputError :message="confirmationForm.errors.code" class="mt-2" />
```

### Refreshing page / state / user:
When starting the confirmation process, `$user->two_factor_secret` is set in the database. There is a [computed property](https://github.com/laravel/jetstream/blob/2.x/stubs/inertia/resources/js/Pages/Profile/Partials/TwoFactorAuthenticationForm.vue#L29) to check if 2FA is enabled. This computed property relies on the user's `two_factor_secret`, see [here](https://github.com/laravel/jetstream/blob/2.x/src/Http/Middleware/ShareInertiaData.php#L51). After providing a wrong/empty conformation code, [this code](https://github.com/laravel/jetstream/blob/2.x/src/Http/Controllers/Inertia/Concerns/ConfirmsTwoFactorAuthentication.php#L38) will be executed and the user gets reset in the database. The computed property then changes its value from `true` to `false` and partially refreshes the UI.

ManuelLeiner:
This PR fixes the error handling during 2FA confirmation for the Inertia Stack. The Livewire Stack already works as expected. 

- the error message for an invalid confirmation code will be correctly read from the form error bag, see [Fortify counterpart](https://github.com/laravel/fortify/blob/1.x/src/Actions/ConfirmTwoFactorAuthentication.php#L43)
- texts and buttons refresh accordingly to the current confirmation state
- confirmation form resets when 2FA process
- adds an additional check to determine if the confirmation process finished

Fixes #1028

### DIFF
--- a/src/Http/Controllers/Inertia/Concerns/ConfirmsTwoFactorAuthentication.php
+++ b/src/Http/Controllers/Inertia/Concerns/ConfirmsTwoFactorAuthentication.php
@@ -77,7 +77,8 @@ trait ConfirmsTwoFactorAuthentication
      */
     protected function neverFinishedConfirmingTwoFactorAuthentication(Request $request, $currentTime)
     {
-        return is_null($request->user()->two_factor_confirmed_at) &&
+        return ! array_key_exists('code', $request->session()->getOldInput()) &&
+            is_null($request->user()->two_factor_confirmed_at) &&
             $request->session()->get('two_factor_confirming_at', 0) != $currentTime;
     }
 }
--- a/stubs/inertia/resources/js/Pages/Profile/Partials/TwoFactorAuthenticationForm.vue
+++ b/stubs/inertia/resources/js/Pages/Profile/Partials/TwoFactorAuthenticationForm.vue
@@ -1,5 +1,5 @@
 <script setup>
-import { ref, computed } from 'vue';
+import { ref, computed, watch } from 'vue';
 import { Inertia } from '@inertiajs/inertia';
 import { useForm, usePage } from '@inertiajs/inertia-vue3';
 import JetActionSection from '@/Jetstream/ActionSection.vue';
@@ -29,6 +29,13 @@ const confirmationForm = useForm({
 const twoFactorEnabled = computed(
     () => ! enabling.value && usePage().props.value.user.two_factor_enabled,
 );
+
+watch(twoFactorEnabled, () => {
+    if (! twoFactorEnabled.value) {
+        confirmationForm.reset();
+        confirmationForm.clearErrors();
+    }
+});
 
 const enableTwoFactorAuthentication = () => {
     enabling.value = true;
@@ -67,6 +74,7 @@ const showRecoveryCodes = () => {
 
 const confirmTwoFactorAuthentication = () => {
     confirmationForm.post('/user/confirmed-two-factor-authentication', {
+        errorBag: "confirmTwoFactorAuthentication",
         preserveScroll: true,
         preserveState: true,
         onSuccess: () => {
@@ -111,7 +119,7 @@ const disableTwoFactorAuthentication = () => {
                 You have enabled two factor authentication.
             </h3>
 
-            <h3 v-else-if="confirming" class="text-lg font-medium text-gray-900">
+            <h3 v-else-if="twoFactorEnabled && confirming" class="text-lg font-medium text-gray-900">
                 Finish enabling two factor authentication.
             </h3>
 


driesvints:
Please note that we don't review draft PR's. Please mark this as ready when you're done 👍

src/Http/Controllers/Inertia/Concerns/ConfirmsTwoFactorAuthentication.php
```diff
@@ -77,7 +77,8 @@ protected function hasJustBegunConfirmingTwoFactorAuthentication(Request $reques
      */
     protected function neverFinishedConfirmingTwoFactorAuthentication(Request $request, $currentTime)
     {
-        return is_null($request->user()->two_factor_confirmed_at) &&
+        return ! array_key_exists('code', $request->session()->getOldInput()) &&
```
ManuelLeiner, line 80: @driesvints this change feels hacky but in my tests everything worked just fine.

Here I try to find out if the user just entered a 2FA confirmation code which got rejected.

src/Http/Controllers/Inertia/Concerns/ConfirmsTwoFactorAuthentication.php
```diff
@@ -77,7 +77,8 @@ protected function hasJustBegunConfirmingTwoFactorAuthentication(Request $reques
...
+            is_null($request->user()->two_factor_confirmed_at) &&
             $request->session()->get('two_factor_confirming_at', 0) != $currentTime;
```
driesvints, line 82: I'm wondering if this like is the culprit and we need to allow a window of time here instead of checking with `!=`?

ManuelLeiner:
Here are some log outputs from debugging the [additional check](https://github.com/laravel/jetstream/blob/cdfa12c98fe8438f77f01ad7640485152cca26fc/src/Http/Controllers/Inertia/Concerns/ConfirmsTwoFactorAuthentication.php#L80).

```
// initial call to /user/profile
[2022-04-05 10:05:23 (UTC+00:00)] local.DEBUG: Session old input:
[2022-04-05 10:05:23 (UTC+00:00)] local.DEBUG: Code exists in old input:  [false]

// click on 'Enable'
[2022-04-05 10:05:39 (UTC+00:00)] local.DEBUG: Session old input:
[2022-04-05 10:05:39 (UTC+00:00)] local.DEBUG: Code exists in old input:  [false]

// confirm with empty input field
[2022-04-05 10:05:51 (UTC+00:00)] local.DEBUG: Session old input:  {"code":null}
[2022-04-05 10:05:51 (UTC+00:00)] local.DEBUG: Code exists in old input:  [true]

// confirm with a wrong code
[2022-04-05 10:06:07 (UTC+00:00)] local.DEBUG: Session old input:  {"code":"123456"}
[2022-04-05 10:06:07 (UTC+00:00)] local.DEBUG: Code exists in old input:  [true]

// updating the user's name
[2022-04-05 10:06:19 (UTC+00:00)] local.DEBUG: Session old input:
[2022-04-05 10:06:19 (UTC+00:00)] local.DEBUG: Code exists in old input:  [false]
```

- Up until the last two log messages the 2FA columns of the user model (`two_factor_secret`, `two_factor_recovery_codes`) stay set --> works as intended
- Updating the user's name while still in the middle of 2FA confirmation resets the 2FA columns in storage --> works as intended
- Opening the profile again while still in the middle of 2FA confirmation resets the 2FA columns in storage --> works as intended

src/Http/Controllers/Inertia/Concerns/ConfirmsTwoFactorAuthentication.php
taylorotwell: This is correct @driesvints

ManuelLeiner: You are right. That works as well. For testing I changed the code to the following:

```
        return is_null($request->user()->two_factor_confirmed_at) &&
            $currentTime - $request->session()->get('two_factor_confirming_at', 0) > 60 * 5;
```

Is that the general direction you had in mind?

Do you prefer the window of time to be configurable (e.g. `config/jetstream.php` as it is not really part of `fortify.php` or `auth.php`) instead of being hard-coded?

If we choose a time window that is too short, we will face problems sooner or later. I see it coming that users search for their mobile phone or get distracted. Then they scan or scanned a meanwhile outdated QR code and get frustrated. I might as well focus too much on that unimportant detail here ;)

If we choose to make the window long enough this might not be a concern.

ps-sean: @ManuelLeiner I think we can copy the timing window from livewire which uses this:
`$maximumSecondsSinceConfirmation = $maximumSecondsSinceConfirmation ?: config('auth.password_timeout', 900);`

ManuelLeiner: I will have a look at it. Thanks for the input. :)

driesvints: @taylorotwell are you sure? I thought we implemented that time session to allow some time for the user to attempt setting up the 2FA. Checking that the saved time session matches exactly with the current time seems to defeat that purpose? Maybe I'm not seeing the full picture here.

taylorotwell: I was trying tos ay I don't think this code should be changed.

taylorotwell:
Thanks!

Issue #1033, opened 2022-04-06
JetConfirmsPassword on Confirming 2FA


hailwood:
- Jetstream Version: 2.7.1
- Jetstream Stack: both
- Uses Teams: N/A but no
- Laravel Version: N/A but 9.6.0
- PHP Version: N/A but 8.1.0
- Database Driver & Version: N/A

### Description:
In the Inertia stack the `Confirm` button for 2FA is wrapped in 
```jsx
<JetConfirmsPassword @confirmed="confirmTwoFactorAuthentication">...</JetConfirmsPassword>
```
However the input on the form the button relates to triggers the same function via the enter key
```
@keyup.enter="confirmTwoFactorAuthentication"
```
Meaning the confirms password can be bypassed by simply pressing <kbd>enter</kbd> on the input.

Similary for the Livewire stack it's wrapped in
```html
<x-jet-confirms-password wire:then="confirmTwoFactorAuthentication">
```
However the input on the form the button relates to triggers the same function via the enter key
```
wire:keydown.enter="confirmTwoFactorAuthentication"
```

I believe this wrapping JetConfirmsPassword/x-jet-confirms-password probably shouldn't be there as the user would have recently confirmed their password when they clicked the `Enable` button, so unless they clicked enable and then left their device long enough for the confirm password timeout to occur (3 hours by default iirc) they'll never get prompted anyway.

### Steps To Reproduce:
Enable 2FA and 2FA Confirming

### Relevant Files
https://github.com/laravel/jetstream/blob/2.x/stubs/livewire/resources/views/profile/two-factor-authentication-form.blade.php
https://github.com/laravel/jetstream/blob/2.x/stubs/inertia/resources/js/Pages/Profile/Partials/TwoFactorAuthenticationForm.vue

driesvints:
Thanks for reporting this. I think removing the key press listeners is the easiest solution for now. I sent in a PR for that here: https://github.com/laravel/jetstream/pull/1035

driesvints:
@hailwood we can't seem to recreate this. Pressing enter on the field invokes the password confirmation.

driesvints:
In livewire I seem to be getting a 403 model when entering in the code field after executing your route. Fresh Laravel app. This isn't ideal indeed.. 

<img width="1876" alt="Screenshot 2022-04-08 at 12 09 58" src="https://user-images.githubusercontent.com/594614/162415341-e58dd495-ce12-4411-92e8-82d33be9575e.png">

PR #1035, opened 2022-04-07
[2.x] Fix confirm password circumvention

driesvints:
This PR removes the ability to submit the two factor form by pressing the enter key in the code fields as those can be used to circumvent the password confirmation (which is only tied to the submit button). It seems removing the enter key press is the easiest solution. Otherwise we'll have to tie the same logic within the password confirm buttons to the key press listeners.

See https://github.com/laravel/jetstream/issues/1033

### DIFF
--- a/stubs/inertia/resources/js/Pages/Profile/Partials/TwoFactorAuthenticationForm.vue
+++ b/stubs/inertia/resources/js/Pages/Profile/Partials/TwoFactorAuthenticationForm.vue
@@ -165,7 +165,6 @@ const disableTwoFactorAuthentication = () => {
                             inputmode="numeric"
                             autofocus
                             autocomplete="one-time-code"
-                            @keyup.enter="confirmTwoFactorAuthentication"
                         />
 
                         <JetInputError :message="confirmationForm.errors.code" class="mt-2" />
--- a/stubs/livewire/resources/views/profile/two-factor-authentication-form.blade.php
+++ b/stubs/livewire/resources/views/profile/two-factor-authentication-form.blade.php
@@ -53,8 +53,7 @@
                         <x-jet-label for="code" value="{{ __('Code') }}" />
 
                         <x-jet-input id="code" type="text" name="code" class="block mt-1 w-1/2" inputmode="numeric" autofocus autocomplete="one-time-code"
-                            wire:model.defer="code"
-                            wire:keydown.enter="confirmTwoFactorAuthentication" />
+                            wire:model.defer="code" />
 
                         <x-jet-input-error for="code" class="mt-2" />
                     </div>


ManuelLeiner:
On the other hand there is already a password confirmation when the user clicks on the `Enable` button. Therefore in most use cases the additional password prompt for the confirmation code doesn't even pop up. Is it necessary at all?

Pressing enter to submit the confirmation code is a nice UX thing.

driesvints:
@ManuelLeiner the confirmation time can be changed to a smaller window. I know that's not much of an argument though. Maybe there's indeed no action to be taken here but if we go by your argument we could remove the password confirm buttons alltogether.

hailwood:
As commented on the original issue, 

What if rather than removing the keyup listener, if we still want to keep the password confirmation, we instead make the keyup listener just click on the button instead of invoking the confirmation directly?

driesvints:
@hailwood if we could do that, that'd be great. How would you tackle this?

hailwood:
Unfortunately I've never used livewire so can't help there,
With the Inertia stack the last version of Vue I used was v2, 
So I just did some quick reading and as vue doesn't have an equivalent of reacts forwardRef this is the best I could come up with, I'm not sure it's correct but it appears to work.

Add a ref on `<Button>` inside Button.vue, and then expose a click method - 
```vue
<script setup>
import {ref} from 'vue';

defineProps({
    type: {
        type: String,
        default: 'submit',
    },
});

const button = ref();

defineExpose({
  click: () => button.value.click(),
})

</script>

<template>
    <button ref="button" :type="type" class="...">
        <slot />
    </button>
</template>
```

In TwoFactorAuthenticationForm.vue
- Add the ref onto the confirm button (I'm calling it confirmButton here)
- add method that's just `confirmButton.value.click()`
- Update the @keyup.enter to call the new function

```vue
...
const confirmButton = ref(null);

...
const triggerConfirmCode = () => {
    confirmButton.value.click();
};

...
<JetInput
    id="code"
    v-model="confirmationForm.code"
    type="text"
    name="code"
    class="block mt-1 w-1/2"
    inputmode="numeric"
    autofocus
    autocomplete="one-time-code"
    @keyup.enter="triggerConfirmCode"
/>

...
<JetConfirmsPassword @confirmed="confirmTwoFactorAuthentication">
    <JetButton
        v-if="confirming"
        type="button"
        class="mr-3"
        :class="{ 'opacity-25': enabling }"
        :disabled="enabling"
        ref="confirmButton"
    >
        Confirm
    </JetButton>
</JetConfirmsPassword>
```

driesvints:
Thanks @hailwood 

@taylorotwell what path do you want to take? I think it might be best to pursue something @hailwood is proposing.

taylorotwell:
Can someone please provide clear, written recreation steps in a bullet list on this PR? Thanks.

ManuelLeiner:
@taylorotwell 

- set password timeout to 1 second `auth.password_timeout => 1`
- open profile
- click `Enable`
- confirm password in the modal
- enter a confirmation code in the input
- click on `Confirm`
- password modal pops up

If you repeat the steps but instead of clicking on `Confirm` you submit by pressing enter, the modal will not pop up.

ManuelLeiner:
> @ManuelLeiner the confirmation time can be changed to a smaller window. I know that's not much of an argument though. Maybe there's indeed no action to be taken here but if we go by your argument we could remove the password confirm buttons alltogether.

@driesvints sorry, I didn't see that my initial reply has already been discussed in the issue 😵‍💫 

One more thought:

If the password confirmation pops up after the user entered the confirmation code, the 30 seconds validity of the confirmation code might run out. It takes some time to type or copy from password manager.

To be honest I never saw an additional password check for a 2FA confirmation code in another application. Some of them even triggered the confirmation process as soon as I type the last digit of the code, without me clicking or pressing anything. However, removing that additional password prompt (after x minutes) could be a security issue - I am no security expert.

taylorotwell:
What if we remove password confirmation from enabling 2FA but preserve it for disabling 2FA?

ManuelLeiner:
> What if we remove password confirmation from enabling 2FA but preserve it for disabling 2FA?

@taylorotwell I just checked in a Gmail account to see how Google implemented the flow. As soon as I start activating, I have to enter my password first - even though I had logged into my account only a couple of seconds before. It might be a good security practice to verify by password, again, that you are the owner of the account.

taylorotwell:
I honestly don't think I'm going to take any action on this right now. It feels fairly edge case (based on lack of reports and complaining). Even if the password confirmation window was lowered to 10 minutes from the current default of 3 hours I think it is unlikely anyone would hit this. 

Of course, feel free to make the keyup change to your own published stubs if you really need a super low password confirmation time.

PR #1044, opened 2022-04-21
2.x hotfix: Add Messaging In Install For Edge Cases

ralphschindler:
When running `artisan jetstream:install livewire` on a project that had its `.env.example` already moved, it kills the install with the following message:

```
# artisan jetstream:install livewire
Migration created successfully.

   ErrorException

  file_get_contents(/app/.env.example): Failed to open stream: No such file or directory

  at vendor/laravel/jetstream/src/Console/InstallCommand.php:687
    683▕      * @return void
    684▕      */
    685▕     protected function replaceInFile($search, $replace, $path)
    686▕     {
  ➜ 687▕         file_put_contents($path, str_replace($search, $replace, file_get_contents($path)));
```

This PR solves that specific problem, but also adds messaging that alerts the installer that database sessions are preferable, and links to the manual with more information.

**Possible addition?** I think in an ideal world, any `replaceInFile` should have some output to call out what is being changed and why, to be more transparent.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -110,7 +110,7 @@ class InstallCommand extends Command
             }
         })->filter()->whenEmpty(function () {
             $this->info(<<<'END_OF_INFO'
-                \nJetstream attempted to update .env, .env.local, and .env.example to ensure the session driver is
+                Jetstream attempted to update .env, .env.local, and .env.example to ensure the session driver is
                 set to "database", but could not find one. Jetstream works best with database sessions, for more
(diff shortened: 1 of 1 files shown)


driesvints:
Jetstream is only intended to be installed on fresh Laravel projects, not existing ones.

driesvints:
Please see the docs.

Screenshot 2022-04-22 at 10 21 12

ralphschindler:
It was about 60 seconds new when I ran it, I generally delete `.env.example` and copy `.env` to `.env.local` in fresh projects. Not seeing the harm in in polishing the install experience, but I guess I see your point. 😅`¯\_(ツ)_/¯`

PR #1048, opened 2022-05-06
[2.x] Verification improvements

driesvints:
This PR improves the email verification process by allowing users to still manage their user profile even though the new email address is unverified. This will prevent users from accidentally changing their email address and being locked out from their account.

Here are some of the new UI updates for both Inertia and Livewire:

Screenshot 2022-05-06 at 14 36 50
Screenshot 2022-05-06 at 15 04 31
Screenshot 2022-05-06 at 15 04 39

### DIFF
--- a/stubs/livewire/resources/views/profile/update-profile-information-form.blade.php
+++ b/stubs/livewire/resources/views/profile/update-profile-information-form.blade.php
@@ -67,7 +67,7 @@
 
             @unless ($this->user->hasVerifiedEmail())
                 <p class="text-sm mt-2">
-                    Your email address is unverified.
+                    {{ __('Your email address is unverified.') }}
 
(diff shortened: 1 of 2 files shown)


stubs/livewire/resources/views/profile/update-profile-information-form.blade.php
```diff
@@ -64,6 +64,22 @@
...
+                    Your email address is unverified.
+
+                    <button type="button" class="underline text-sm text-gray-600 hover:text-gray-900" wire:click.prevent="sendEmailVerification">
+                        {{ __('Click here to re-send the verification email.') }}
+                    </button>
+                </p>
+
+                @if ($this->verificationLinkSent)
+                    <p v-show="verificationLinkSent" class="mt-2 font-medium text-sm text-green-600">
+                        A new verification link has been sent to the above email address you.
```
taylorotwell, line 79: This line isn't localized but the others are.

PR #1059, opened 2022-05-30
[2.x] Switch to ESM imports

jessarcher:
Carrying on from the PR at https://github.com/laravel/laravel/pull/5895, this replaces `require` statements with `import` statements where possible in the JS resources in order to smooth the transition for Vite users.

Note that with Inertia there is one remaining dynamic `require` that does not have a replacement that works with webpack _and_ Vite.


driesvints:
Thanks @jessarcher

PR #1060, opened 2022-05-30
[2.x] Simplify Tailwind installation

jessarcher:
This PR simplifies the Tailwind installation in Jetstream and is the counterpart of [this Breeze PR](https://github.com/laravel/breeze/pull/155):

* Uses the `@tailwind` directive instead of `@import`, per the standard [installation guide](https://tailwindcss.com/docs/guides/laravel).
* Removes the unnecessary `postcss-import` dependency and Mix configuration.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -142,7 +142,6 @@ class InstallCommand extends Command
                 '@tailwindcss/forms' => '^0.4.0',
                 '@tailwindcss/typography' => '^0.5.0',
                 'alpinejs' => '^3.0.6',
-                'postcss-import' => '^14.0.1',
                 'tailwindcss' => '^3.0.0',
             ] + $packages;
         });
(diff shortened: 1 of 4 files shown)


driesvints (approved):

NicoHood:
I am wondering how would I get those fixes, if I already installed jetstream? They seem to be minor for sure, but let's say I want to be in sync with the latest version, isn't there a simple way to upgrade all my files?

driesvints:
You're going to have to do the changes manually as Jetstream is a scaffold tool.

NicoHood:
That's a pitty. Isnt there a way to at least document those changes? Or is there a way to reinstall and check the changes via git and undo them, if the file was changed by myself (which is very unlikely). How can I ensure, that I don't miss critical updates?

satoved:
@NicoHood There are no critical updates, until you need a new feature.

It will continue working as it is in your project.

Generally, I tend to look over all releases of Jetstream with code diffs, and I take what I like.

E.g. one cool undocumented feature was 2FA confirmation.

NicoHood:
Could this method be at least documented somewhere? I guess you will do a diff any only take code from a special skeleton folder that is usually copied when installing. I understand that this is a manual process, but documenting this would help a lot.

PR #1065, opened 2022-06-09
Profile photo picker accepts jpg, jpeg and png only

Rakesh717:
In [backend](https://github.com/laravel/jetstream/blob/c25d17ddca6d197a8035a9bc160e716b7e0a07cf/stubs/app/Actions/Fortify/UpdateUserProfileInformation.php#L24) we are expecting image to be `jpg,jpeg,png` then why not do it on frontend too.


taylorotwell:
Feel free to add this to your templates if you wish.

Rakesh717:
Thank you for the reply.

I sure will do it but its good to have it out of the box. It would be nice UX to be able to just choose from allowed file than filtering from all other files.

PR #1067, opened 2022-06-11
Add .vue extension to imports

SamuelMwangiW:
This adds the .vue extension to match all other Vue files in the repository. 
It also plays nice with Vite.


jessarcher:
Looks good, thanks! :+1:

PR #1073, opened 2022-06-22
[3.x] Change $page.user to $page.auth.user

SamuelMwangiW:
This PR is proposed to nest the shared user prop under auth key as proposed in #323.

Since this is a breaking change, am targeting the next major release https://github.com/laravel/jetstream/pull/323#issuecomment-1162846960.

If approved, the docs and upgrade guide will also need to be updated.

No hard feelings on closure


jessarcher:
I haven't tested this yet, but I've bumped into the issue when building a `/users/{user}` page, so I'm happy to see it :+1:

SamuelMwangiW:
Thanks Taylor

PR #1076, opened 2022-06-23
bump @tailwindcss/forms

timacdonald:
Although not technically necessary as the version constraints allows for the latest `0.5.2` version and the latest allowed version would be installed on `npm install`, sending this PR through to match the `laravel/breeze` PR / versioning which we also bumped to this version to avoid a Vite console warning.

See: https://github.com/laravel/breeze/pull/161


timacdonald:
The test are failing as the Vite stuff has not yet been tagged in laravel/laravel.

PR #1078, opened 2022-06-27
Fix ziggy location instead of base url

matthieumota:
I've noticed that using the URL property to handle the active route on Ziggy in SSR is not correct. Indeed, we need to know the current URL and the URL property returns the base URL of the application.

To reproduce the bug, just disable the JS and find that the active route in the menu does not work. With JS, it works because the client re-hydrates the server-generated DOM.

So, I think we need to get current URL via request and totally replace window.location so.

### DIFF
--- a/stubs/inertia/app/Http/Middleware/HandleInertiaRequests.php
+++ b/stubs/inertia/app/Http/Middleware/HandleInertiaRequests.php
@@ -35,8 +35,10 @@ class HandleInertiaRequests extends Middleware
     public function share(Request $request)
     {
         return array_merge(parent::share($request), [
-            'ziggy' => function () {
-                return (new Ziggy)->toArray();
+            'ziggy' => function () use ($request) {
+                return array_merge((new Ziggy)->toArray(), [
+                    'location' => $request->url(),
+                ]);
             },
         ]);
     }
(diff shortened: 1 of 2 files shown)


prestonholt:
> This recently got worked on by @prestonholt at #1069. @prestonholt was this working for you before?

@driesvints I just tested and this wasn't working before #1069 either- this PR appears to fix the issue both pre and post #1069 changes.

I also tried fixing by following @aarondfrancis [method](https://aaronfrancis.com/2022/using-ziggy-with-inertia-server-side-rendering#sending-the-routes-over), but it isn't fixing the `route().current()` issue either. 

It seems like this PR fixes it because it keeps `ziggy.url` as the base URL, and also passes the current URL as `ziggy.location`, but I'm unsure whether `ziggy.url` needing to be the base URL is intended or a bug

timacdonald (approved):

timacdonald:
Regarding `ziggy.url` it seems to me that if you set this to anything other than the base url, it is added to every URL generation, thus I believe that "url" config option is synonymous with "baseUrl" in this context.

timacdonald:
I've opened a matching PR for breeze: https://github.com/laravel/breeze/pull/163

PR #1086, opened 2022-07-08
Solved Not to make more than one query on the database when the user …

michaelnabil230:
This PR is to minimize the queries. When the user update to profile or delete it.

Before:

Screen Shot 2022-07-08 at 12 43 53 AM

After:
Screen Shot 2022-07-08 at 12 57 38 AM


taylorotwell:

PR #1090, opened 2022-07-15
Create cors.php

MooseSaeed:
Currently, a user can't redirect to a subdomain route due to CORS errors when using inertiaJS. The only workaround is using `Inertia::location($url)` which causes a complete page refresh.

These changes are necessary to allow redirection to routes that contains subdomains and avoid CORS errors.

1- Added '*' to 'paths'

> Must be added to accept all user generated subdomain wildcards

2- Restricted allowed methods to only 'GET' requests.

> Can be changed to allow different methods according to the user needs.

3- Restricted allowed origins to the App URL

> Or it can be replaced with '*'

4- Exposed 'x-inertia' header

> This is necessary for Inertia to handle the request

### DIFF
--- a/config/cors.php
+++ b/config/cors.php
@@ -1,0 +1,34 @@
+<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Cross-Origin Resource Sharing (CORS) Configuration
+    |--------------------------------------------------------------------------
+    |
+    | Here you may configure your settings for cross-origin resource sharing
+    | or "CORS". This determines what cross-origin operations may execute
+    | in web browsers. You are free to adjust these settings as needed.
+    |
+    | To learn more: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
+    |
+    */
+
+    'paths' => ['api/*', '*', 'sanctum/csrf-cookie'],
+
+    'allowed_methods' => ['GET'],
+
+    'allowed_origins' => [env('APP_URL')],
+
+    'allowed_origins_patterns' => [],
+
+    'allowed_headers' => ['*'],
+
+    'exposed_headers' => ['x-inertia'],
+
+    'max_age' => 0,
+
+    'supports_credentials' => false,
+
+];


driesvints:
Is this related to https://github.com/laravel/framework/issues/43184 ?

MooseSaeed:
@driesvints This PR solves a different issue than the one mentioned in the bug report. This PR solves CORS errors mentioned [here](https://github.com/inertiajs/inertia/issues/99) when a user try to redirect to a subdomain route or external route.

To explain further, redirecting to a subdomain route with InertiaJS has 2 issues:
1- CORS errors that prevents the user from completing the redirection.
2- After those CORS errors are gone, the user is redirected to the desired page, BUT the url showing in the browser is incorrect which is explained in the bug report and bug-report repo.

This PR solves the first issue only. The bug report is reporting the second issue only.

Making those changes in this PR allows the user to get redirected without CORS errors. The bug report https://github.com/laravel/framework/issues/43184 is that after this CORS errors is resolved by those changes and the redirection is successful to a working page, the url in the browser shows the incorrect url.

config/cors.php
```diff
@@ -0,0 +1,34 @@
...
+    |
+    | Here you may configure your settings for cross-origin resource sharing
+    | or "CORS". This determines what cross-origin operations may execute
+    | in web browsers. You are free to adjust these settings as needed.
+    |
+    | To learn more: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
+    |
+    */
+
+    'paths' => ['api/*', '*', 'sanctum/csrf-cookie'],
```
driesvints, line 18: This doesn't makes sense. You're saying "all api paths", "all paths" & "the sanctum csrf cookie path".

driesvints: Isn't the issue that you aren't adding your subdomains to the `allowed_origins` list?

taylorotwell:

MooseSaeed:
@taylorotwell I received a notification via email that the pull request merging will be delayed and that If I think this corrects a bug then I should mention you, but I can't seem to find this response here, so I thought of explaining further what this PR solves anyway.

Redirection from the root domain to subdomain routes isn't working while using Laravel & Jetstream InertiaJS stack. Whenever you try to get redirect from root domain to subdomain route you will get the below CORS errors in browser and it won't allow the redirection to proceed.

```
Access to XMLHttpRequest at 'http://user.example.test/my-first-product' (redirected from 'http://example.test/products/create') from origin 'http://example.test' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

In a fresh installed Laravel and Jetstream Inertia app, this is how cors.php looks like by default:

```php
    'paths' => ['api/*', 'sanctum/csrf-cookie'],

    'allowed_methods' => ['*'],

    'allowed_origins' => ['*'],

    'allowed_origins_patterns' => [],

    'allowed_headers' => ['*'],

    'exposed_headers' => [],

    'max_age' => 0,

    'supports_credentials' => false,
```

Making the below changes to cors.php makes those CORS errors disappear and the redirection to proceed and the desired view to open: 

```php

// Adding '*' to 'paths'
    'paths' => ['*'],
// Adding 'x-inertia' to 'exposed_headers'
    'exposed_headers' => ['x-inertia'],
```

After those changes, a user can be redirected to the desired page without CORS errors. But then, they will face another issue which is described [here](https://github.com/laravel/framework/issues/43184). This PR solved the first issue which is CORS errors and makes a successful redirection to the desired page.

To summarize, by adding `'*'` to paths and exposing `'x-inertia'` header, CORS errors won't show or stop the redirection process.

taylorotwell:
Hmm, I'm not quite sure on this for now so for now why don't you just edit the `cors` configuration file in your own application. Thanks!

PR #1102, opened 2022-07-29
[2.x] Adds Expired At Input To Api Token Create

tcampbPPU:
With the latest release of [Sanctum v3](https://github.com/laravel/sanctum/releases/tag/v3.0.0) an expires_at field was added. So this PR adds that additional input so when users are creating tokens they can set that optional field if needed

Here is an example of what the additional input looks like in the Interia/Vue stack
image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     },
     "require-dev": {
         "inertiajs/inertia-laravel": "^0.5.2",
-        "laravel/sanctum": "^2.7",
+        "laravel/sanctum": "^3.0",
         "mockery/mockery": "^1.0",
         "orchestra/testbench": "^7.0",
         "phpunit/phpunit": "^9.3"
--- a/src/Http/Controllers/Inertia/ApiTokenController.php
+++ b/src/Http/Controllers/Inertia/ApiTokenController.php
@@ -4,6 +4,7 @@ namespace Laravel\Jetstream\Http\Controllers\Inertia;
 
 use Illuminate\Http\Request;
 use Illuminate\Routing\Controller;
+use Illuminate\Support\Carbon;
 use Laravel\Jetstream\Jetstream;
 
 class ApiTokenController extends Controller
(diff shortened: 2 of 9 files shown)


tcampbPPU:
Sorry about the other PR I created `https://github.com/laravel/jetstream/pull/1101` I didn't realize my fork was out of sync by a lot 😅

tcampbPPU:
The reason why the newest tests are failing, are due to the following, the step
```
composer create-project laravel/laravel:^9 .
```
from `.github/workflows/tests.yml`

still pulls in the older version of Sanctum
```
"laravel/sanctum": "^2.14.1",
```
if this was pulling in the latest version, it would work fine now that the column would exist.

taylorotwell:
Going to leave them as non-expiring by default.

tcampbPPU:
They would be non-expiring by default unless you set the input

tcampbPPU:
@taylorotwell I get If you do not want to add the inputs to the FE to support the latest version 3 of sanctum since that UI might not work for everyone, but shouldn't Jetstream's BE support this updated feature of sanctum? For these files:
- `src/Http/Controllers/Inertia/ApiTokenController.php`
- `src/Http/Livewire/ApiTokenManager.php`

Especially when using the Inertia stack you cannot configure it to use a custom `ApiTokenController` unless you copy that file and routes and make the changes you want, which seems like a lot of extra work needed to add functionality from an existing Laravel package that it's a dependent of.

To clarify more.. If you need to use a different model with Jetstream you would somthing like so
> JetstreamServiceProvider.php
```php
public function boot()
{
  Jetstream::useUserModel(User::class);
}
```
There is not a similar method I am aware of for the ApiTokenController.
```php
public function boot()
{
  Jetstream::useApiTokenController(ApiTokenController::class);
}
```
I must copy all of the routes and controller it uses in my project to add to support 1 extra column that was added to Sanctum.

PR #1103, opened 2022-07-31
[2.x] Fixes pest-plugin's install

mozex:
As of Composer 2.2.0, the allow-plugins option adds a layer of security allowing you to restrict which Composer plugins are able to execute code during a Composer run. ([Read More](https://getcomposer.org/doc/06-config.md#allow-plugins))
This produces an error with the pest-plugin composer package, running `php artisan jetstream:install livewire --pest` throws this error:
`pestphp/pest-plugin contains a Composer plugin which is blocked by your allow-plugins config. You may add it to the list if you consider it safe.`
So in order to fix this in the install command, I added a feature to run `composer config allow-plugins.pestphp/pest-plugin true` right before installing Pest packages.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -81,6 +81,7 @@ class InstallCommand extends Command
         $stubs = $this->getTestStubsPath();
 
         if ($this->option('pest')) {
+            $this->addPestToAllowedPlugins();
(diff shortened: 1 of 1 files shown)


driesvints:
Here's a more simpler alternative: https://github.com/laravel/laravel/pull/5959

taylorotwell:
Closed in favor of: https://github.com/laravel/laravel/pull/5959

PR #1108, opened 2022-08-04
[2.x] Allow optional Expires At Token Field To Be Passed

tcampbPPU:
Allow Inertia And Livewire BE to accept a `expires_at` field it needed and passed to work with the latest version 3 of Sanctum.

This PR is similar to the [PR](https://github.com/laravel/jetstream/pull/1102) I submitted earlier that was closed, but this time it does not implement any UI changes.

The reason why I re-submitted this PR is because in order to implement using this new feature from V3 of sanctum I have to copy all the provided Jetstream api token routes and `ApiTokenController` (there might be additional or less work needed with liveware but idk I don't use that) just to include the addition of a simple singular field, if it was needed, since by default all tokens need to expire. This is just updating the functionality of if you need those tokens to expire.

[Here in this comment](https://github.com/laravel/jetstream/pull/1102#issuecomment-1199895334) I describe in some more details why this would be very tedious to do, but to summarize unlike in the `JetstreamServiceProvider` if I want to use a different model I can do so simply with something like `Jetstream::useUserModel(User::class);` 

But that does not exist for the `ApiTokenController` I must use what it offers, or if I need to make 1 change I must copy all the routes it uses and handle in my own version of `ApiTokenController`. And in this case the 1 change I need to make is to include passing an optional field that a dependency package of this uses.

A note why the updated tests are not passing is because `composer create-project laravel/laravel:^9 .` does not use the latest version of sanctum, more info about that [here](https://github.com/laravel/jetstream/pull/1102#issuecomment-1199721151) locally the test are all passing when running `./vendor/bin/phpunit`

### DIFF
--- a/src/Http/Controllers/Inertia/ApiTokenController.php
+++ b/src/Http/Controllers/Inertia/ApiTokenController.php
@@ -4,6 +4,7 @@ namespace Laravel\Jetstream\Http\Controllers\Inertia;
 
 use Illuminate\Http\Request;
 use Illuminate\Routing\Controller;
+use Illuminate\Support\Carbon;
 use Laravel\Jetstream\Jetstream;
 
 class ApiTokenController extends Controller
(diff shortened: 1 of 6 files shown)


driesvints:
> A note why the updated tests are not passing is because composer create-project laravel/laravel:^9 . does not use the latest version of sanctum,

That's not true. See here: https://github.com/laravel/jetstream/runs/7665643723?check_suite_focus=true#step:3:199

There's a different reason why tests aren't passing anymore. Right now, we cannot provide support to look into that. Please re-attempt this once tests are passing. You can try by pushing to your fork.

PR #1109, opened 2022-08-04
Update install command output to display vite build command

RhysLees:
This PR changes `Please execute the [npm install && npm run dev] commands to build your assets.`
to `Please execute the [npm install && npm run build] commands to build your assets.` to reflect the move to vite.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -233,7 +233,7 @@ class InstallCommand extends Command
 
         $this->line('');
         $this->components->info('Livewire scaffolding installed successfully.');
-        $this->components->warn('Please execute the [npm install && npm run dev] commands to build your assets.');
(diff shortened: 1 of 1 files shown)


jessarcher:
I'm not suggesting you change this PR as I don't know Taylor's thoughts, but the laravel installer (also used by laravel.build) [automatically runs](https://github.com/laravel/installer/blob/master/src/NewCommand.php#L180) `npm install && npm run build` on the Jetstream install. I wonder whether that should be moved to this repo so that it happens consistently no matter how you install it.

We could then mention both options so the user knows their options going forward:

```
Run [npm run build] to re-build your assets, or [npm run dev] to start the hot module replacement server.
```

taylorotwell:
@jessarcher I would be open to some improvements as you mention.

PR #1110, opened 2022-08-08
[2.x] Move Jetstream components to Components directory

jessarcher:
This PR renames the `resources/js/Jetstream` directory to `resources/js/Components`.

This signals to the user that the components belong to them and can be used outside the Jetstream pages and customized as needed. I've seen repos where people create their own `Components` directory and maintain two sets of buttons, etc.

This is also consistent with Breeze.


driesvints (approved):

joelbutcher:
@jessarcher Would it also be worth renaming the imports of these components to equally detach the perceived ownership from Jetstream? e.g `import JetButton from '@/components/Button.vue` to `import Button from '@/components/Button.Vue`?

jessarcher:
@joelbutcher I'd love to, but the [Vue style guide](https://vuejs.org/style-guide/rules-essential.html#use-multi-word-component-names) says not to.

Yesterday [I asked](https://twitter.com/jessarchercodes/status/1556493498118836224) Evan You whether he'd still advocate against it and this was his response:

> Not as strictly, but I’d still avoid names that overlap with known native elements.

So for components like `Modal` I think we should be fine. For things like `Button`, maybe we figure out some more descriptive prefixes for them. E.g. `PrimaryButton`, `SecondaryButton`. `Input` and `Label` might be `TextInput` and `InputLabel` or something.

PR #1111, opened 2022-08-08
Validate the stack argument

jessarcher:
Currently when running `jetstream:install`, if you specify an invalid stack such as "vue", you'll end up with a partial Jetstream install and no error message.

This PR validates that the stack is either `inertia` or `livewire`.

Note that there is no default stack in Jetstream and it already displays an error when the stack is not provided.

Ultimately I'd love to prompt the user if they don't specify a stack, but that raises bigger questions of whether it should prompt for options like `teams` etc.


driesvints:
Thanks!

PR #1115, opened 2022-08-09
Display validation errors alongside the fields [Inertia]

weavdale:
Update the Inertia stack to match the Breeze Vue components.

See @jessarcher pull request in Breeze https://github.com/laravel/breeze/pull/175


SamuelMwangiW:
My 0.0002 cents: @weavdale kindly revert the formatting changes.

weavdale:
> My 0.0002 cents: @weavdale kindly revert the formatting changes.

I tried to figure out the coding style Laravel is using for Javascript, but didn't want to spend a lot of time on it since the docs say StyleCI will fix it during merge. https://laravel.com/docs/9.x/contributions#styleci

stubs/inertia/resources/js/Pages/Auth/TwoFactorChallenge.vue
```diff
@@ -82,6 +81,7 @@ const submit = () => {
                     class="mt-1 block w-full"
                     autocomplete="one-time-code"
                 />
+                <JetInputError class="mt-2" :message="form.errors.code" />
```
jessarcher, line 84: It's a little unfortunate that this needs to be `code` rather than `recovery_code`, but it looks like Fortify puts the error here even when it's the recovery code that was attempted. The only time an error will appear under `recovery_code` is when the provided recovery code is not a string, and I don't think can ever happen in normal usage of the app.

jessarcher (approved):

stubs/inertia/resources/js/Pages/Auth/TwoFactorChallenge.vue
jessarcher: I've created a PR for Fortify to fix this: https://github.com/laravel/fortify/pull/401

If that PR is accepted, this will need to change to:

```suggestion
                <JetInputError class="mt-2" :message="form.errors.recovery_code" />
```


If that PR is not accepted, it might be a good idea to change this to something like this (untested):

```suggestion
                <JetInputError class="mt-2" :message="form.errors.recovery_code ?? form.errors.code" />
```

taylorotwell:
Will let @jessarcher make any PRs here if needed - we are doing some refactoring on Jetstream.

PR #1117, opened 2022-08-10
[2.x] Remove "Jet" component prefix

jessarcher:
Continuing from https://github.com/laravel/jetstream/pull/1110, this PR removes the "Jet" prefix from the component imports in the Vue stack to give users a better sense of ownership over the scaffolding.

This meant that some components were now a single word, which goes against the [Vue style guide](https://vuejs.org/style-guide/rules-essential.html#use-multi-word-component-names). I [asked](https://twitter.com/jessarchercodes/status/1556493498118836224) Evan You whether he'd still advocate against this, and [he responded](https://twitter.com/youyuxi/status/1556660031197548545):

> Not as strictly, but I’d still avoid names that overlap with known native elements.

After the rename, there were three components that overlapped with native elements, which I've renamed as follows:

* `Button` -> `PrimaryButton` (this creates symmetry with the `SecondaryButton` and `DangerButton` components)
* `Input` -> `TextInput`
* `Label` -> `InputLabel` (this creates symmetry with the `InputError` component)

If this is accepted, I will update the references in the Jetstream docs.


driesvints:
> This meant that some components were now a single word, which goes against the [Vue style guide](https://vuejs.org/style-guide/rules-essential.html#use-multi-word-component-names).

I feel like this is a weird rule as it's still obvious that these are components rather than native elements. But as long as we don't change the Livewire side of things, all good to me 😅

jessarcher:
Yeah, I agree it's a little weird. I guess because with Vue you can use kebab-case or PascalCase, and if you chose kebab-case then the single-word components could have identical names to native elements. Personally, I don't mind this and just disable that particular rule when using Vue's eslint prefix.

I wish it wasn't breaking to change the Livewire components, otherwise, I'd love to make them match.

chrysanthos:
Won't this just make upgrading harder for little to no benefit?

driesvints:
@chrysanthos there isn't really an upgrade path for Jetstream. It's a scaffold tool.

chrysanthos:
> @chrysanthos there isn't really an upgrade path for Jetstream. It's a scaffold tool.

I know, and the way to go about upgrading is to manually monitor the releases for changes. This PR will make this harder for no benefit.

driesvints:
You're still talking about "upgrading" but there really isn't any such thing.

PR #1118, opened 2022-08-10
[3.x] Copy Blade files on installation

jessarcher:
This PR updates Jetstream so that it copies the Livewire components on installation to make it more like the Vue stack, and more like Breeze.

I've placed them in `resources/views/components` instead of `resources/views/vendor/jetstream/components` where they are currently published, to give the user a better sense of ownership over the components, and to leverage Laravel's [anonymous components](https://laravel.com/docs/9.x/blade#anonymous-components). This also means the `jet-` prefix is no longer used, but I think that's a good thing.

Finally, I've also made it copy the email views on installation for both stacks so there are now no resources that aren't copied on installation. I've placed them in `resources/views/emails` rather than `resources/views/vendor/jetstream/mail` for the reasons previously mentioned. I've used `emails` instead of `mail` to match the Laravel docs ([here](https://laravel.com/docs/9.x/mail#configuring-the-view) and [here](https://laravel.com/docs/9.x/mail#generating-markdown-mailables)).

I've targetted `master` because I believe it is a breaking change to remove the ability for existing Jetstream users to publish the views.


taylorotwell:
@jessarcher thanks - will we need any docs changes here?

jessarcher:
@taylorotwell indeed! I have created a draft PR at https://github.com/laravel/jetstream-docs/pull/70 to add docs for `3.x`

amihailov:
No offense but the prefix for jetstream components was very handy since it is package specific. Whitout prefix, given that they are very generically named, just copying them in the `resources/views/components` potentially overwrites any user / project components which were already there. It's just happened to me... So now either I move all of my own under new namespace/folder or move again all of the Jetstream components...

PR #1119, opened 2022-08-10
[2.x] Install NPM dependencies and build assets

jessarcher:
This PR updates the Jetstream installer to automatically run the `npm install` and `npm run build` commands for you.

Currently, the [Laravel installer](https://github.com/laravel/installer/blob/df70a9391dbd5f911565312f8afc125a5e86e0d9/src/NewCommand.php#L173) automatically runs these commands when it installs Jetstream. Moving it to this package makes the installation more consistent no matter how you install it. I have created https://github.com/laravel/installer/pull/242 to remove these from the Laravel installer.

* [x] Tested on Linux
* [x] Tested on Mac
* [x] Tested on Windows

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -722,7 +722,7 @@ EOF;
      * Run the given commands.
      *
      * @param  array  $commands
-     * @return voidSymfony\Component\Process\Process
+     * @return void
      */
     protected function runCommands($commands)
     {
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -6,6 +6,7 @@ use Exception;
 use Illuminate\Console\Command;
 use Illuminate\Filesystem\Filesystem;
 use Illuminate\Support\Str;
+use RuntimeException;
 use Symfony\Component\Process\PhpExecutableFinder;
 use Symfony\Component\Process\Process;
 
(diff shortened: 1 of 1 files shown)


joelbutcher:
Love this! I'm currently doing this for users of Socialstream 🎉

joelbutcher:
@jessarcher ohh just had a thought - with this addition to Jetstream, users of my package would now need to wait twice whilst the frontend is compiled... what are your thoughts on making this option configurable for other package maintainers that rely on Jetstream as a basis for their packages?

timacdonald:
I have just tested this on my Mac, and the NPM install process worked as expected and had the colourised output as well 🤩

jessarcher:
Hey @joelbutcher, how much of an impact do you expect this to have?

I don't see an issue with `npm install` running twice, as the second time it should be very quick to realise nothing needs to be done. You could probably even remove this from your package if you aren't changing `package.json` at all.

As for the build, it's a bit unfortunate that it would run twice, but the build is very fast with Vite.

I'm hesitant to add another option to the install command, but if this PR is accepted, you're of course welcome to open a PR for it.

joelbutcher:
@jessarcher The impact probably wouldn't matter that much. I'm not updating the package.json AFAIK (will check later). The build time with Vite is much less of a concern actually, I hadn't considered that tbh

PR #1120, opened 2022-08-12
Check current model supports HasTeams before trying to access team information.

crynobone:
This would be useful for applications with multiple user providers and uses Inertia (as `ShareInertiaData` middleware is appended to global web middleware). As an example, this will fix laravel/nova-issues#4683

Signed-off-by: Mior Muhammad Zaki <crynobone@gmail.com>

### DIFF
--- a/src/Http/Middleware/ShareInertiaData.php
+++ b/src/Http/Middleware/ShareInertiaData.php
@@ -19,12 +19,14 @@ class ShareInertiaData
      */
     public function handle($request, $next)
     {
+        $user = $request->user();
+
         Inertia::share(array_filter([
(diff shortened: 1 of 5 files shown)


taylorotwell:
Do you have to resolve the user outside of the closures? I think it changes the time the user is resolved and may introduce some sort of bugs?

crynobone:
> Do you have to resolve the user outside of the closures?

Not really.

> I think it changes the time the user is resolved and may introduce some sort of bugs?

Most of the time it would force a page reload but I reverted that changes since it really trivial to the original changes for this PR.

PR #1131, opened 2022-09-05
Check feature traits when deleting a user

rinodrummer:
Recently I created a new project using Jetstream and disabled profile photo and API features, not needing them, I removed the related traits too.

Removing them caused `DeleteAccountTest` to fail because the methods where not defined, so I added a check to verify if traits are recursively used in the user class.


rinodrummer:
I would like to ask if another solution could be considered:

Instead of repeating the user deletion logic in `DeleteUser` and `DeleteUserWithTeams`, precisely the following lines

```php
if (in_array(HasProfilePhoto::class, $userModelTraits)) {
    $user->deleteProfilePhoto();
}

if (in_array(HasApiTokens::class, $userModelTraits)) {
    $user->tokens->each->delete();
}

$user->delete();
```

In `DeleteUserWithTeams` it could be replaced with:

```php
$userDeleteAction = new DeleteUser();
$userDeleteAction->delete($user);
```

taylorotwell:
You should just update your code if you remove those features.

PR #1140, opened 2022-09-16
fix error 'Ziggy' is not defined

gerardroche:
For example when running eslint your will get a 'Ziggy' undefined error:

  ..../resources/js/app.js
  16:22  error  'Ziggy' is not defined  no-undef

I'm not super familiar with this code but I assume it was added by mistake, see d33e331ffc7da6c69bdd9452b1b7075e24f0557d

### DIFF
--- a/stubs/inertia/resources/js/app.js
+++ b/stubs/inertia/resources/js/app.js
@@ -15,7 +15,7 @@ createInertiaApp({
     setup({ el, app, props, plugin }) {
         return createApp({ render: () => h(app, props) })
             .use(plugin)
-            .use(ZiggyVue, Ziggy)
+            .use(ZiggyVue)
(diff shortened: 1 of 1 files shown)


taylorotwell:
I get no such error in my browser?

gerardroche:
fyi The Ziggy variable was uses a global Ziggy variable output by the `@routes` blade directive. If the `@routes` is not in place you get an error. It seems passing Ziggy to `use()` is not required if using `@routes` blade directive. 

>    Note: If you use the `@routes` Blade directive in your views, Ziggy's configuration will already be available globally, so you don't need to import the `Ziggy` config object and pass it into `use()`. -- https://github.com/tighten/ziggy#vue

PR #1147, opened 2022-09-22
Make .defer optional on @entangle for x-jet-modal

selfsimilar:
There are times when I would like to take an action when a user dismisses the modal window via `esc` or a click outside. This is a backwards-compatible and opt-in change to allow for this. Just call e.g. `<x-jet-modal :deferEntangle="false" wire:model="showModalFlag">`


taylorotwell:
You can add this to your own templates if you wish.

PR #1148, opened 2022-09-25
[2.x] Only update user model if two factor confirmation has truly started

hopkins385:
The Inertia stack is using a session var based workflow to determine if the user has started confirming two factor auth activation. If the user aborts confirming the two factor activation, for example by reloading the page, the two factor model-attributes are being reset accordingly. Such as two_factor_secret and others, which is working fine in the current implementation. 

But, **on every /user/profile visit or refresh** the two factor model-attributes are reset causing unnecessary database updates, and also causing for example re-indexing of the whole model when Laravel scout is used, as long as two factor is not activated. This is because the current implementation does not check if the two factor confirmation has truly started.

With this PR the session variable 'two_factor_confirming_at' is being checked if present in the 'neverFinishedConfirmingTwoFactorAuthentication' method, resulting two factor model-attributes resets only if the two factor auth activation has been started by the user. If the user has not started the two factor auth activation, then the model-attributes are not reset.

The livewire stack is not using a session based workflow which is the reason why there are no updates and test required.


driesvints:
@hopkins385 can you fix the tests here?

driesvints:
It's because previously, `two_factor_confirming_at` wasn't required to be set but now it is. The `UserProfileController` uses the `DisableTwoFactorAuthentication` action.

I'm not 100% sure but it seems this change would directly conflict with the test here: https://github.com/laravel/jetstream/blob/2.x/tests/UserProfileControllerTest.php#L21 because `two_factor_confirming_at` isn't set.

hopkins385:
But the failure is saying it’s missing the invoke call, and that’s what confuses me.
To me the error looks like as if somewhere else is an issue, or do I miss something?

`UserProfileControllerTest::test_empty_two_factor_state_is_noted
[19](https://github.com/laravel/jetstream/actions/runs/3122592574/jobs/5064630244#step:5:20)
Mockery\Exception\InvalidCountException: Method __invoke(<Any Arguments>) from Mockery_4_Laravel_Fortify_Actions_DisableTwoFactorAuthentication should be called
[20](https://github.com/laravel/jetstream/actions/runs/3122592574/jobs/5064630244#step:5:21)
 exactly 1 times but called 0 times.`

driesvints:
Trait is added here:

https://github.com/laravel/jetstream/blob/68b9c9a92a80a1f21b1c3deffb317159318d3fc2/src/Http/Controllers/Inertia/UserProfileController.php#L17

This check is invoked:

https://github.com/laravel/jetstream/blob/68b9c9a92a80a1f21b1c3deffb317159318d3fc2/src/Http/Controllers/Inertia/Concerns/ConfirmsTwoFactorAuthentication.php#L37

Which now returns false because of your changes since `two_factor_confirming_at` isn't set when executing the below test:

https://github.com/laravel/jetstream/blob/68b9c9a92a80a1f21b1c3deffb317159318d3fc2/tests/UserProfileControllerTest.php#L21

So your change is triggering a change in behavior. It's the exact issue you're trying to solve. We'll need to adjust the above test to accommodate for this change.

hopkins385:
Ok, thanks for pointing that out. I guess I need to dig a bit deeper.

hopkins385:
Ok, now here is what I understand.

The test 'test_empty_two_factor_state_is_noted' checks wether the __invoke method of the DisableTwoFactorAuthentication::class was called or not. 

The previous test was asserting that the __invoke must be called.

But that is essentially what we want to avoid, and what my PR changes.
We do not want to call the __invoke (and with that update the user-model) each and every time if there hasn't been started the two factor confiming process. Instead we just want to call the __invoke method if the two factor confirming process was recently started.

So I changed the Test from 

`$disable->shouldReceive('__invoke')->once();`
to 
`$disable->shouldNotReceive('__invoke');`

Ans the other tests are still checking if the __invoke was called, once the confirm process has been started, but somehow aborted. And they are all still passing.

@driesvints What do you think?

hopkins385:
Cool. I tried to improve the formatting by using Laravel Pint.

taylorotwell:
Unable to recreate this.

I abort the 2FA process and am logging every DB query in ray().

I see these queries on refresh and subsequent visits to the user profile screen. Note I do not see the users table being updated as you state:

CleanShot 2022-09-29 at 12 44 09@2x

hopkins385:
Scout is doing a re-index job on every profile reload, because the user model is updated. You can see this in the horizon dash, or in the shell.

taylorotwell:
In the queries I posted the user is not being updated?

hopkins385:
The Job 'Laravel\Scout\Jobs\MakeSearchable' is fired every time the '/user/profile' route is reloaded. That is because the method `neverFinishedConfirmingTwoFactorAuthentication` returns always true as long as 2FA is not activated, resulting to fire `app(DisableTwoFactorAuthentication::class)(Auth::user());` on every refresh. Resulting to execute 

```
class DisableTwoFactorAuthentication
{
    /**
     * Disable two factor authentication for the user.
     *
     * @param  mixed  $user
     * @return void
     */
    public function __invoke($user)
    {
        $user->forceFill([
            'two_factor_secret' => null,
            'two_factor_recovery_codes' => null,
        ] + (Fortify::confirmsTwoFactorAuthentication() ? [
            'two_factor_confirmed_at' => null,
        ] : []))->save();

        TwoFactorAuthenticationDisabled::dispatch($user);
    }
}
```

I am not sure if we can see database updates of a dispatched Job in Ray or Telescope. Anyways, the code is clearly executing a `->save()` operation.

driesvints:
@hopkins385 there still seems to be something different between your installation and a default setup of jetstream. Maybe try creating a repo that reproduces this?

```bash
laravel new bug-report --github="--public"
```

hopkins385:
Ok, will check that.

hopkins385:
There you go: https://github.com/hopkins385/bug-report.git

1. Use env.example as your main .env (using redis as cache, queue and session driver)
2. Open console and run php artisan horizon
3. Register new user and login. 
4. Visit /user/profile
5. Reload /user/profile (as many times as u like)
6. Check console output after every reload of /user/profile

```
INFO  Processing jobs from the [default] queue.

  2022-09-30 10:58:36 Laravel\Scout\Jobs\MakeSearchable ......... 49.76ms DONE
  2022-09-30 10:58:39 Laravel\Scout\Jobs\MakeSearchable ......... 34.98ms DONE
```

driesvints:
With this repo, when I try to register a new user, I get a `419 PAGE EXPIRED` error every time I attempt registration. You'll need to fix that or provide more clear instructions on how to set up the repo.

hopkins385:
You need to link via Valet that repo and open in ur browser http://bug-report.test as this is the APP_URL

driesvints:
Thanks. I reverted it to use artisan serve since that doesn't matter.

@taylorotwell I can confirm the bug with the above repo. Scout executes every time I reload the /user/profile page.

taylorotwell:
Fixed here: https://github.com/laravel/fortify/commit/9a68cf2deb37d1796b6e2fd97d3c61f086868914

hopkins385:
Perfect.

driesvints:
Thanks @taylorotwell

PR #1153, opened 2022-09-29
Remove empty properties when submitting 2FA form with Inertia stack

### LINKED ISSUE #1152: 2FA form not getting correct errors - Inertia stack
- Jetstream Version: 2.12.3
- Jetstream Stack: Inertia
- Uses Teams: Does not matter
- Laravel Version: 9.32.0 (Does not matter)
- PHP Version: 8.1.10 (Does not matter)
- Database Driver & Version: mysql 8.0.30 (Does not matter)

### Description:
When using the Inertia stack with Jetstream, there is a bug when submitting the two factor challenge form. Since the form is initialized with both code and recovery_code, it will always submit both values to the server. In the Fortify response class, `Laravel\Fortify\Http\Responses\FailedTwoFactorLoginResponse.php` there is a check that the request has the recovery_code key and does not check if it has a value. It then is setting the invalid field and message accordingly. Because of this check, the inertia stack causes fortify to think that we are using the recovery code to login instead of the standard code.

Other Note:
Livewire stack submits this correctly. 

### Steps To Reproduce:

1. `Laravel new jetstream-test --jet --stack=inertia`
2. Finish install
3. Setup a user with 2FA enabled
4. Log out
5. Log back in
6. Enter an invalid 2FA code
7. No message will display

likeadeckofcards:
When using the Inertia stack with Jetstream, there is a bug when submitting the two factor challenge form. Since the form is initialized with both code and recovery_code, it will always submit both values to the server. In the Fortify response class, `Laravel\Fortify\Http\Responses\FailedTwoFactorLoginResponse.php` there is a check that the request has the recovery_code key and does not check if it has a value. It then is setting the invalid field and message accordingly. Because of this check, the inertia stack causes fortify to think that we are using the recovery code to login instead of the standard code.

This PR is one of the possible solutions that fixes #1152 and does not change the current functionality of the Fortify library. This change will remove any empty properties from the form payload before submitting it to the server for processing.

### DIFF
--- a/stubs/inertia/resources/js/Pages/Auth/TwoFactorChallenge.vue
+++ b/stubs/inertia/resources/js/Pages/Auth/TwoFactorChallenge.vue
@@ -1,4 +1,5 @@
 <script setup>
+import {pickBy, property} from 'lodash';
 import { nextTick, ref } from 'vue';
 import { Head, useForm } from '@inertiajs/inertia-vue3';
 import AuthenticationCard from '@/Components/AuthenticationCard.vue';
@@ -33,7 +34,7 @@ const toggleRecovery = async () => {
 };
 
 const submit = () => {
-    form.post(route('two-factor.login'));
+    form.transform((data) => pickBy(data, property)).post(route('two-factor.login'));
 };
 </script>
 


taylorotwell:
Please share screenshots of the problem the existing situation causes on Jetstream.

PR #1156, opened 2022-10-07
fixes and improvements

A-w-x:


driesvints:
We'll need reasoning for all of these and it's probably best that these are sent as separate PRs

A-w-x:
reasoning is in commit message

PR #1157, opened 2022-10-07
Do no blindly read files in UpdateUserProfileInformation

A-w-x:
there is no check, it will just read whatever the user selected from browser and this is not only wrong but it also may crash the browser tab or just display broken preview since the file is not an image

crashes may occur when, for example, a big mp4 video is selected

with these changes an error is shown when big not valid files are selected and no changes are made to the photo preview if the file is not valid (as you would expect)

also use mimetypes in validation for consistency and to use the list already defined by laravel (eg. the jpeg mimetype lists one more type than the ones in the previous validation list)


taylorotwell:
Feel free to adjust your own application.

A-w-x:
lol, thought framework or starter kit were supposed to be working by default, but..fine :)

PR #1159, opened 2022-10-07
Add default min validation for user name

A-w-x:
nit, for completeness, just a default provided by the kit


taylorotwell:

PR #1160, opened 2022-10-08
Add jsconfig.json to Inertia stack

claytonfaria:
PhpStorm was showing errors in the imports of the components, then I noticed the `jsconfig.json` was missing and adding it solve it. The `jsconfig.json` file is the same as the Breeze inertia Vue stack

Before
Screenshot 2022-10-08 at 3 03 50 PM

After
Screenshot 2022-10-08 at 3 07 26 PM


taylorotwell:
Your PR never actually installs the stub?

PR #1162, opened 2022-10-14
Fixed tests that fail when Eloquent's "strict mode" is enabled.

ash-jc-allen:
Hey! This PR makes some really small updates to the tests that are copied into fresh applications when installing Jetstream.

With the new "strict mode" changes in Eloquent, if you enable it in your application, some of the tests will fail because certain attributes don't exist on the model. All I'm doing in the tests is grabbing fresh versions of the models that were causing the errors so that the fields are available.

I'm not entirely sure if what I've done is correct, and if I might be covering up an underlying issue in the new "strict mode" code. If so, apologies, feel free to close this PR! 😄

### DIFF
--- a/stubs/tests/EmailVerificationTest.php
+++ b/stubs/tests/EmailVerificationTest.php
@@ -21,7 +21,7 @@ class EmailVerificationTest extends TestCase
             return $this->markTestSkipped('Email verification not enabled.');
         }
 
-        $user = User::factory()->withPersonalTeam()->unverified()->create();
+        $user = User::factory()->withPersonalTeam()->unverified()->create()->fresh();
 
         $response = $this->actingAs($user)->get('/email/verify');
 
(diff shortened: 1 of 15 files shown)


taylorotwell:
@ash-jc-allen I think the "real" way to fix this would be to put `current_team_id` in the `UserFactory`?

ash-jc-allen:
@taylorotwell Yeah, I had a feeling that this might not have been the right approach to fix the issue. I'll bear this in mind and add the fields to my factories when I run into this issue in the future. Thanks! 😄

heychazza:
> @ash-jc-allen I think the "real" way to fix this would be to put `current_team_id` in the `UserFactory` so that it actually exists for new models in your application?

What do we set this to? I had the issue when setting strict mode, and wasn't sure how to fix.

CC: @taylorotwell

PR #1167, opened 2022-10-20
Skip tests when the feature is disabled

calvertyang:
Added checks to skip the tests if the update profile information, reset password or team feature is disabled.


taylorotwell:
Just remove the tests if you disable those features.

PR #1169, opened 2022-10-24
[2.x] Vite preloading

timacdonald:
Depends on: https://github.com/laravel/framework/pull/44096

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "phpunit/phpunit": "^9.3"
     },
     "conflict": {
-        "laravel/framework": "<9.19.0"
+        "laravel/framework": "<9.37.0"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -29,7 +29,7 @@
         "phpunit/phpunit": "^9.3"
     },
     "conflict": {
-        "laravel/framework": "<9.19.0"
+        "laravel/framework": "<9.37.0"
```
timacdonald, line 32: Taking a guess that this is the version we release this in. Will need to confirm on release day.

src/Console/InstallCommand.php
```diff
@@ -366,6 +366,7 @@ protected function installInertiaStack()
             });
 
         $this->installMiddlewareAfter('SubstituteBindings::class', '\App\Http\Middleware\HandleInertiaRequests::class');
+        $this->installMiddlewareAfter('\App\Http\Middleware\HandleInertiaRequests::class', '\Illuminate\Http\Middleware\AddLinkHeadersForPreloadedAssets::class');
```
driesvints, line 369: Can we use a combination of a class import and `::class` string addition here?

PR #1175, opened 2022-11-01
[2.x] Uses PHP Native Type Declarations 🐘 

nunomaduro:
Please check https://github.com/laravel/laravel/pull/6010 for details about this pull request.

### DIFF
--- a/src/Contracts/InvitesTeamMembers.php
+++ b/src/Contracts/InvitesTeamMembers.php
@@ -2,15 +2,10 @@
 
 namespace Laravel\Jetstream\Contracts;
 
+/**
+ * @method void invite(\Illuminate\Foundation\Auth\User $user, \Illuminate\Database\Eloquent\Model $team, string $email, string $role = null)
+ */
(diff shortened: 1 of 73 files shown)


src/Contracts/InvitesTeamMembers.php
```diff
@@ -2,15 +2,10 @@
...
 {
-    /**
-     * Invite a new team member to the given team.
-     *
-     * @param  mixed  $user
-     * @param  mixed  $team
-     * @param  string  $email
-     * @return void
-     */
-    public function invite($user, $team, string $email, string $role = null);
```
driesvints, line 15: I'm a bit unsure about removing the methods from these contracts. Having these methods enforces the implementing class to provide the exact method name and parameters that's needed to run the program. Removing them will make them hollow and not really useful anymore. I realise there's no way around adding stricter argument types in userland for these methods but I'd argue that the method signature (except the argument types) is more important than the types themselves.

driesvints (changes requested):

PR #1183, opened 2022-11-14
Support for the Team Invitation model in the TeamInvitationController

gabeta:
I noticed that there was the possibility to load the models we want for Jetstream. Which is handy if the models don't use the same database connection as our default connection. However the TeamInvitationController was left behind.
It always uses the model the `Laravel\Jetstream\TeamInvitation` instead of loaded from the `Jetstream::newTeamInvitationModel()` like the CurrentTeamController does for example.

Sorry for my English, I use Google Translate to help me a bit thanks in advance.


taylorotwell:

PR #1186, opened 2022-12-01
Use x focus alpine plugin

x7ryan:
This PR replaces the custom focus trapping logic within the Livewire stack's modal with Alpine JS's x-Focus plugin.

The down-side is this does introduce a new dependency. However at the same time it reduces overall code needed to be maintained directly, is cleaner, widely used, all while improving accessibility. 

This improves accessibility bu utilizing the `.inert` modifier when defining the focus trapping (see: https://alpinejs.dev/plugins/focus#inert), which will add an `aria-hidden="true"` property to all elements outside the modal while it is open.

I believe this is a net improvement and worth bringing in the dependency. However if you disagree Taylor, no hard feelings if you close this.

### DIFF
--- a/resources/views/components/modal.blade.php
+++ b/resources/views/components/modal.blade.php
@@ -13,35 +13,9 @@ $maxWidth = [
 @endphp
 
 <div
-    x-data="{
-        show: @entangle($attributes->wire('model')).defer,
-        focusables() {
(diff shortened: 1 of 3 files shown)


driesvints:
@x7ryan great pr!

PR #1189, opened 2022-12-05
Gives Users an Extra Option To Choose Companies Feature Instead Of Only Teams Feature

andrewdwallo:
**This feature gives users an extra option to use --companies instead of only --teams on installation.**

This is probably quite of an unusual feature pull request compared to what would usually happen but I wanted to at least try and see if ya'll were interested after I explained more about it.

**Summary**
Currently, when using/installing Jetstream into an application/project a person is allowed to choose what options and features they want to implement into their project that Jetstream currently supports. As of right now, one of the most used features (if I had to guess) would be the "Teams" option whenever someone installs Jetstream. As you already know... by default every user has a personal team, the owner of this team can invite "members" to their team, these members can have "memberships", and the owner can manage these members. Jetstream does really well with this setup and one day I was thinking, it would be nice If I can change the "entity group" to "Companies" instead of "Teams", and perhaps change "Members" to "Employees" where these "Employees" I suppose have an "Employeeship". Essentially all of Jetstream's functionality/logic isn't specific to a certain group of entities,... only the naming of them is... (which ya'll are already probably aware of), but If you actually thought about it there are many "entity groups" that share a common/similar function to that of a Team...

The current PR I am submitting essentially gives someone the option to choose from a "Company" Entity that they can implement into their project. I thought this might be a cool feature to implement because... maybe perhaps someone's project doesn't deal with teams but maybe some other group type. Only having the option of using "Teams" could break someone's decision to install Jetstream considering all of its routes, namespaces, paths, etc.. are using the same name.

I also believe that a Major trend right now with programming projects is building SaaS based applications. I think the option of using Companies as a group type to choose from will be a great feature that many people will choose while installing the GREAT Jetstream package. I am sure companies will like the option as well as programmers considering the entire world of business is based more on Companies with employees instead of Teams having members in my opinion.

In my PR the entire stub folder of Jetstream is more "modular" looking to where everything is separated by their "group entity", in which the entities as of right now, in my PR,... are Companies and Teams (default). The base files of each folder are the files that are used for either the regular Livewire or Inertia setup. Of course, as per usual, if a person did not opt for a group entity they will be installing the base Jetstream installation files that come from the stubs,.. With the setup being more modular looking,... more entities can be added in the future if Laravel decided they wanted to do so or if someone in the community wanted to.

I would understand if Taylor and Laravel did not want to implement something like this into their package considering it does add more files... But all of the logic is the same as it was with the teams option so there wouldn't be that much more maintenance then their was before. Essentially,... only the naming of the Files are different and the words inside of them are different based on just the words "team" vs "company". It isn't a dealbreaker to me if not accepted so I won't take any hard feelings...

That is all. Thank you Laravel and Taylor for the work you do for the community!


driesvints:
Thank you for your PR but we have no plans for this, sorry. Maybe it's cool if you released this as a plugin package?

PR #1190, opened 2022-12-06
Use RedirectsActions on the delete user form

riasvdv:
This uses the same implementation as the DeleteTeamForm so a redirect after deleting a resource is configurable.


taylorotwell:
But this would redirect to `fortify.home` by default which is not what we want?

riasvdv:
Ah right, I guess there's no way to use the trait then in this case, should I open a new PR which adds a `fortify.redirects.delete-user` config key instead?

taylorotwell:
Little unsure what right course of action would be, but yeah that sounds closer to what we need.

PR #1194, opened 2022-12-09
use if-else

### LINKED ISSUE #1192: TwoFactorAuthenticationDisabled event dispatched on every Profile page load
- Jetstream Version: 2.12.6
- Jetstream Stack: Inertia
- Uses Teams: no
- Laravel Version: 9.43.0
- PHP Version: 8.1.4
- Database Driver & Version: MySQL 8.0.27

### Description:
I'm attempting to trigger an email notification when a user disables Two Factor Authentication in their account. To do this I've bound a listener to the event `Laravel\Fortify\Events\TwoFactorAuthenticationDisabled`. If the currently logged in user does not have 2FA enabled, and they visit the Profile page, the event is fired (incorrectly IMO).

The event is also fired (correctly, again IMO) when the user disables 2FA, however because the page then reloads, you end up in the situation above and the event gets fired again.

I think this is caused due to the [`validateTwoFactorAuthenticationState()`](https://github.com/laravel/jetstream/blob/2.x/src/Http/Controllers/Inertia/Concerns/ConfirmsTwoFactorAuthentication.php#L18) method in the [`ConfirmsTwoFactorAuthentication`](https://github.com/laravel/jetstream/blob/2.x/src/Http/Controllers/Inertia/Concerns/ConfirmsTwoFactorAuthentication.php) trait (specifically the [last conditional](https://github.com/laravel/jetstream/blob/2.x/src/Http/Controllers/Inertia/Concerns/ConfirmsTwoFactorAuthentication.php#L37)). I'm not sure if it is best to update this method, or having seen https://github.com/laravel/jetstream/pull/1148, move the dispatching of the event within Fortify's [DisableTwoFactorAuthentication action](https://github.com/laravel/fortify/blob/1.x/src/Actions/DisableTwoFactorAuthentication.php) to within the [conditional](https://github.com/laravel/fortify/blob/1.x/src/Actions/DisableTwoFactorAuthentication.php#L18). Happy to open a pull request based on the repository it might best be resolved.

### Steps To Reproduce:
- Create a new Laravel + Jetstream (with Inertia) app.
- Add an event listener to `Laravel\Fortify\Events\TwoFactorAuthenticationDisabled` and log something in the `handle()` method.
- Register a fresh account and visit /user/profile
- Observe text in the log file.

I've created a demo repository at:
https://github.com/tomwelch/jetstream-bug-report

The commit to reproduce is here:
https://github.com/tomwelch/jetstream-bug-report/commit/a2c68f3cef92057f1270a74ea3d4bc73ea6d9013

Thanks,
Tom.

joelbutcher:
fixes #1192.

The above issue is caused by every conditional inside `ConfirmsTwoFactorAuthentication:: validateTwoFactorAuthenticationState(...)` being checked regardless of whether or not a previous condition was met.

There are two possible solutions to this problem:
1. Change the consecutive `if` conditions to `else if` conditions, meaning they're only checked if the previous check has failed.
2. Add a `return;` statement after each subsequent block if the condition is met

Since the method type hints a `void` return type, I've opted for the first option.

### DIFF
--- a/src/Http/Controllers/Inertia/Concerns/ConfirmsTwoFactorAuthentication.php
+++ b/src/Http/Controllers/Inertia/Concerns/ConfirmsTwoFactorAuthentication.php
@@ -29,12 +29,12 @@ trait ConfirmsTwoFactorAuthentication
         }
 
         // If was previously totally disabled this session but is now confirming, notate time...
-        else if ($this->hasJustBegunConfirmingTwoFactorAuthentication($request)) {
+        elseif ($this->hasJustBegunConfirmingTwoFactorAuthentication($request)) {
             $request->session()->put('two_factor_confirming_at', $currentTime);
         }
 
         // If the profile is reloaded and is not confirmed but was previously in confirming state, disable...
-        else if ($this->neverFinishedConfirmingTwoFactorAuthentication($request, $currentTime)) {
+        elseif ($this->neverFinishedConfirmingTwoFactorAuthentication($request, $currentTime)) {
             app(DisableTwoFactorAuthentication::class)(Auth::user());
 
             $request->session()->put('two_factor_empty_at', $currentTime);
--- a/src/Http/Controllers/Inertia/Concerns/ConfirmsTwoFactorAuthentication.php
+++ b/src/Http/Controllers/Inertia/Concerns/ConfirmsTwoFactorAuthentication.php
@@ -29,12 +29,12 @@ trait ConfirmsTwoFactorAuthentication
         }
 
         // If was previously totally disabled this session but is now confirming, notate time...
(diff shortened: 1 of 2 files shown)


taylorotwell:
Rather than changing the logic in this class, which honestly received a lot of thought to cover a lot of edge cases, I would rather change the logic around firing the event in DisableTwoFactorAuthentication action.

taylorotwell:
Updated: https://github.com/laravel/fortify/commit/04b4b9c20e421c415d0427904a72e08a21bdec27

joelbutcher:
@taylorotwell Yeh that makes sense. MFA is one of those complex issues that requires a lot of time and concentration to get it right. Thanks for taking a look though!

PR #1197, opened 2022-12-13
Fix english grammar in the Config Stub

bomshteyn:
* use a comma before 'or'
* A comma missing after the conjunctive/linking adverb 'Typically'
* Use a comma before 'but'


taylorotwell:
I will need to adjust the spacing to stay consistent 😅

PR #1203, opened 2022-12-23
add focus styling for settings dropdown trigger button

thema-d:
This change adds a focus style to **Settings Dropdown** trigger button in a similar manner to the **Teams Dropdown**.

Thank you.


taylorotwell:
The Blade version doesn't have this - so it either needs to be not be added or added to both locations. Also include before and after screenshots.

thema-d:
Sure I will do that.

PR #1206, opened 2022-12-30
add focus styling for settings dropdown trigger button

thema-d:
This change adds a focus style to **Settings Dropdown** trigger button in a similar manner to the **Teams Dropdown.**

The following screenshots show the focused state of the buttons before and after.
**P.S**: I have bumped up the background color of the buttons in the screenshot for the sake of contrast.

**Before**
Screenshot from 2022-12-30 12-49-40

**After**
Screenshot from 2022-12-30 12-50-35

### DIFF
--- a/stubs/inertia/resources/js/Layouts/AppLayout.vue
+++ b/stubs/inertia/resources/js/Layouts/AppLayout.vue
@@ -123,7 +123,7 @@ const logout = () => {
                                         </button>
 
                                         <span v-else class="inline-flex rounded-md">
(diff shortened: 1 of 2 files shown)


driesvints:
I think the current behavior is more desireable. The former is a switcher while the latter is a dropdown.

thema-d:
Hi @driesvints,  Thanks for replying.
What about when navigating using a keyboard? The Trigger button will be focused but no indication that it can be clicked or so.

driesvints:
ah sorry, didn't realize this was a focus style. Seems okay to me 👍

PR #1208, opened 2023-01-01
Fix the order of the v-if attribute.

milorojas:
Attribute "v-if" should go before "class" to keep the same style than are using the other elements.


driesvints:
Thanks!

PR #1212, opened 2023-01-08
[2.x] Replace width prop with widthClass in Dropdown component

thema-d:
This PR addresses an issue with Dropdown component (both `Dropdown.vue` and `dropdown.blade.php`) where the **width** prop only works if we pass in **48**. i.e `<Dropdown width="48" />`, `<x-jet-dropdown width="48" />`.

The current implementation accepts a number and converts it to a TailwindCSS class. And in this PR, I thought it would be better if we just accept TailwindCSS class in the first place, thus renaming the prop to **widthClass**. I have also updated the component's usage in a couple of places to reflect the change.

Thank you :)

### DIFF
--- a/resources/views/components/dropdown.blade.php
+++ b/resources/views/components/dropdown.blade.php
@@ -1,4 +1,4 @@
-@props(['align' => 'right', 'widthClass' => 'w-48', 'contentClasses' => 'py-1 bg-white', 'dropdownClasses' => ''])
+@props(['align' => 'right', 'contentClasses' => 'py-1 bg-white', 'dropdownClasses' => ''])
 
 @php
 switch ($align) {
@@ -31,7 +31,7 @@ switch ($align) {
             x-transition:leave="transition ease-in duration-75"
             x-transition:leave-start="transform opacity-100 scale-100"
             x-transition:leave-end="transform opacity-0 scale-95"
-            class="absolute z-50 mt-2 {{ $width }} rounded-md shadow-lg {{ $alignmentClasses }} {{ $dropdownClasses }}"
+            class="absolute z-50 mt-2 rounded-md shadow-lg {{ $alignmentClasses }} {{ $dropdownClasses }}"
             style="display: none;"
             @click="open = false">
         <div class="rounded-md ring-1 ring-black ring-opacity-5 {{ $contentClasses }}">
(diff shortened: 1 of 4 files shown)


resources/views/components/dropdown.blade.php
```diff
@@ -1,4 +1,4 @@
-@props(['align' => 'right', 'width' => '48', 'contentClasses' => 'py-1 bg-white', 'dropdownClasses' => ''])
+@props(['align' => 'right', 'widthClass' => 'w-48', 'contentClasses' => 'py-1 bg-white', 'dropdownClasses' => ''])
```
jessarcher, line 1: Ordinarily, I think this would be a breaking change because the Blade components in Jetstream aren't published by default. However, in this case, the prop wasn't actually doing anything so I think it's okay.

stubs/inertia/resources/js/Components/Dropdown.vue
```diff
@@ -6,9 +6,9 @@ const props = defineProps({
         type: String,
         default: 'right',
     },
-    width: {
+    widthClass: {
```
jessarcher, line 9: This isn't really width-specific any longer, so I wonder whether `contentClass` would be more appropriate. It's effectively just exposing the `class` attribute for the `content` slot.

jessarcher (commented):
In the places where `width` was set to "60" (which previously wouldn't have done anything as you've pointed out) there is still a `<div class="w-60">...</div>` wrapper in the `content` slot which is no longer necessary.

thema-d:
Hey @jessarcher, thanks for replying.
I too think you are right. We can just use `contentClasses` or `dropdownClasses`. I also just noticed that `Dropdown.vue` does not have `dropdownClasses` unlike its Blade counterpart. This prop should be there in case we want to apply some styles to the parent element of the content element.
So I am going to remove **"w-60"** and width prop. And add `dropdownClasses` prop to the Vue component.

taylorotwell:
Honestly I don't really follow this change. If we need more widths just PR more widths or add them to your own application?

thema-d:
Sure. Unrelated but HUGE thanks for everything you have done and built though some are hateful and ungrateful. I am a big fan. @taylorotwell

PR #1216, opened 2023-01-10
[2.x] Laravel v10 Support

driesvints:


joshhanley:
@driesvints have tagged a Livewire release v2.11.0 with Laravel 10 support, so that should unblock you here 🙂

Issue #1220, opened 2023-01-15
TwoFactorAuthenticationSettingsTest failed when eloquent strict mode is enabled.


Lakshan-Madushanka:
- Jetstream Version: 2.14.1
- Jetstream Stack: Livewire
- Uses Teams: no
- Laravel Version: 9.4.0
- PHP Version: 8.1.10
- Database Driver & Version: MySQL 8.0.30

### Description:
I decided to use eloquent strict mode in our new project but surprisingly jetstream TwoFactorAuthenticationSettingsTest failed with the following message.

![Screenshot 2023-01-15 184054](https://user-images.githubusercontent.com/47297673/212544104-c76e9c8a-758b-4c30-9ba7-1bbe009a6aec.png)

### Steps To Reproduce:

1). Setup laravel jetstream project.
2). [Enable strict mode](https://laravel.com/docs/9.x/eloquent#enabling-eloquent-strict-mode).
3. Run tests `php artisan test`

driesvints:
You've sent in a PR so let's see how it goes.

PR #1221, opened 2023-01-15
Fix breaking livewire TwoFactorAuthenticationSettingsTest when eloquent strict mode is enabled.

Lakshan-Madushanka:
This is a solution to issue #1220 please refer to it for more details.

This solves the minor bug of TwoFactorAuthenticationSettingsTest failing when eloquent strict mode is enabled.

### DIFF
--- a/src/Http/Livewire/TwoFactorAuthenticationForm.php
+++ b/src/Http/Livewire/TwoFactorAuthenticationForm.php
@@ -51,7 +51,7 @@ class TwoFactorAuthenticationForm extends Component
     public function mount()
     {
         if (Features::optionEnabled(Features::twoFactorAuthentication(), 'confirm') &&
-            is_null(Auth::user()->two_factor_confirmed_at)) {
+            isset(Auth::user()->two_factor_confirmed_at)) {
             app(DisableTwoFactorAuthentication::class)(Auth::user());
         }
     }


driesvints:
Doesn't this flips the behavior? You'r not checking if the timestamp is set in contract to before when you were checking if it wasn't.

Lakshan-Madushanka:
Yes, it can be. It should be the inverse of `isset()`. Sorry for that. I think you already have realized the reason anyway I decided to explain it.

TwoFactorAuthenticationSettingsTest case this `$this->actingAs($user = User::factory()->create())` method uses user factory which hasn't set value for two_factor_confrmed_at column (see the user factory definition). So that TwoFactorAuthenticationForm `Auth::user()` method does not contain two_factor_confrmed_at (Model cashing should be the reason) property that caused to throw the exception because of `Model::preventsAccessingMissingAttributes()` method of strict mode.

I think There are a couple of solutions to solve this,

1). Set default value for two_factor_confrmed_at column in `User::factory()->create()` method in each test case.
     `User::factory()->create([two_factor_confrmed_at => null])`
2). Refresh the $user returned from the factory create method.
    ` $user->refresh()`
3). Replace `isNull() ` with `! isset()` in TwoFactorAuthenticationForm `mount()` method. (This is what I tried to do).
Please let me know if I miss something. If not I can update the pull request.

PR #1224, opened 2023-01-17
[2.x] Strip dark classes from welcome component when dark mode isn't selected.

### LINKED ISSUE #1223: Welcome Component Is darkmode support without providing --dark on install
- Jetstream Version: ^2.15
- Jetstream Stack: Inertia & Livewire
- Uses Teams: yes & no
- Laravel Version: ^9.19
- PHP Version: 8.2
- Database Driver & Version: Mysql

### Description:

image

### Steps To Reproduce:

Fresh install laravel Jetstream without --dark. You can optionally enable Teams and choose either stack.

login and view the Welcome component

jessarcher:
Jetstream has a welcome page and a welcome component. The previous implementation was excluding both when stripping dark classes, but it should only exclude the welcome page so that the welcome component does not end up with dark classes.

Fixes #1223

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -262,10 +262,7 @@ class InstallCommand extends Command
             $this->removeDarkClasses((new Finder)
                 ->in(resource_path('views'))
                 ->name('*.blade.php')
-                ->notName('welcome.blade.php')
-                ->append((new Finder)
-                    ->in(resource_path('views/vendor/jetstream/components'))
-                    ->name('welcome.blade.php'))
+                ->filter(fn ($file) => $file->getPathname() !== resource_path('views/welcome.blade.php'))
             );
         }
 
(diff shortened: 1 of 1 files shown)


src/Console/InstallCommand.php
```diff
@@ -262,7 +262,7 @@ protected function installLivewireStack()
             $this->removeDarkClasses((new Finder)
                 ->in(resource_path('views'))
                 ->name('*.blade.php')
-                ->notName('welcome.blade.php')
+                ->filter(fn ($file) => $file->getPathname() !== resource_path('views/welcome.blade.php'))
```
jessarcher, line 265: I couldn't use `notPath` here in the same way as the Inertia stack because the `welcome.blade.php` file we want to exclude is in the root of the directory we're searching, and so anything I passed to `notPath` would also match the `welcome.blade.php` file in the `components` directory.

Issue #1226, opened 2023-01-18
Opt in dark mode support is actually not opt in


ziming:
- Jetstream Version: 2.15
- Jetstream Stack: Livewire
- Uses Teams: yes
- Laravel Version: 9.48
- PHP Version: 8.2.1
- Database Driver & Version: MySQL 8

### Description:

I originally commented in with screenshots proof. But thought i should create an issue instead as that is a closed PR and it is likely nobody sees it

https://github.com/laravel/jetstream/pull/1209#issuecomment-1386977200

But tldr, I installed jetstream months ago, composer update this morning and tonight my jetstream sites all look very weird with some components dark and some components white.

So it looks like it is actually not opt in as I never ran the jetstream:install --dark command since my installation of Jetstream is months ago, not today.

All I did is a composer update, nothing else.

RhysLees:
This was fixed here: https://github.com/laravel/jetstream/pull/1224

ziming:
I hope so. release is not tagged yet. will find out later if it did indeed fix the issue or not.

jessarcher:
> This was fixed here: #1224

That won't fix this particular issue. The problem is that existing users of the Livewire stack often won't have published views. After updating, they'll now be using views with dark classes.

> @ziming you should first publish your components as recommended on the [docs](https://jetstream.laravel.com/2.x/installation.html#livewire), then you can update.

This is correct, but I don't think it's something people will know about or should have to do.

At this stage, the only solution I can think of is to revert the dark mode changes.

We could then either:

1. Push them forward to a major release and mention it in the upgrade guide, or
2. Keep the original "legacy" views in the packages resources directory and put the dark mode versions in the stubs directory, and always publish them on new installs to avoid problems like this going forward.

jessarcher:
Hey @ziming, we've reverted the dark mode changes for now. The issue should be resolved when you update to Jetstream `2.16.0`.

PR #1225, opened 2023-01-18
Update nav menu dark text to be consistent with nav link

RhysLees:
Currently the dark dropdown text is gray-500 compared to nav-links which have text of gray-400.

This PR aims to make the nav menu items consistent.

PR #1209, opened 2023-01-02
Opt-in dark mode support 

fotrino:
This PR introduces opt-in dark mode support to Jetstream, borrowed from @jessarcher [PR on Breeze](https://github.com/laravel/breeze/pull/209)

```
php artisan jetstream:install {stack} --dark
```

Following the previous mentioned PR, the dark mode classes will be stripped unless dark mode is enabled. 
When installing livewire stack without dark mode, the components will be published in order to strip the dark classes from them.

### Screenshots
|Light|Dark|
|-----|-----|
|register-light|register-dark|
|tos-light|tos-dark|
|privacy-policy-light|privacy-policy-light dark|
|login-light|login-dark|
|reset-password-light|reset-password-dark|
|dashboard-light|dashboard-dark|
|profile-light|profile-dark|
|confirm-password-modal-light|confirm-password-modal-dark|
|log-out-sessions-light|log-out-sessions-dark|
|delete-account-light|delete-account-dark|
|create-team-light|create-team-dark|
|edit-team-light|edit-team-dark|
|delete-team-member-light|delete-team-member-dark|
|delete-team-light|delete-team-dark|
|manage-role-light|manage-role-dark|
|api-tokens-light|api-tokens-dark|
|token-confirmation-light|token-confirmation-dark|
|token-delete-light|token-delete-dark|
|edit-token-light|edit-token-dark|
|active-user-menu-light|active-user-menu-dark|
|team-menu-light|team-menu-dark|
|focus-menu-button-light|focus-menu-button-dark|
|responsive-menu-light|responsive-menu-dark|

### Notes
- I copied the styles of inputs & buttons from Breeze PR, which will change colors from blue to indigo, happy to change back all to blue if this presents a problem.
- I used Livewire stack as a basis and then copied to Vue, while doing this I found minor inconsistencies between the stacks  which are now improved. This includes the caret icon in the user's dropdown and some minor visual differences - keep this in mind when reviewing the diff.
- I've included the fix [focus styling in this](https://github.com/laravel/jetstream/commit/e03c3a9df7e33951ea428cff493e5ffb1b8c32ac) and included the dark mode options.

PR #1229, opened 2023-01-19
Dark Mode v2

fotrino:
This PR brings back Dark Mode changes introduced in [PR 1209](https://github.com/laravel/jetstream/pull/1209) without causing issues to existing applications with Jetstream installed as it was mentioned in [Issue 1226](https://github.com/laravel/jetstream/issues/1226).

## Changes
- As suggested by @jessarcher in [this comment](https://github.com/laravel/jetstream/issues/1226#issuecomment-1396196078), the views from `views/resources` remain unchanged to avoid issues with previous installs.

-  Original files from `views/resources` have been copied to `stubs/livewire/resources/views/` to keep track of changes.

- On new installations, the following directories `stubs/livewire/resources/views/components` `stubs/livewire/resources/views/mail` will be copied to `views/vendor/jetstream/`.

- On new installations without `--dark` directive, the command `php artisan vendor:publish --tag=jetstream-views --force` was removed. Views are already published as mentioned in point above.

- The changes introduced to strip dark classes [PR 1224](https://github.com/laravel/jetstream/pull/1224), dark nav menu in [PR 1225](https://github.com/laravel/jetstream/pull/1225) have been included.

## Step to verify Issue 1226 is Fixed

1. Install new laravel app
2. Install Jetstream 2.16 `composer require laravel/jetstream`
3. Run ```php artisan jetstream:install livewire --teams```
4. Check that there are not any published jetstream views in `views/vendor/jetstream`
5. Update Jetstream to `"laravel/jetstream": "dev-dark-mode-v2"` 
6. Include the PR repo.
```
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/fotrino/jetstream.git"
        }
      ],
```
7. Run `composer update`


jessarcher:
Thanks for tackling this one again @fotrino! I'll check it out properly once it's marked ready.

fotrino:
Hi @jessarcher when you have a chance, take a look at the PR. Thanks!

taylorotwell:
So does this essentially duplicate the Livewire views in Jetstream for the time being? The old legacy views and the new views that support dark mode?

fotrino:
Yes, exactly. This will prevent any issues in previous installations without published views & new ones will work as user defines upon installation (with or without dark mode) because all the components are being published from the livewire stubs.

taylorotwell:
@fotrino ... @jessarcher can correct me if I'm wrong, but we also batted around the idea of just tagging a Jetstream 3.0 - this would allow us to make the change without having to have the duplicated views for backwards compatibility.

fotrino:
@taylorotwell that sounds even better! 

What do you think about moving the files from `resources/views/` to `stubs/livewire/resources/views/`? 
I think it would make it easier to determine that those components belongs to the livewire stubs.

Also, the following command `php artisan vendor:publish --tag=jetstream-views` from **[JetstreamServiceProvider](https://github.com/laravel/jetstream/blob/0c4dd31a2d9d3d3b3e8ea9a892fcd5d050be343c/src/JetstreamServiceProvider.php#L164)** could be removed, since the files would already be published in new installations & previous ones upgrading from 2.x to 3.x should have already publish their assets in order to upgrade.

jessarcher:
I agree re: moving this to Jetstream 3.0. We'll just need to mention in the upgrade guide about publishing views prior to upgrading (if the user hasn't already).

> What do you think about moving the files from `resources/views/` to `stubs/livewire/resources/views/`? I think it would make it easier to determine that those components belongs to the livewire stubs.

I had already done some work on this in https://github.com/laravel/jetstream/pull/1118. If you take a look at [`master`](https://github.com/laravel/jetstream/tree/master) you'll notice the root-level `resources` directory is already gone.

> Also, the following command `php artisan vendor:publish --tag=jetstream-views` from **[JetstreamServiceProvider](https://github.com/laravel/jetstream/blob/0c4dd31a2d9d3d3b3e8ea9a892fcd5d050be343c/src/JetstreamServiceProvider.php#L164)** could be removed, since the files would already be published in new installations & previous ones upgrading from 2.x to 3.x should have already publish their assets in order to upgrade.

I'd already removed this in `master` as well :raised_hands:

I've just gone ahead and merged the latest `2.x` changes into `master` and fixed all of the conflicts, so it should _hopefully_ be straight-forward to apply these dark mode changes there :crossed_fingers:

The main conflict you may run into with the Inertia stack is that in 3.x the authenticated user is now shared under `auth.user` instead of just `user` (https://github.com/laravel/jetstream/pull/1073).

Thanks again for your work on this!

taylorotwell:
Hey @fotrino! Based on @jessarcher's feedback, could you send this PR to 10.x accordingly?

fotrino:
Hey @taylorotwell, just created a [PR](https://github.com/laravel/jetstream/pull/1234) on `master` based on the feedback from @jessarcher  😃

PR #1241, opened 2023-02-06
Improve permissions performance

DanielMcAssey:
This improves the performance of checking user roles in a team, by reducing how much is fetched from the DB and instead does the check on the DB instead.


taylorotwell:
This is actually going to be much worse performance if you call the method many times in one request.

DanielMcAssey:
Understandable, but what we have found in our testing is that `contains()` and the `users->where()` fetches all the teams the user belongs to as well as all the users in that team, And depending on how many teams that user is part of and how many users are in that team, this improves performance significantly.
I understand this needs to work accross the board and with loads of different use cases, but for us this takes 20% of the time, the other call did.

PR #1248, opened 2023-02-10
Import models from the App namespace in stubs

ozanakman:
As the actions are already being imported from the `App` namespace, this will apply the same import rule for the model files.

### DIFF
--- a/stubs/app/Providers/JetstreamServiceProvider.php
+++ b/stubs/app/Providers/JetstreamServiceProvider.php
@@ -3,6 +3,7 @@
 namespace App\Providers;
 
 use App\Actions\Jetstream\DeleteUser;
+use App\Models\User;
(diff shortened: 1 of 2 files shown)


taylorotwell:
Hmm, I'm not totally clear on why this is necessary.

ozanakman:
Oh I totally missed this one, the issue was related to using "cast" fields on Jetstream actions. Basically, they were not defined at all. This undocumented way solved my issue and I thought it could help others, or prevent issues. @taylorotwell

ozanakman:
> Hmm, I'm not totally clear on why this is necessary.

If I remember correctly, the issue was "cast" fields. They were not available in either Jetstream actions or tests. I could solve that issue with this undocumented `Jetstream::use...Model()` methods. I thought it could help others prevent dealing with the same issue in the future. @taylorotwell

PR #1255, opened 2023-02-15
[3.x] Upgrade guide

jessarcher:
This PR adds an upgrade guide from Jetstream 2.x to Jetstream 3.x.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,7 +2,8 @@
 
 ## Upgrading from Jetstream 2.x to Jetstream 3.x
 
-> **Note:** This upgrade guide only discusses upgrading to Jetstream 3.x. Upgrading your Laravel, Tailwind, Livewire or Inertia installations is outside the scope of this documentation and is not strictly required in order to use Jetstream 3.x. Please consult the upgrade guides for those libraries for information on their upgrade process.
+> **Note**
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -1,8 +1,86 @@
...
+## Upgrading from Jetstream 2.x to Jetstream 3.x
+
+> **Note**
+> This upgrade guide only discusses upgrading to Jetstream 3.x. Upgrading your Laravel, Tailwind, Livewire or Inertia installations is outside the scope of this documentation and is not strictly required in order to use Jetstream 3.x. Please consult the upgrade guides for those libraries for information on their upgrade process.
+
+- [Changes Common To Both Stacks](#jetstream-3x-changes-common-to-both-stacks)
+- [Livewire Stack Upgrade Guide](#jetstream-3x-livewire-stack)
+- [Inertia Stack Upgrade Guide](#jetstream-3x-inertia-stack)
+
+### Jetstream 3.x Changes Common To Both Stacks
```
jessarcher, line 12: Note that I've prefixed the headings in this section with "Jetstream 3.x" so that they generate predictable anchors that are different to the 2.x section below.

Otherwise, GitHub will suffix the duplicate anchors with an auto-incrementing number, which might be confused with the version number, and which would require us to manually bump them in our TOCs each time we add a new set of headings at the top of the page.

PR #1258, opened 2023-02-15
[3.x] Ensure CI considers backwards compatibility with Laravel 9

joelbutcher:
This PR updates the `Test Stubs` CI pipeline for Jetstream to ensure backwards compatibility with Laravel ^9.21, as per the `composer.json` minimum requirements.


driesvints:
This isn't needed. Just adds too much matrix builds. Just the latest one is enough. We technically don't support new Laravel v9 installs as of now.

PR #1260, opened 2023-02-15
[3.x] Run stub tests on PHPUnit and Pest independently

joelbutcher:
This PR iterates on my previous one and add's a new `tester` key to the "Test Stubs" job. This will ensure that tests pass for developers wanting to use PHPUnit instead of pest.


driesvints:
This all really isn't needed, thanks!

PR #1266, opened 2023-02-19
Add banner() and dangerBanner() macros to livewire redirector

iWader:
This PR enables the use of `redirect()->banner()` and `redirect()->dangerBanner()` macros from Livewire components


taylorotwell:
See: https://github.com/laravel/jetstream/blob/3.x/src/InteractsWithBanner.php

PR #1267, opened 2023-02-19
Fix eloquent strict mode compatibility

iWader:
If you enable eloquent strict mode the `TwoFactorAuthenticationSettingsTest` tests fail because the `UserFactory` does not include the `two_factor_confirmed_at` attribute


driesvints:
Seems this breaks the test suite.

PR #1271, opened 2023-02-22
Uses Native PHP Return Types & Fixes

andrewdwallo:
This PR implements the use of PHP Return Types and corrects wrong return types that were initially implemented. 

Here are the benefits of this PR:

** I want to mention first that this PR will not break any user experiences.
** After implementing this there was a dramatic performance increase in the compilation and response time (try it yourself).
** I think this is a good final adjustment to the backend for a while.

Using return types in PHP offers several benefits, including:

Type Safety: By specifying a return type, you can ensure that the function always returns a value of the expected type. This can help catch bugs and errors early and make it easier to reason about the behavior of your code.

Improved Readability: Return types make it clear what type of value a function returns, which can improve code readability and make it easier to understand the purpose of the function.

Better Documentation: Return types can be used to document the expected return value of a function, which can make it easier for other developers to understand how to use the function.

Easier Maintenance: By specifying return types, you can make it easier to update and modify code in the future, since you'll know what type of data the function is returning and what other parts of the code depend on that return value.

Performance Improvements: Return types can improve performance by reducing the amount of type juggling and type casting that PHP needs to do at runtime.

Error Prevention: Return types can help prevent errors by ensuring that the function returns a valid value of the expected type. This can help catch errors early and reduce the amount of debugging that needs to be done later on.

### DIFF
--- a/src/Actions/UpdateTeamMemberRole.php
+++ b/src/Actions/UpdateTeamMemberRole.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Jetstream\Actions;
 
+use Illuminate\Auth\Access\AuthorizationException;
 use Illuminate\Support\Facades\Gate;
 use Illuminate\Support\Facades\Validator;
 use Laravel\Jetstream\Events\TeamMemberUpdated;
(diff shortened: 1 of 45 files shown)


andrewdwallo:
Sorry, I guess this isn't wanted right now or did I do something incorrectly here?

driesvints:
@andrewdwallo we only apply type hints to userland code, not internal code.

andrewdwallo:
Oh okay, sorry... I really enjoy Jetstream and I was only trying to help. Is there a reason for this or is it just a decision the Laravel Team decided on? Everything worked smoothly with the PR...

andrewdwallo:
I also didn't make anything stricter in my PR, essentially this just corrected some return types and made everything more readable, and it also made the PHP compiler compile the code quicker which led to a much quicker response time in the browser, I'm not sure why that wouldn't be wanted...

PR #1274, opened 2023-02-25
[3.x] Use Pest

joelbutcher:
This PR updates Jetstream to use Pest as the default pest runner and migrates the **some** tests over to pest syntax.

Only some tests have been migrated as attempting to migrate the tests moved to the `Feature` test suite caused them to fail – any help / pointers on how to successfully do this would be appreciated!

### DIFF
--- a/tests/Unit/InviteTeamMemberTest.php
+++ b/tests/Unit/InviteTeamMemberTest.php
@@ -13,69 +13,59 @@ use Laravel\Jetstream\Tests\Fixtures\TeamPolicy;
 use Laravel\Jetstream\Tests\Fixtures\User;
 use Laravel\Jetstream\Tests\OrchestraTestCase;
 
-class InviteTeamMemberTest extends OrchestraTestCase
-{
-    public function setUp(): void
-    {
-        parent::setUp();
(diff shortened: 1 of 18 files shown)


tests/Unit/InviteTeamMemberTest.php
```diff
@@ -0,0 +1,64 @@
...
+
+    $team = $team->fresh();
+
+    $this->assertCount(0, $team->users);
+    $this->assertCount(1, $team->teamInvitations);
+    $this->assertEquals('adam@laravel.com', $team->teamInvitations->first()->email);
+    $this->assertEquals($team->id, $team->teamInvitations->first()->team->id);
+});
+
+test('user cant already be on team', function (): void  {
```
joelbutcher, line 46:
```suggestion
test('user cant already be on team', function (): void {
```

timacdonald:
If we are gonna use Pest, I feel the tests should be written in a Pest-centric style with expectations, etc., rather than a lift and shift.

Just a thought.

joelbutcher:
@timacdonald I've updated the pest tests to use pests expectations API

PR #1281, opened 2023-03-18
Get fresh user instance on two factor authentication tests

telkins:
In my installation, these tests fail because the user doesn't have the `two_factor_confirmed_at` attribute:
```
The attribute [two_factor_confirmed_at] either does not exist or was not retrieved for model [App\Models\User]. (View: /var/www/html/vendor/livewire/livewire/src/views/mount-component.blade.php)
```

I believe this is because the factory doesn't specify the attribute.

I considered updating the factory to include it (set to `null`), but then I thought it might be a little messy because there should be a check to see if the feature's enabled.

This solution, however, requires everyone to remember to get a fresh user instance, when needed, without it being clear as to why.

So...let me know if I should try a different approach...


taylorotwell:
I assume this is because you're using some of Eloquent's strict features?

Issue #1289, opened 2023-04-03
Page reload/refresh causes short-term view of other form input on Two-Factor Challenge Page


andrewdwallo:
- Jetstream Version: 3.1.0
- Jetstream Stack: Livewire
- Uses Teams: Yes
- Laravel Version: 10.5.1 
- PHP Version: 8.2.4
- Database Driver & Version: MySQL 8.0.30

### Description:
When a user refreshes or reloads the Two-factor Challenge page, there is a brief moment where the `Recovery Code` form input appears before disappearing. This is an issue because the default form input is supposed to be just Code. Obviously, one would have to click on "Use a recovery code" for the `Recovery Code` form input to appear by itself. I have included a video below that is slowed down so that you can see the issue more clearly. Even with a page load time of 0.2ms, it is still quite noticeable. If a user had a slow server or page load time, this issue could look very much worse.

### Steps To Reproduce:
This is on a fresh installation:

https://user-images.githubusercontent.com/104294090/229633978-2646135e-32f8-4c59-ac98-9781570cc068.mp4

andrewdwallo:
It would probably be better to use Alpine's "x-cloak" directive here.

andrewdwallo:
Definitely an alpine issue. Page flickering caused by `x-show` directive. Should be able to use `x-cloak` to fix it but I had no luck with that.

andrewdwallo:
Fixed: https://github.com/laravel/jetstream/pull/1291#issue-1655185396

andrewdwallo:
@driesvints Sorry to @ you but I am trying to bring attention to this here: https://github.com/laravel/jetstream/pull/1291#issuecomment-1498476444

The pull request was already merged but it needs to be reverted/changed.

PR #1291, opened 2023-04-05
Fix issue with page reload/refresh on two-factor-challenge page

andrewdwallo:
The following pull request fixes the issue as explained here: https://github.com/laravel/jetstream/issues/1289#issue-1652817995

Reloading/refreshing the page no longer causes the recovery code form input to appear for a short amount of time. The "blip" as Alpine calls it is no longer occuring, which was causing the issue in the first place. 

In Alpine's documentation:

"_Sometimes, when you're using AlpineJS for a part of your template, there is a "blip" where you might see your uninitialized template after the page loads, but before Alpine loads._"

"_`x-cloak` addresses this scenario by hiding the element it's attached to until Alpine is fully loaded on the page._"

### DIFF
--- a/stubs/livewire/resources/views/auth/two-factor-challenge.blade.php
+++ b/stubs/livewire/resources/views/auth/two-factor-challenge.blade.php
@@ -1,7 +1,3 @@
-<style>
-    [x-cloak] { display: none; }
-</style>
-
 <x-guest-layout>
     <x-authentication-card>
         <x-slot name="logo">
--- a/stubs/livewire/resources/views/auth/two-factor-challenge.blade.php
+++ b/stubs/livewire/resources/views/auth/two-factor-challenge.blade.php
@@ -1,3 +1,7 @@
+<style>
+    [x-cloak] { display: none; }
+</style>
+
 <x-guest-layout>
     <x-authentication-card>
         <x-slot name="logo">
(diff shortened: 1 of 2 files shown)


taylorotwell:
Can the x-cloak styling go into a stylesheet?

fotrino:
Yes, I've it in my `app.css` 
```css
@tailwind base;
@tailwind components;
@tailwind utilities;

[x-cloak] { display: none; }
```

andrewdwallo:
Apologies, I believe that the `x-cloak` directive should only be applied where `x-show="recovery"` is, not where 
`x-show="! recovery"` is. This is because refreshing the page while you are actually showing the "recovery code" form input, by default, will just bring you back to the form input "code". I will test it out again, but I have just learned that I think that is the case after using what we have now on a very slow server. What we have now works, but I don't think it is the best way to handle it. If I do find that this is the case, do you want me to create another pull request to fix it or do you want to? All that would be needed to be changed is to remove `x-cloak` wherever `x-show="! recovery"` is.

andrewdwallo:
Yep, I was right. `x-cloak` should only be where `x-show="recovery"` is. We need to remove it wherever `x-show="! recovery"` is. Here are the differences. I purposely made the server slow here. 

**What we have now, which worked to remove the brief view of the "recovery code" form input:**

[](https://user-images.githubusercontent.com/104294090/230276853-24568e8e-a050-435d-bcd2-9a5286e42000.mp4)

&ensp;

**What we should use. Which is remove `x-cloak` where `x-show="! recovery"` and keep `x-cloak` where `x-show="recovery"`:**

[](https://user-images.githubusercontent.com/104294090/230277074-ac67acb4-cad2-47d7-9241-1f0d41296edc.mp4)

taylorotwell:
Pushed your suggested fix.

andrewdwallo:
> Pushed your suggested fix.

Great! Thanks

PR #1292, opened 2023-04-05
[3.x] Remove `v-show` directive from "user profile information form" template

ndresx:
Hi, I've noticed that there is a `v-show` directive used on a div in the Livewire-powered `update-profile-information-form` Blade template, which is also wrapped into a `@if` statement. I'd assume it's the one or the other, in this case, without it.


driesvints (approved):

driesvints:
Thanks!

driesvints:
Thanks!

PR #1295, opened 2023-04-10
Team Member Invitations: Use AlpineJS for Role selection

andrewdwallo:
This pull request creates an easier process for the user to switch between the current role they want to choose for the given team member when sending an invitation and editing a current team member's role. There is no reason to use Livewire to change the team member's role. This only creates an extra Ajax request every time a switch between roles is made. I changed this to use the `@entangle` directive and chained a `.defer` property so that it isn't necessary to update Livewire on every change, and to only bundle the change with the next Livewire request that goes out.

I also fixed a CSS issue that was causing a notable change in the size of the form/selection box whenever a user switches between the current role for the team member by removing an unnecessary `focus:border-none` TailwindCSS utility class. Everything here should make the process quicker, easier, and more effective.

### DIFF
--- a/stubs/livewire/resources/views/teams/team-member-manager.blade.php
+++ b/stubs/livewire/resources/views/teams/team-member-manager.blade.php
@@ -33,22 +33,20 @@
                             <x-label for="role" value="{{ __('Role') }}" />
                             <x-input-error for="role" class="mt-2" />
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Make the change in your own application if you wish.

andrewdwallo:
> Make the change in your own application if you wish.

Okay.

PR #1299, opened 2023-04-18
Fix validation of the two factor challenge with an empty recovery code input

sebastiansulinski:
### Description

This PR together with the one I've submitted to [laravel/fortify](https://github.com/laravel/fortify/pull/459) is providing the fix for the following scenario:

---

While authenticating with 2FA using the code option with empty input, on submission we get the validation message:

Screenshot 2023-04-15 at 12 10 44

however when doing the same using recovery code option - we do not see the error:

Screenshot 2023-04-15 at 12 12 06

### Steps To Reproduce

Authenticate using username and password, followed by 2FA form - first try the `code` and then `recovery code` - both times submit form with empty input. 

Only `code` version will display validation message.


taylorotwell:
Your component name has a spelling error?

PR #1303, opened 2023-04-25
set email as required in input for UpdateUserProfileInformation

ordago:
The controller validation has it as required. This change reflects that in the frontend.

https://github.com/laravel/jetstream/blob/e9a3879058bce5a4d25b6b2e7b70c961d7909c75/stubs/app/Actions/Fortify/UpdateUserProfileInformation.php#L20-L22

Avoids waiting for the server to show the error to the user and helps with accessibility by informing a priori of the requirement.


andrewdwallo:
Same would need to be done with the Name field then, and also the Livewire input fields for both of them.

driesvints:
Heya, please update the Livewire side of things as well, thanks!

ordago:
@andrewdwallo @driesvints made the change for the name field and the Livewire component

PR #1308, opened 2023-05-05
Added 'prefix' to Jetstream config for localization.

RalphMRivera:
**This addition is intended to clarify where localization prefixes go in the Jetstream config as a parallel to how prefixes work in Fortify.**

This change benefits those developing localized sites that use Fortify and Jetstream and want to normalize the behavior between the two. This change also adds clarity and brings attention to the fact that Jetstream can be prefixed similarly to Fortify.

Note: I did not add the 'domain' section in the Jetstream config that appears in the Fortify config because it doesn't seem to apply; however, if it does, I can also add it.

Cheers,

Ralph


taylorotwell:

PR #1309, opened 2023-05-08
Ensure email is included in state

antja0:
If email is hidden for serialization (in the User model `hidden` array) the view will crash, as `toArray()` won't include the email.

Hiding email field in particular is useful to avoid accidentally exposing other users emails to the frontend / client.


taylorotwell:
Wrong branch. Send to 3.x.

PR #1311, opened 2023-05-08
Install Pint & Format

fotrino:
This PR install `laravel/pint` as a dev dependency & formats the code using the Laravel's default.
This will ensure future contributions are always properly formatted & we maintain consistency in the coding styles.

It might be a good idea to create a GitHub Action to run when a PR is created.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
     },
     "require-dev": {
         "inertiajs/inertia-laravel": "^0.6.5",
+        "laravel/pint": "^1.10",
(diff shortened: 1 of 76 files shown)


taylorotwell:
I think this is largely something we would handle internally in a uniform way across all repositories.

driesvints:
Hi @fotrino. I sent in an alternative where we only lint the stub files: https://github.com/laravel/jetstream/pull/1313. Thanks!

PR #1318, opened 2023-05-27
added space in 'DropDownLink.vue' component

filippetrovicAT:
Added missing space to if statement inside of DropDownLink.vue component.

### DIFF
--- a/stubs/inertia/resources/js/Components/DropdownLink.vue
+++ b/stubs/inertia/resources/js/Components/DropdownLink.vue
@@ -13,7 +13,7 @@ defineProps({
             <slot />
         </button>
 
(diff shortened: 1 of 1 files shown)


stubs/inertia/resources/js/Components/Dropdown.vue
```diff
@@ -65,9 +65,8 @@ const alignmentClasses = computed(() => {
         >
             <div
                 v-show="open"
-                class="absolute z-50 mt-2 rounded-md shadow-lg"
+                class="absolute hidden z-50 mt-2 rounded-md shadow-lg"
                 :class="[widthClass, alignmentClasses]"
-                style="display: none;"
```
jessarcher, line 70: The `hidden` class will prevent `v-show` from working so you'll never be able to see the dropdown.

I'm not sure why we needed the inline style though - I'd assume it could just be removed?

PR #1329, opened 2023-06-19
Use appName directly from .env in <title>

domnantas:
Currently, appName in app.tsx is pulled via DOM from app.blade.php <title> tag. It works, but using it from .env seems like a better approach. Additionally, it serves as an example of how .env variables should be accessed in React/Vue.

The ssr template had 'Laravel' hardcoded as the app name, which is not great for SEO. Replaced that part with import.meta.env.VITE_APP_NAME as well.

This depends on https://github.com/laravel/laravel/pull/6204 where I added VITE_APP_NAME as an alias to APP_NAME

I don't think any changes in this PR could be considered breaking.

Similar issue fixed in Breeze: https://github.com/laravel/breeze/pull/292


driesvints (approved):
I like this as it allows to set a different app name for when you're working in different environments.

timacdonald (approved):

jessarcher (approved):
Nice improvement!

PR #1338, opened 2023-07-09
Turn on text light when dark mode.

irsyadadl:


taylorotwell:
No explanation and no before / after screenshots.

PR #1345, opened 2023-07-19
Update AuthenticationCard Component

D4ve-R:
Change background in AuthenticationCard to fit Welcome page


taylorotwell:
I don't think it's necessarily intended to have this.

PR #1353, opened 2023-08-08
[3.x] Pest detection

jessarcher:
This PR updates the standalone Jetstream installer to default to Pest when it is already installed, either via the Laravel installer or installed manually.

When running `php artisan jetstream:install` with no arguments, the testing framework prompt will still be displayed, but defaulting to Pest when installed:

image

When passing arguments to the installer (e.g. `php artisan jetstream:install inertia`) the prompt does not appear. Currently, it will install PHPUnit tests unless the `--pest` flag was passed. With this change, it will automatically install Pest tests when appropriate.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -649,7 +649,7 @@ EOF;
      */
     protected function getTestStubsPath()
     {
-        return $this->option('pest')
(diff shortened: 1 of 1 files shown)


nunomaduro (approved):
@jessarcher I've improved the `isUsingPest` function.

PR #1361, opened 2023-08-20
Ensure current team is set when rendering menu

itbm:
This PR ensures the user has a current team set when rendering the menu. It is necessary to prevent errors in scenarios where the user doesn't have any teams.


taylorotwell:
You can add this to your own application. Jetstream is designed with the intention that the user always has some team.

PR #1360, opened 2023-08-17
[4.x] Migrates to Livewire `v3.x`

nunomaduro:
This pull request migrates Jetstream to Livewire `v3.x`. Wish to help with this pull request? Simply follow this instructions, and report any bugs you find:

```
laravel new nuno_pr --pest --no-interaction
cd nuno_pr
composer config minimum-stability dev
composer require laravel/jetstream:dev-feat/livewire-3
php artisan jetstream:install livewire --dark --teams --api --verification --pest
sed -i.bak '/DB_/d' ./.env
echo "DB_CONNECTION=sqlite" >> .env
php artisan migrate --force
php artisan serve # Click on http://127.0.0.1:8000
```

PR #1365, opened 2023-08-22
Update action-message.blade.php

AbdullahObaid:
On action-message.blade.php component file, @this.on should be replaced by x-on since @this.on() deprecated in livewire 3 https://livewire.laravel.com/docs/upgrading#thison-deprecated

### DIFF
--- a/stubs/livewire/resources/views/components/action-message.blade.php
+++ b/stubs/livewire/resources/views/components/action-message.blade.php
@@ -1,7 +1,7 @@
 @props(['on'])
 
 <div x-data="{ shown: false, timeout: null }"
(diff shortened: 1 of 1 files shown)


AbdullahObaid:
Replacing @this.on to x-on:____.window will listen to events from any Livewire component on the page, so this solution will be triggered from any component in the page with same event name. On the other hand using  x-on:____ without .window will not work because It only listen for events on child components.

AbdullahObaid:
closed due to @this.on added to livewire 3 https://github.com/laravel/jetstream/pull/1360#issuecomment-1688939143

PR #1376, opened 2023-09-16
Prevent calling Inertia before it's installed

jrd-lewis:
When you pull in Jetstream into a fresh Laravel project manually and try accessing it in a browser before running the ```php artisan jetstream:install``` command, it will have the error that "Class 'Inertia\Inertia' not found" from the ShareInertiaData middleware. To prevent that, I added a check to ensure the class is present before calling the ```bootInertia``` method in the ServiceProvider.

image

### DIFF
--- a/src/JetstreamServiceProvider.php
+++ b/src/JetstreamServiceProvider.php
@@ -88,7 +88,7 @@ class JetstreamServiceProvider extends ServiceProvider
             ]);
         });
 
-        if (config('jetstream.stack') === 'inertia') {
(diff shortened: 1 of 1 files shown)


driesvints:
I don't think this is necessary? Why would you try to run the app before running the install command?

jrd-lewis:
It would bring consistency with line 40 of this service provider where it ensures Livewire is installed before registering the Livewire components.

One use case is someone is doing TDD for their project and pulls in Jetstream to handle authentication redirections for guest users but doesn't run the install command immediately. In that case, it would bring up the error about the Inertia class not being found as it does in the browser.

PR #1382, opened 2023-09-24
Replace @livewireScripts with @livewireScriptConfig

stephaneisenbach:
According to Livewire Upgrade Guide from version 2 to 3.

See: https://livewire.laravel.com/docs/upgrading#including-via-js-bundle


taylorotwell:
Not the same thing I don't think?

PR #1383, opened 2023-09-26
Allow banner.blade.php to be a link

ziming:
Example use case, we want to encourage email verification but didn't want to strictly enforce it

```blade
@if(auth()->user()->hasVerifiedEmail() === false)
    <x-banner style="info" message="We encourage you to verify your email for bonus points, click here!" href="{{ route('verification.notice') }}" target="_blank" />
@endif
```

There are many other potential use cases as well for allowing the banner to be a link

You can still click on the x button to close the banner like the past if you want


taylorotwell:
Hey there - this belongs to your application so you can change it however you want in your own application.

PR #1399, opened 2023-11-06
[4.x] Replace `jenssegers/agent` and use latest `mobiledetect/mobiledetectlib`

crynobone:
* Port `Jenssegers\Agent\Agent` to `Laravel\Jetstream\Agent`
  - Without `isRobot()` detection

### DIFF
--- a/src/Agent.php
+++ b/src/Agent.php
@@ -44,13 +44,11 @@ class Agent
         'IE' => 'MSIE|IEMobile|MSIEMobile|Trident/[.0-9]+',
         'Netscape' => 'Netscape',
         'Mozilla' => 'Mozilla',
-        'WeChat'  => 'MicroMessenger',
+        'WeChat' => 'MicroMessenger',
     ];
 
     /**
      * Construct a new agent.
-     *
-     * @param  string  $userAgent
      */
     public function __construct(
         protected string $userAgent
@@ -99,7 +97,6 @@ class Agent
     /**
      * Match a detection rule and return the matched key.
      *
-     * @param  array  $rules
      * @return string|null
      */
     protected function findDetectionRulesAgainstUserAgent(array $rules)
@@ -137,7 +134,7 @@ class Agent
                 } elseif (is_array($merged[$key])) {
                     $merged[$key][] = $value;
                 } else {
-                    $merged[$key] .= '|' . $value;
+                    $merged[$key] .= '|'.$value;
                 }
             }
         }
--- a/src/Agent.php
+++ b/src/Agent.php
@@ -1,0 +1,146 @@
+<?php
+
+namespace Laravel\Jetstream;
+
+use Detection\MobileDetect;
+
+class Agent
+{
+    /**
+     * List of operating systems.
+     *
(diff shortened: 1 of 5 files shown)


src/Agent.php
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * @return bool
+     */
+    public function isDesktop()
+    {
+        // Check specifically for cloudfront headers if the useragent === 'Amazon CloudFront'
+        if (
+            $this->getUserAgent() === static::$cloudFrontUA
+            && $this->getHttpHeader('HTTP_CLOUDFRONT_IS_DESKTOP_VIEWER') === 'true'
+        ) {
```
crynobone, line 85: This requires https://github.com/serbanghita/Mobile-Detect/blob/b73a5bdbcec74a6d33389cc8fe3a5af5d0965d01/src/MobileDetect.php#L1036-L1038 to actually work.

driesvints (approved):
Amazing work @crynobone! 👏

Sarsmooch:
I don't know if I could be doing something wrong, but after the update, `$session->agent->platform()` and `$session->agent->browser()` return just 1. I created a new empty laravel project to make sure of this, and the same thing happens.

Jared87:
> I don't know if I could be doing something wrong, but after the update, `$session->agent->platform()` and `$session->agent->browser()` return just 1. I created a new empty laravel project to make sure of this, and the same thing happens.
> 
> What seems even stranger to me is that during testing, when doing `{{ $session->agent->platform() }}`, this prints the platform correctly, while `{{ $session->agent->platform() ? $session->agent->platform() : __('Unknown') }}` prints '1' (true)

I have the same issue, as i see the problem seems to occur somewhere in the caching, but i cannot say it in detail. 
The Agent::retrieveUsingCacheOrResolve() Method has the correct result, but the return value seems to be true.

Sarsmooch:
> I have the same issue, as i see the problem seems to occur somewhere in the caching, but i cannot say it in detail. The Agent::retrieveUsingCacheOrResolve() Method has the correct result, but the return value seems to be true.

I also looked into this, including the Agent::retrieveUsingCacheOrResolve()` method, I cleared the cache and everything I could, I hosted it on a server in order to check if it could be a problem on my machine, so far I haven't identified the real problem. Maybe later I look for more.

For now, what worked for me was simply modifying the echo to `{{ $session->agent->platform() ?: 'Unknown' }}`.

crynobone:
@Sarsmooch can you submit a new bug report with Browser + Platform + User Agent string so we can debug this further.

Rastian95:
I don't understand why remove functionalities instead of add new ones!?
 There is no more, isPhone(), isRobot(), robot(), deviceType(), device(), languages()
So now we need to extend the class and add these functions from jenssegers/agent? or what do you recommend to do?

crynobone:
> So now we need to extend the class and add these functions from jenssegers/agent? or what do you recommend to do?

Feel free to add class/feature based on your requirements. `Agent` usage is limited to what's being used within Jetstream.

PR #1412, opened 2023-11-29
[4.x] Fix Browser Sessions not showing platform and browser

olumby:
Browser Sessions are currently showing '1-1'  when they should be displaying 'Platform - Browser', the image below shows reality/expected behaviour.

browser sessions

#1399 used the cache implemented in `mobiledetect/mobiledetectlib` which [only supports `bool` values](https://github.com/serbanghita/Mobile-Detect/blob/90416e034adbf939c436c8bbb16877d776fc239b/src/Cache/CacheItem.php#L15-L18). As multiple calls to `agent->platform()` and `agent->browser()` are made in the component the cache is engaged and produces boolean values instead of the expected strings.

This pull request replaces the `mobiledetect` cache implementation with a simple key value store on the `Agent` class.

Current tests pass as the first call to `agent->platform()` and `agent->browser()` do not hit the cache. Hopefully I will have time tomorrow to update the tests too.

I am not sure if this is the ideal solution or if the `mobiledetect` cache should be reimplemented to support mixed values. I briefly considered using the Laravel cache, but considering this data is only stored for the request lifecycle it seemed a little over the top.

### DIFF
--- a/src/Agent.php
+++ b/src/Agent.php
@@ -3,8 +3,6 @@
 namespace Laravel\Jetstream;
 
 use Closure;
-use Detection\Cache\CacheException;
-use Detection\Exception\MobileDetectException;
 use Detection\MobileDetect;
 
 /**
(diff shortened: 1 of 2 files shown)


crynobone:
@olumby We can utilise `Cache::driver('array')->rememberForever($cacheKey)` which will only exist during the request lifecycle.

crynobone (approved):

olumby:
nice, thanks for sorting out the tests 😄

PR #1414, opened 2023-12-03
Update inertiajs/inertia-laravel

HPWebdeveloper:
Updating inertiajs/inertia-laravel version

**Before:**

"inertiajs/inertia-laravel": "^0.6.5",

**After:**

"inertiajs/inertia-laravel": "^0.6.11",


driesvints:
Thanks but this already pulls in that version.

PR #1419, opened 2023-12-19
changed the `users` table reference to model reference in validations…

olaitan6201:
… - `stubs\app\Actions`


driesvints:
Don't think this is necessary?

olaitan6201:
Not really necessary. But to take an instance that I had to work on converting an existing application to Laravel where the table name for `users` changed.

PR #1422, opened 2023-12-22
Fixed email autocomplete field for inertia and livewire in UpdateProfileInformationForm component.

christophheich:
As the title states the autocomplete field has been changed to email instead of username for the email form of the UpdateProfileInformationForm.


taylorotwell:
Change in your own app if you wish.

christophheich:
> Change in your own app if you wish.

I did already, but why keeping an incorrect autocomplete field (username) for an email field?

PR #1433, opened 2024-01-17
[4.x] Check for confirm in fortify

michail-polatoglou:
There is a bug in the current logic for the `two_factor_enabled` in the SharedInertiaData middleware. It returns true even if the confirmed option in `fortify.php` config is set to true leading in an error if you refresh the page in the confirmation page. 

This is not a problem in the default configuration of the 2FA in the jetstream profile page but if you want to move this form to a different page it causes a problem with the logic since the front end thinks the `two_factor_enabled` is true even if it's not confirmed and only shows you the recovery codes. 

The function could be shortened to:
```
if (
    Features::enabled(Features::twoFactorAuthentication()) &&
    !is_null($user->two_factor_secret) &&
    (
        !Features::enabled(Features::twoFactorAuthentication(['confirm'])) ||
        !is_null($user->two_factor_confirmed_at)
    )
) {
    return true;
} else {
    return false;
}
```
but i expanded it for readability

### DIFF
--- a/src/Http/Middleware/ShareInertiaData.php
+++ b/src/Http/Middleware/ShareInertiaData.php
@@ -83,6 +83,7 @@ class ShareInertiaData
                 }
             }
         }
+        
         return false;
     }
 }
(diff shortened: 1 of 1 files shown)


taylorotwell:
No plans to change this atm.

PR #1464, opened 2024-03-30
Support confirmPassword false for 2FA

### LINKED ISSUE #1463: confirmPassword-config currently broken
### Jetstream Version

v5.0.1

### Jetstream Stack

Livewire

### Laravel Version

v11.1.1

### PHP Version

8.3.0

### Database Driver & Version

_No response_

### What I want to do

In my settings I want to turn off password-confirmations, as my users do not have passwords. 

### Desc

Fortify has a feature/config-flag to enable or disable password confirmations for certain actions:

https://github.com/laravel/fortify/blob/1.x/stubs/fortify.php#L154

It seems that this should, at least, toggle wether or not a password is required for activating/modifying 2FA: 

https://github.com/laravel/fortify/blob/0c1721f0e50dd63fa3a4b4ad7ade553b821ec015/routes/routes.php#L139

On the PHP-side in Jetstream, that is also supported:

https://github.com/laravel/jetstream/blob/a57f6e9bc24a8befb74c5d668cd247ffe39bc40a/src/Http/Livewire/TwoFactorAuthenticationForm.php#L65

But it seems that the feature is hardcoded into the view and thus, setting it to `false`, has no effect.

https://github.com/laravel/jetstream/blob/5.x/stubs/livewire/resources/views/profile/two-factor-authentication-form.blade.php#L81

### Steps To Reproduce

1. Start a fresh project
2. Install Jetstream with livewire
3. Switch the setting https://github.com/Suven/jetstream-example/commit/2b5ac739ebade8352d33c2ab7826a309a50c10c9 
4. It does not take effect

Suven:
Fortify, as well as the Jetstreams 2FA component class has code that supports skipping the password confirmation, if set in the according configuration via a feature-flag.

The current implementation of the livewire view however, does not respect that config.

This admittedly naive approach fixes this. 

The feature is especially useful for applications that don't make use of passwords (for example because they are using business SSO but still want to enforce 2FA)

This adresses issue #1463

### DIFF
--- a/src/Http/Livewire/TwoFactorAuthenticationForm.php
+++ b/src/Http/Livewire/TwoFactorAuthenticationForm.php
@@ -37,6 +37,13 @@ class TwoFactorAuthenticationForm extends Component
     public $showingRecoveryCodes = false;
 
     /**
+     * Wether or not to confirm actions with the current password.
+     *
+     * @var bool
+     */
+    public $confirmPassword = true;
+
+    /**
      * The OTP code for confirming two factor authentication.
      *
      * @var string|null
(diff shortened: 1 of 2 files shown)


taylorotwell:
I don't think we ever really meant this to be disabled. It was only added for backwards compatibility at the time. I suggest just removing stuff from your stubs if you want to disable it.

PR #1467, opened 2024-04-03
Fix incorrect annotations in Livewire components using "@return void"

coreymcmahon:
Hi!

This is a _tiny_ update to correct some of the type annotations which are currently `@return void` when a return type is possible.

This obviously shouldn't impact any runtime behaviour, but it's causing PHPStan errors in one of my projects so it would be appreciated if we could correct the type annotations 🙂 

I tried to follow the return type conventions used in other places, but please feel free to suggest superior alternatives if you believe they exist.

Thanks for all you do 🙏

### DIFF
--- a/src/Http/Livewire/TeamMemberManager.php
+++ b/src/Http/Livewire/TeamMemberManager.php
@@ -183,7 +183,7 @@ class TeamMemberManager extends Component
      * Remove the currently authenticated user from the team.
      *
      * @param  \Laravel\Jetstream\Contracts\RemovesTeamMembers  $remover
-     * @return \Illuminate\Routing\Redirector|\Illuminate\Http\RedirectResponse
+     * @return \Illuminate\Http\RedirectResponse
      */
     public function leaveTeam(RemovesTeamMembers $remover)
     {
--- a/src/Http/Livewire/TeamMemberManager.php
+++ b/src/Http/Livewire/TeamMemberManager.php
@@ -183,7 +183,7 @@ class TeamMemberManager extends Component
      * Remove the currently authenticated user from the team.
      *
      * @param  \Laravel\Jetstream\Contracts\RemovesTeamMembers  $remover
-     * @return void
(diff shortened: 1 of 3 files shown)


src/Http/Livewire/UpdateProfileInformationForm.php
```diff
@@ -50,7 +50,7 @@ public function mount()
      * Update the user's profile information.
      *
      * @param  \Laravel\Fortify\Contracts\UpdatesUserProfileInformation  $updater
-     * @return void
+     * @return \Illuminate\Routing\Redirector|\Illuminate\Http\RedirectResponse|void
```
driesvints:
```suggestion
     * @return \Illuminate\Routing\Redirector|\Illuminate\Http\RedirectResponse|null
```

driesvints (approved):

src/Http/Livewire/TeamMemberManager.php
```diff
@@ -183,7 +183,7 @@ public function stopManagingRole()
      * Remove the currently authenticated user from the team.
      *
      * @param  \Laravel\Jetstream\Contracts\RemovesTeamMembers  $remover
-     * @return void
+     * @return \Illuminate\Routing\Redirector|\Illuminate\Http\RedirectResponse
```
driesvints:
```suggestion
     * @return \Illuminate\Http\RedirectResponse
```

src/Http/Livewire/UpdateProfileInformationForm.php
```diff
@@ -50,7 +50,7 @@ public function mount()
      * Update the user's profile information.
      *
      * @param  \Laravel\Fortify\Contracts\UpdatesUserProfileInformation  $updater
-     * @return void
+     * @return \Illuminate\Routing\Redirector|\Illuminate\Http\RedirectResponse|null
```
driesvints:
```suggestion
     * @return \Illuminate\Http\RedirectResponse|null
```

PR #1469, opened 2024-04-03
fix: position of confirm 2fa button to right

allyson-silva:
I changed the position of the confirmation button for enabling two-factor authentication. It was previously displayed to the left of the cancel button, which differs from the Jetstream standard (cancel on the left, confirm on the right). To achieve this, I only needed to change the position of the cancel button code and modify an if statement. I am not yet familiar with integration testing, otherwise I would have proceeded with this action.

I believe that this modification improves the user experience, which is why I have implemented it in a personal project of mine.

button change
This is my application, please ignore the changes in theme (im using daysi ui).


taylorotwell:
I think it's OK as is.

PR #1470, opened 2024-04-05
[5.x] Install API optionally

hafezdivandari:
Currently, install command doesn't respect `--api` option and always call `install:api`. This PR fixes that.


driesvints:
We reverted this unfortunately because of https://github.com/laravel/jetstream/pull/1472

PR #1471, opened 2024-04-08
[5.x] Use `ValidationRule` instead of the deprecated `Rule` class

jdecode:
Use `ValidationRule` instead of the deprecated `Rule` class.

When using `enlightn/enlightn` deprecated code warning started to show up - and this seems to be the cause/fix for that.


taylorotwell:
That's not what this method returns.

PR #1472, opened 2024-04-10
[5.x] Remove Sanctum when api disabled

puklipo:
- Laravel 11
- Jetstream 5

Probably due to #1470, when the API is disabled, Sanctum is missing and an error occurs.

Fix by removing Sanctum.
Since it has been copied to the project, it will be removed both User and UserWithTeams.


taylorotwell:
No instructions on how to recreate this.

taylorotwell:
Reverted breaking PR.

PR #1478, opened 2024-04-24
Added feature check to ProfileInformationTest.php on Inertia

Primilink:


taylorotwell:
If you disable that just remove the test.

PR #1481, opened 2024-04-30
[4.x] Use native <dialog> cancel event, remove unneeded showSlot

Smef:
This PR changes the <dialog> to use a native dialog @cancel event instead of manually adding and removing document-level event listeners when the dialog is shown and dismissed.

This also removes an unnecessary ref which wasn't really doing anything.


taylorotwell:
Feel free to make these adjustments in your own application.

PR #1482, opened 2024-05-04
Add Disabled Class to Primary Button

alnutile:
I add this on all projects maybe the buttons should come with this by default?

Benefit to the user: Allows the developer to show a button is not enabled to the user.

CleanShot 2024-05-04 at 18 51 59@2x

then when enabled:

CleanShot 2024-05-04 at 18 52 15@2x


alnutile:
🙃

PR #1489, opened 2024-05-09
Fix banner

gqrdev:
I eliminated unnecessary line breaks.


taylorotwell:
I can't tell what you are changing. I think you should *only* add the code you want. No formatting changes.

PR #1492, opened 2024-05-14
Change update password message to use passphrase term instead

emmadesilva:
Since it's generally recommended to use passphrases instead of passwords, this updates the "Ensure your account is using a long, random password to stay secure." to tell the user to use a "random passphrase" instead. This change is not breaking as it does not change any code logic. It makes things easier for developers and end users by nudging them to use a more secure password type.


taylorotwell:
No plans to merge atm.

PR #1493, opened 2024-05-15
[5.x] Remove `_method` in `UpdateProfileInformationForm.vue`

osbre:
Makes it a bit simpler. I assume the `_method` property was used when there were no other ways to submit the form.

Also, this is the only place with `_method` in Jetstream and Breeze.


taylorotwell:
Hmm, not sure we should change this without full understanding of why it was historically done.

osbre:
You're 100% right 😅 I just checked the docs:

> Uploading files via put or patch is not supported in Laravel. Instead, make the request via post, including a _method field set to put or patch. This is called [form method spoofing](https://laravel.com/docs/10.x/routing#form-method-spoofing). 

https://inertiajs.com/manual-visits#method

PR #1500, opened 2024-06-05
Update role rule from deprecated Rule to ValidationRule

colq2:
This PR updates the role rule from deprecated `Rule` to `ValidationRule` contract.


taylorotwell:
I think this is fine for now.

PR #1504, opened 2024-06-20
Removed duplicate [x-cloak] styling

benjamincrozat:
Livewire v3 automatically injects styles for `x-cloak` (an attribute provided by Alpine.js) when using the `@livewireStyles` Blade directive.

Since Jetstream does not offer an Alpine.js-only setup, I guess we can safely remove it in _resources/css/app.css_.


taylorotwell:
It's fine if it's not broken. 👍 The "I guess" makes me nervous. 😅

PR #1516, opened 2024-07-14
Fix `CancelTeamInvitation` in the view stub

bretterer:
I found that if using `ULIDs` for teams and team invitations, the `wire:click` for canceling a team invitation fails due to the id not being an integer. 

This is easily fixed by putting single quotes around the property


taylorotwell:
You should be able to correct this in your own application 👍

bretterer:
Hey @taylorotwell.  I'm aware that I can fix this in my own app, however its the ONLY remove in the full jetstream suite that does not support this natively.  I think it should be addressed for better developer experience

PR #1529, opened 2024-08-29
allUsers may error if owner is null

dorkster100:
`Team->allUsers()` may error out, if `$this->owner` is `null`. This may happen if user was deleted (team owner has no foreign index ), meaning user referenced in `user_id` column will not exist in `users` table, thus returning null. In this case `$this->users->merge([null])` which errors.

On further work on my features I found more issues, for example when adding team members, the owner is expected in the stub `AddTeamMember`, I believe it should be possible to add team members to a team without owner (for example in my case I have Admin users).

### DIFF
--- a/src/Contracts/AddsTeamMembers.php
+++ b/src/Contracts/AddsTeamMembers.php
@@ -3,7 +3,7 @@
 namespace Laravel\Jetstream\Contracts;
 
 /**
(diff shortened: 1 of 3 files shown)


taylorotwell:
This should never be null.

dorkster100:
Can you please elaborate a bit? Should users be not deleteable or do you think if user is deleted, then the team should be deleted as well or team should automatically get new owner?

Actually nevermind, I see in the `DeleteUser` action that idea is to delete all owned teams.

PR #1536, opened 2024-09-14
Internationalize role names and descriptions

viktorruskai:
Updated the Jetstream roles to use localization functions for role names and descriptions, enhancing the application’s support for multiple languages. This change ensures that role information is consistent with localized content across the application.


crynobone:
Roles are already translated via `jsonSerialize()`

PR #1551, opened 2024-12-09
Fix stubs team testing - tests fails when feature disabled

ProtocolNebula:

### DIFF
--- a/stubs/tests/inertia/CreateTeamTest.php
+++ b/stubs/tests/inertia/CreateTeamTest.php
@@ -4,11 +4,21 @@ namespace Tests\Feature;
 
 use App\Models\User;
 use Illuminate\Foundation\Testing\RefreshDatabase;
+use Laravel\Jetstream\Features;
 use Tests\TestCase;
 
 class CreateTeamTest extends TestCase
 {
     use RefreshDatabase;
+
+    protected function setUp(): void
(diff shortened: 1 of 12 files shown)


ProtocolNebula:
Hi @taylorotwell , what’s the reason of closing the PR? No reason provided.

When a new project is created with the boilerplates and teams feature is disabled temporary, tests fails.

crynobone:
You can override the stubs for your application if you have these requirements. Temporarily disabling teams Is not a feature intended by Jetstream and having those conditions built-in could cause a false-negative test suite if configuration wasn't done correctly.

ProtocolNebula:
Hi @crynobone , so what about these conditionals? I cannot understand why some are valid but not others (some examples):

- jetstream\stubs\tests\inertia\ApiTokenPermissionsTest.php
- jetstream\stubs\tests\inertia\CreateApiTokenTest.php
- jetstream\stubs\tests\livewire\DeleteApiTokenTest.php

On the other hand, are these valid because the developer might want the teams feature but not the invitation feature?
- jetstream\stubs\tests\livewire\InviteTeamMemberTest.php

Based on the first block (API), they should've removed or this PR is valid, because "API" is an extra component you can not install.

Thanks in advance for the clarification.

PR #1557, opened 2024-12-23
Disallowed create users if feature is disabled

ProtocolNebula:
I noticed that removing `Features::registration()` from `config/fortify.php` still allows external users to create accounts.

I'm not sure if is a bug related to fortify (the repo seems to be fine), but is very hard to go deeper with this quantity of dependencies.

Anyways, this behavior must be controlled by the final application, just like these cases that the feature removal is not working (you can just access to /register and create an account).

The website I'm creating (and for sure, a lot of websites created with laravel) must disable the registration.

So this is more a workaround than a true fix, but I don't know how to fix it correctly, additional supoprt would be welcome.


taylorotwell:
We are rebuilding our starter kits now.

PR #1561, opened 2025-02-08
Allow users to create a Page hierarchy inside the Pages directory

PHPGuus:
This will support Inertia's advocated way of not constantly reloading the Layout component for Member Pages, for example.

Currently, the Jetstream-created Pages, while being adaptable, are loaded from the "Pages/Team/...vue" directory, for example. If one has an Inertia page resolve function that loads and applies a page's layout based on page name prefix (e.g. "Members/Teams/...vue") then the hardcoded paths in the src/Http/Controllers/Inertia/...Controller.php files will not be able to find the pages inside the subdirectories.

This change allows users to set a page prefix so that pages are loaded from the correct subdirectory.

The default value is an empty string.

### DIFF
--- a/config/jetstream.php
+++ b/config/jetstream.php
@@ -6,5 +6,6 @@ return [
     'stack' => 'inertia',
     'middleware' => ['web'],
     'features' => [Features::accountDeletion()],
+    'page_prefix' => '',
     'profile_photo_disk' => 'public',
 ];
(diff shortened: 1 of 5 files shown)


taylorotwell:
We are releasing our new starter kits in 2 weeks so would rather address there if needed.

PR #1582, opened 2025-09-24
Fix Jetstream migration publishing to avoid overwriting existing files

devrabiul:
### Summary
This PR fixes the migration publishing logic in `InstallCommand.php` to ensure existing migrations are not overwritten and preserve the correct order when publishing Jetstream and Fortify migrations.

### Changes
- Collect stub files manually instead of using glob.
- Sort migrations by filename to preserve order.
- Add timestamp increments to new migrations to avoid conflicts.
- Add informative messages when migrations are skipped or published.

### Testing
- Verified migration publishing in a fresh Laravel installation.
- Confirmed existing migrations are skipped.
- Checked that the timestamp order is preserved.


taylorotwell:

PR #1583, opened 2025-10-21
fix(hasTeamRole): prevent null membership role from causing error

mattwells-coex:
The [migration allows `team_user.role` to be `null`](https://github.com/laravel/jetstream/blob/e01b945c9e89d272e7aa10ad2e58a54b194f008c/database/migrations/2020_05_21_200000_create_team_user_table.php#L18) and the `HasTeam::hasTeamRole()` method passed the value directly into the [`Jetstream::findRole()` method which does not allow `null`](https://github.com/laravel/jetstream/blob/e01b945c9e89d272e7aa10ad2e58a54b194f008c/src/Jetstream.php#L96) which can result in an error.

### DIFF
--- a/src/HasTeams.php
+++ b/src/HasTeams.php
@@ -166,9 +166,19 @@ trait HasTeams
             return true;
         }
 
-        return $this->belongsToTeam($team) && optional(Jetstream::findRole($team->users->where(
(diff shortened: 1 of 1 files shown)


mattwells-coex:
Package abandoned?

